PR #12, opened 2015-01-30
bundle loader fix

lapistano:
it turned out, that the G2 setup somehow prevented the SF2 components to be loaded ... this is now fixed.


lapistano:
closed due to wrong branch naming ... successor: https://github.com/libgraviton/graviton/pull/13

PR #14, opened 2015-02-05
Feature/apidoc

narcoticfresh:
This feature branch contains
* 2 new service classes (symfony services) - one has some convenient functions for dealing with our services; the other generates a swagger compliant spec of our service
* Some changes to composer to include swagger-ui (from zip to not include the *.git files) and a bash script onInstall/onUpdate that copies swagger-ui to web/
* A fix of a regression @lapistano did that broke our autoloader in the vendor/ context when graviton is a dependency itself


app/autoload.php
```diff
@@ -4,9 +4,20 @@
 use Composer\Autoload\ClassLoader;
 
 /**
- * @var ClassLoader $loader
+ * don't touch this unless you know what you're doing..
+ * basically we have to autoloaders provided by Composer.
```
hairmare: __two__ autoloaders

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -447,6 +443,26 @@ protected function serialize($result)
     }
 
     /**
+     * Public function to serialize stuff according to the serializer rules.
```
hairmare: I believe this should be in its own *Util class if you really want to use it from other places. Personally I don't feel it's very wise to use the container as a DIC even if its possible to ``getControllerFromRoute`` somehow.

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -83,13 +83,20 @@ public function findAll(Request $request)
 
         // *** do we have an RQL expression, do we need to filter data?
         if (count($request->query->all()) > 0) {
-            $queryParser = new Query(urldecode($request->getQueryString()));
+
+            // prefer explicit filter param!
+            if ($request->query->get('filter') != null && strlen($request->query->get('filter')) > 0) {
```
hairmare: So ``q=`` died again?

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -145,7 +155,7 @@ public function find($documentId)
      * {@inheritDoc}
      *
      * @param string $documentId id of entity to update
-     * @param Object $entity     new enetity
+     * @param Object $entity new enetity
```
hairmare: Doesn't this violate the ENTB phpcs check?

src/Graviton/RestBundle/Model/DocumentModel.php
narcoticfresh: ah, that was before we had our talk..

narcoticfresh:
@hairmare I agree that the need to external calling to serialize/deserialize content reasons for extracting that stuff from the RestController. But an *util class just for that, hmm.. Wouldn't it be suited for the new RestUtils service?

hairmare:
Running ``composer install`` throw the ``Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.`` warning.

hairmare:
Moving ``serializeContent`` ti the RestUtil would in fact be a better solution than keeping it in the controller.

hairmare:
btw, the phpcs slowness seems to have been introduced in 	6840ea0. I can't really explain why, but this also needs fixing before we can merge this (travis is timing out).

narcoticfresh:
yes, i know why it's stalled.. needs to be running and green before we merge..

hairmare:
Weirdly the phpcs seems to have been fixed with your last pushes... Not that I see why, but we seem to be green again :)

narcoticfresh:
I don't know what scrutinizers problem is..?!

src/Graviton/RestBundle/Service/Apidoc.php
```diff
@@ -0,0 +1,253 @@
...
+     *
+     * @return array Basic structure
+     */
+    private function getBasicStructure()
+    {
+        $ret = array();
+        $ret['swagger'] = '2.0';
+        $ret['info'] = array(
+            // @todo this should be a real version - but should it be the version of graviton or which one?
+            'version' => '0.1',
```
hairmare: I had a short talk about this one with @lapistano.

I expect us to use the graviton version as described, but we both agreed that this needs more thinking before we can decide how to solve API versioning properly.

hairmare:
Also, I feel that you should write at least some unittests for ``src/Graviton/RestBundle/Service`` as well as some acceptance tests for ``src/Graviton RestBundle Controller ApidocController.php``.

Maybe those files should also get moved into a bundle of their own (ie. GravitonSwaggerBundle). Right now this would make it pretty hard to move on if we ever need to deprecate swagger.

narcoticfresh:
Hmm.. this makes it hard for me to finish this in a reasonable amount of time.. i need to continue on my critical sprint tasks..

On what can we agree I can do *now* to be able to merge this and define follow-ups for the next sprints on what you just said?

Or do I need to extract some urgent fixes and merge them to develop and we'll have to wait another 2 weeks for our apidoc?

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -72,7 +72,7 @@ public function getContainer()
     public function getAction($id)
     {
         $response = $this->getResponse()
-            ->setStatusCode(Response::HTTP_OK);
+                         ->setStatusCode(Response::HTTP_OK);
```
barumel: Not realy important but this looks a bit ugly. Line indent should be 4 spaces...

narcoticfresh:
So ok.. Swagger stuff is now a bundle by itself. I also refactored some stuff into RestUtils that was in RestController..

composer.json
```diff
@@ -13,6 +13,19 @@
     "autoload": {
         "psr-0": {"": "src/", "Graviton": "src/", "Graviton": "/app/htdocs/src"}
     },
+    "repositories": [
```
lapistano: guess this should be moved of the GravitonSwaggerBundle composer.json

composer.json
```diff
@@ -38,10 +52,10 @@
         "php-jsonpointer/php-jsonpointer": "1.1.*@dev",
         "php-jsonpatch/php-jsonpatch": "1.1.*@dev",
         "nulpunkt/monolog-elasticsearch-logstashformat": "*",
-        "graviton/php-rql-parser": "dev-develop"
+        "graviton/php-rql-parser": "dev-develop",
+        "libgraviton/swagger-ui": "dev-master"
```
lapistano: guess this should be moved of the GravitonSwaggerBundle composer.json

composer.json
```diff
@@ -54,6 +68,7 @@
         ],
         "post-install-cmd": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
+            "sh resources/scripts/swagger-copy.sh",
```
lapistano: guess this should be moved of the GravitonSwaggerBundle composer.json

composer.json
```diff
@@ -61,6 +76,7 @@
         ],
         "post-update-cmd": [
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
+            "sh resources/scripts/swagger-copy.sh",
```
lapistano: guess this should be moved of the GravitonSwaggerBundle composer.json

narcoticfresh:
@lapistano OK, thanks for the input - as the SwaggerBundle is not an bundle in packagist (and imho too specific to be of any use to the public), so no composer.json available..

As I guess @hairmare has day off and I spoke with him that all major issues are addressed; I'll merge it..

lapistano:
@narcoticfresh there is no need for a composer managed repository to life in packagist (like the other repos in the swisscom git).  So why not adding a composer.json to the bundle root and move the bundle to it's own repos - like we already do with a lot of other bundles? I just don't want to be the SwaggerBundle to be a hard dependency in graviton.

narcoticfresh:
AFAIK composer hook scripts are not executed on referenced packages. Only the scripts from the currently executing composer.json are executed.

So as I see it, even if we move the bundle out; we still need to include the swagger-copy in the graviton composer.json.

lapistano:
@narcoticfresh this is true. I was referring to the 'repository' section of the composer.json. in my opinion this shall be part of a composer.json living in the  SwaggerBundle.

narcoticfresh:
@lapistano sadly, the same issue arises there. I've seen that within the composer.json of my graviton wrapper. i needed to include the repository.package also there, it couldn't install it (libgraviton/swagger-ui) from the reference in the graviton composer.json ;-/

so also this needs to be notated in both places anyway ;-/

hairmare:
@narcoticfresh Please refrain from merging your own pull-requests. I was never going to merge this with 0% test coverage. You not being willing to write them ,lead to it being clear to me that I would be writing some before merging.

narcoticfresh:
@hairmare OK..
I thought you're off for the weekend; I need this to be merged now; we can also add tests into develop.. 

Next time, I'll not mix a feature with some fixes; that was a mistake on my part ;-/ sry..

PR #15, opened 2015-02-05
GRV-1221; Feature: authentication listener airlock header

lapistano:
This PR adds the ability to Graviton to authenticate and authorize transactions by identifying a client or account from information provided by an Airlock WAF. 
Currently cookies and http header fields are supported.


app/config/security.yml
```diff
@@ -1,52 +0,0 @@
-# you can read more about security in the related section of the documentation
```
hairmare: Why don't we keep this and use it as base config? Routing also has ``routing.yml`` rather than having ``routing_prod.yml``.

composer.json
```diff
@@ -13,19 +13,6 @@
     "autoload": {
         "psr-0": {"": "src/", "Graviton": "src/", "Graviton": "/app/htdocs/src"}
     },
-    "repositories": [
```
hairmare: Looks like you should merge ``develop`` before we can proceed.

app/config/security.yml
lapistano: the thing is that then every request in (test|dev) has to provide this header filed or cookie. This will be annoying while coding and if you verify the behavior via webbrowser.

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    protected function validateField($header, $fieldName)
+    {
+        $passed = $header->has($fieldName);
+
+        // get rid of anything not a valid character
+        $authInfo = filter_var($header->get($fieldName), FILTER_SANITIZE_STRING);
+
+        // get rid of whitespaces
+        $authInfo = str_replace(array("\r\n", "\n", "\r"), "", trim($authInfo));
```
hairmare: Shouldn't this also take care of ``\t`` and others?

composer.json
lapistano: consider it done

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
```diff
@@ -0,0 +1,81 @@
...
+
+        if (false !== $passed && !empty($authInfo)) {
+            $passed = true;
+        } else {
+            $passed = false;
+        }
+
+        // get rid of control characters
+        if (false !== $passed && $authInfo === preg_replace('#[[:cntrl:]]#i', '', $authInfo)) {
```
hairmare: You might want to wrap this into the above str_replace and do it in one go.

composer.json
lapistano: »nothing to merge« from origin/develop

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
lapistano: since trim() does remove tabs and whitespaces wrapping a string, I guess it is ok to remove them from within the string.

composer.json
hairmare: I'm not sure why, but it doesn't look like you should really be removing the ``libgraviton/swagger-ui`` package.

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
lapistano: I actually like it to be separated. It is much more readable for me.

PR #17, opened 2015-02-10
Feature/grv 1231 bundle name validation

barumel:


src/Graviton/GeneratorBundle/Tests/Generator/ResourceGeneratorProxy.php
```diff
@@ -0,0 +1,30 @@
...
+ * @license  http://opensource.org/licenses/MIT MIT License (c) 2015 Swisscom
+ * @link     http://swisscom.ch
+ */
+class ResourceGeneratorProxy extends ResourceGenerator
+{
+    /**
+     * (non-PHPdoc)
+     * @see \Graviton\GeneratorBundle\Generator\ResourceGenerator::generateDocument()
+     */
+    public function generateDocument($parameters, $dir, $document, $withRepository)
```
hairmare: Do we need this wrapper method? I looks like we could just let ``ResourceGeneratorProxy`` fallback to its parent.

lapistano: this methods does a visibility raise (protected » public) since the method should be tested exclusively.
What we could do is replace hte *Proxy classes by introducing https://github.com/lapistano/proxy-object.
This will create the proxy classes on the fly (like Mock objects are done) and will free us from self defined *Proxy classes (files)

PR #19, opened 2015-02-11
RQL Filter mess cleanup (GRV-1299)

narcoticfresh:
This includes
* We now only accept RQL expressions per "q" GET param. We don't pass the whole query to the RQL Parser anymore..
* The "q" param gets included in the self and paging link headers (triggered via request attribute)
* Added some tests in the controllers that have usable data fixtures to test the new behavior.
* I also consolidated "per_page"/"perPage" mess. Now, we only use "perPage" everywhere, also corrected the test where this was asserted. I think it's clear that all params we accept should be camelCase when we use it everywhere else.

### DIFF
--- a/src/Graviton/EntityBundle/Tests/Controller/CountryControllerTest.php
+++ b/src/Graviton/EntityBundle/Tests/Controller/CountryControllerTest.php
@@ -62,15 +62,15 @@ class CountryControllerTest extends RestTestCase
         $this->assertResponseContentType(self::COL_TYPE, $response);
 
         $this->assertContains(
-            '<http://localhost/entity/country?page='. $pageNumber .'&per_page=10>; rel="self"',
+            '<http://localhost/entity/country?page='. $pageNumber .'&perPage=10>; rel="self"',
             $linkHeader
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page='. ($pageNumber - 1) .'&per_page=10>; rel="prev"',
+            '<http://localhost/entity/country?page='. ($pageNumber - 1) .'&perPage=10>; rel="prev"',
             $linkHeader
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page='. ($pageNumber + 1) .'&per_page=10>; rel="next"',
+            '<http://localhost/entity/country?page='. ($pageNumber + 1) .'&perPage=10>; rel="next"',
             $linkHeader
         );
     }
@@ -105,15 +105,15 @@ class CountryControllerTest extends RestTestCase
         $this->assertResponseContentType(self::COL_TYPE, $response);
 
         $this->assertContains(
-            '<http://localhost/entity/country?page=26&per_page=10>; rel="last"',
+            '<http://localhost/entity/country?page=26&perPage=10>; rel="last"',
             $linkHeader
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page=1&per_page=10>; rel="self"',
+            '<http://localhost/entity/country?page=1&perPage=10>; rel="self"',
             $linkHeader
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page=2&per_page=10>; rel="next"',
+            '<http://localhost/entity/country?page=2&perPage=10>; rel="next"',
             $linkHeader
         );
     }
@@ -133,11 +133,11 @@ class CountryControllerTest extends RestTestCase
         $this->assertResponseContentType(self::COL_TYPE, $response);
 
         $this->assertContains(
-            '<http://localhost/entity/country?page=26&per_page=10>; rel="self"',
+            '<http://localhost/entity/country?page=26&perPage=10>; rel="self"',
             explode(',', $response->headers->get('Link'))
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page=1&per_page=10>; rel="first"',
+            '<http://localhost/entity/country?page=1&perPage=10>; rel="first"',
             explode(',', $response->headers->get('Link'))
         );
     }
@@ -158,28 +158,28 @@ class CountryControllerTest extends RestTestCase
         $this->assertResponseContentType(self::COL_TYPE, $response);
 
         $this->assertContains(
-            '<http://localhost/entity/country?page=1&per_page=10>; rel="self"',
+            '<http://localhost/entity/country?page=1&perPage=10>; rel="self"',
             explode(',', $response->headers->get('Link'))
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page=2&per_page=10>; rel="next"',
+            '<http://localhost/entity/country?page=2&perPage=10>; rel="next"',
             explode(',', $response->headers->get('Link'))
         );
         $this->assertContains(
-            '<http://localhost/entity/country?page=26&per_page=10>; rel="last"',
+            '<http://localhost/entity/country?page=26&perPage=10>; rel="last"',
             explode(',', $response->headers->get('Link'))
         );
     }
 
     /**
-     * check if per_page param works on collections
+     * check if perPage param works on collections
      *
      * @return void
      */
     public function testFindAllWithPerPageParam()
     {
         $client = static::createRestClient();
-        $client->request('GET', '/entity/country?per_page=20');
+        $client->request('GET', '/entity/country?perPage=20');
 
         $response = $client->getResponse();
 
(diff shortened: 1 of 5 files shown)


src/Graviton/EntityBundle/Tests/Controller/CountryControllerTest.php
```diff
@@ -62,15 +62,15 @@ public function testFindAllWithPaging($queryString, $pageNumber)
         $this->assertResponseContentType(self::COL_TYPE, $response);
 
         $this->assertContains(
-            '<http://localhost/entity/country?page='. $pageNumber .'&per_page=10>; rel="self"',
+            '<http://localhost/entity/country?page='. $pageNumber .'&perPage=10>; rel="self"',
```
hairmare: Why are we changing this? For some reason the old way ``per_page`` seems to still lurk somewhere. It there a reason for this change?

src/Graviton/EntityBundle/Tests/Controller/CountryControllerTest.php
```diff
@@ -188,15 +188,36 @@ public function testFindAllWithPerPageParam()
...
+        $client = static::createRestClient();
+        $client->request('GET', '/entity/country?q='.urlencode('eq(capitalCity,Luanda)'));
+
+        $response = $client->getResponse();
+
+        $this->assertEquals(1, count($client->getResults()));
+        $this->assertResponseContentType(self::COL_TYPE, $response);
+
+        $this->assertContains(
+            '<http://localhost/entity/country?q=eq%28capitalCity%2CLuanda%29>; rel="self"',
```
hairmare: There should also be some tests that check for paging headers on filtered pages.

narcoticfresh: There is, the one in /app/core..

narcoticfresh:
@hairmare The change from per_page to perPage just seemed logical - I started to accept perPage in RestController quite some time ago.. it just makes sense that the param is camelCase - don't you agree?

If it's still lurking somewhere, then I can clean them all up and then it's ok? Or you don't want it to be camelCase?

hairmare:
To me it seemed logical that parameters should be underscore based. I choose ``per_page`` due to the fact that github is using that in their v3 api (ie. the one that also does header based pagination). I had originally planned against using camelCase in get params.

Personally i find the underscore syntax much friendlier to read in URLs.

src/Graviton/EntityBundle/Tests/Controller/CountryControllerTest.php
hairmare: Thx, I probably haven't seen that one yet :P

narcoticfresh:
@hairmare We have camelCase everywhere.

Anyway - please note that "per_page" is still working if you pass it yourself (in DocumentModel) but we'll generate perPage from that point on.. 

Is that acceptible or do we really need to change it back and have one lonely c_named_case field in this wide ocean of camelCase fields? ;-)

hairmare:
Well symfony coding guidelines suggested using _ for parameters, but I digress.

If per_page still works we should remove. I doesn't make sense to support more than one different style of this param. Could you check that we really got rid of all per_page cases? I believe it's also still mentioned in some docs and needs fixing there as well.

src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php
```diff
@@ -101,15 +101,21 @@ private function generateParameters($routeType, Request $request)
...
-            if ($request->attributes->get('perPage')) {
-                $parameters['per_page'] = $request->attributes->get('perPage');
+        } elseif ($routeType == 'all') {
+            if ($request->attributes->get('paging')) {
+                $parameters = array('page' => $request->get('page', 1));
+                if ($request->attributes->get('perPage')) {
+                    $parameters['perPage'] = $request->attributes->get('perPage');
+                }
+            }
+            if ($request->attributes->get('filtering')) {
```
hairmare: This should not be nested in the if/elseif block that only pertains to paging.

narcoticfresh: Hm, no it doesn't, look at it again pls..
The elseif here is the "all" routeType, down we handle paging & filtering seperately..

src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php
```diff
@@ -101,15 +101,21 @@ private function generateParameters($routeType, Request $request)
...
             // handle post request by rewriting self link to newly created resource
             $parameters = array('id' => $request->get('id'));
         } elseif ($routeType != 'all') {
             $parameters = array('id' => $request->get('id'));
-        } elseif ($request->attributes->get('paging')) {
-            $parameters = array('page' => $request->get('page', 1));
-            if ($request->attributes->get('perPage')) {
-                $parameters['per_page'] = $request->attributes->get('perPage');
+        } elseif ($routeType == 'all') {
+            if ($request->attributes->get('paging')) {
```
hairmare: Somehow these two nested ifs don't make me happy. Maybe this should only check for ``perPage`` and ``page`` and pass those on. I can't see the profit we gain by checking for ``paging`` here.

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -75,23 +75,18 @@ public function setRepository(ObjectRepository $countries)
...
-        $numberPerPage = (int) $request->query->get(
-            'perPage',
-            $request->query->get('per_page', 10)
-        );
+        $numberPerPage = (int) $request->query->get('perPage', 10);
         $startAt = ($pageNumber - 1) * $numberPerPage;
 
         // *** do we have an RQL expression, do we need to filter data?
-        if (count($request->query->all()) > 0) {
+        if ($request->query->get('q') != null && strlen($request->query->get('q')) > 0) {
```
hairmare: Wouldn't ``!empty`` just work here? This kinda looks like a WTF...

src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php
narcoticfresh: this is not new. i did the same with 'filtering' to follow that schema and implement a suggestion from @lapistano (using request attributes)..

src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php
hairmare: Yeah i do know that it only belongs to the routeType = 'all' case. Still, we need to keep the complexity of this method where it was. Nesting ifs so deeply leads to this having to many execution paths.

src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php
hairmare: This was pertaining to the inner if/if nesting. The outer if/elseif probably needs to stay.

src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php
narcoticfresh: You do realize that we need to handle paging and filtering together in the 'all' routeType, but only if they are requested (each checked independently)?

Do you prefer to have less nesting but more elseif'es for each and every case? (a) only paging, b) only filtering, c) paging + filtering). The amount of execution pathes will be the same, but with less calculated complexity because it will have less nesting - but it will be hell more unreadable..

Or probably i just don't get the point..

hairmare: Something like follows would probably be preferable:

````php
        $isAllRoute = $routeType == 'all';
        if (blah) {
        } elseif ($isAllRoute) {
            if ($request->attributes->get('paging')) {
                $parameters = array('page' => $request->get('page', 1));
            }
            if ($request->attributes->get('perPage')) {
                $parameters['perPage'] = $request->attributes->get('perPage');
            }
        }

        if ($isAllRoute && $request->attributes->get('filtering')) {
            $parameters = array('q' => $request->get('q', ''));
        }
````
As it stands this will most definitively need refactoring as soon as we need to add the next param.

File: src/Graviton/RestBundle/Listener/SelfLinkResponseListener.php [hunk 6]
```diff
@@ -101,18 +101,25 @@ private function generateParameters($routeType, Request $request)
...
-                $parameters['per_page'] = $request->attributes->get('perPage');
+                $parameters['perPage'] = $request->attributes->get('perPage');
             }
         }
 
+        if ($routeType == 'all' && $request->attributes->get('filtering')) {
+            $parameters = array('q' => $request->get('q', ''));
+        }
+
         return $parameters;
```
narcoticfresh: @hairmare  I now broke it up more so we have less nesting..
This should be fine now, so merge if it's ok like that..

hairmare: Thx. I'll let travis-ci run through and maybe wait for a scrutinizer run. I'll probably get to merging it in a couple of hours (mainly due to travis). All in all it lgtm now.

hairmare: Unexpectedly, I'm still not happy. In fact, The last change made it worse rather than better, sorry about that.

I'll just mark this up in the code for now. If we ever need to add more params we have to split this up into multiple classes that get it right without resorting to having pagination and filtering intertwined in one method. The decorator pattern might work for this.

hairmare:
Merged as per my last comment

PR #24, opened 2015-02-18
get rid of now unused GravitonSecurityBundleTestCase; tests now inherit ...

narcoticfresh:
...WebTestCase from our TestBundle

As discussed, remove unsused TestCase class in SecurityBundle


hairmare:
Thanks

PR #25, opened 2015-02-19
fix rql parser dep; we need to integrate 1.* now

narcoticfresh:
as said.. rql is completely broken if we don't fix it now; also in the wrappers it's broken now..


narcoticfresh:
I checked; composer installs 1.1.0 currently; just that what we need now until we have the bundle..

hairmare:
We might make the bundle provide the 1.x version when it's done. For now this is very much needed.

PR #27, opened 2015-02-23
as discussed, let's remove translatables from schema document.. 

narcoticfresh:
this way, swagger will display them under the model..

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
@@ -336,7 +336,7 @@ class AppControllerTest extends RestTestCase
         $this->assertResponseContentType('application/schema+json', $response);
         $this->assertEquals(200, $response->getStatusCode());
 
-        $this->assertEquals('Array of app objects', $results->title->en);
+        $this->assertEquals('Array of app objects', $results->title);
         $this->assertEquals('array', $results->type);
         $this->assertIsAppSchema($results->items);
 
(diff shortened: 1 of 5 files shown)


hairmare:
We should also not implement ``TranslatableDocumentInterface`` to really get rid of this. I still need to look into what repercussions that might have though. Implementing the interface and returning empty does not make alot of sense.

You'll also need to update the now failing tests that look for a translated schema.

narcoticfresh:
OK; I removed the interface and changed the tests accordingly..

Don't be confused that you'll still find 'name->en->type' somewhere in the tests (as I was confused) - those are assertions that check the type of the 'en' property of an effectively Translatable property, so this is correct. At least until we remove the whole Translatable thing, if we do that..

hairmare:
Would you like to do the honors of adding the JSON-Schema ``$ref`` or should we leave that until later?

narcoticfresh:
Well, that's a different topic, I'm doing that in another branch.. I'm also not sure what u mean..

Thing is, I just want to show the user the response type of a OPTIONS request - if I notate a $ref to the JSON schema by http (to //json-schema.org/draft-04/schema), sadly swagger-ui is not able to show that schema..

So question is, if we still notate it like that or show our own small schema file for the schema..

hairmare:
I know that its different and yeah I was taking about (re-)adding the $ref tag to the OPTIONS schema.

I had removed it because of this i18n issue. So if you already have it on a branch I'll be happy to merge it ASAP. Otherwise I'll add a todo somewhere...

narcoticfresh:
the problem is not that something is missing in OPTIONS schema..

it's purely that we need something to show swagger users what response to expect from a OPTIONS request.. and as said, a $ref to the official schema doesn't produce any visible response spec in swagger-ui..

hairmare:
true, all it does is make our schema json-schema compliant

PR #28, opened 2015-02-23
refactor file loading in dymanic generator

hairmare:
I split this out from my working branch so we can merge it before it becomes to large.

I'm taking one small step towards refactoring the ``graviton::generate::dynamicbundle`` command by refactoring the file scanner that get the json files.

It adds a Loader class that implements a file, dir and scan strategy, scan being the current default.

Right now this only takes care of file based definitions. We should also take care of the mongodb loader, but that needs more refactoring to fit in here as well. It's supported by a monkey-patch in the command.

It still has some small issues where I need input on:
* [x] why is the ``--srcDir`` option using during scanning, is this a needed feature?
* [x] both dir and scan now ignore files starting with ``_``. This was only in dir before, does scan need to find ``_`` json files?

Oh and btw, the json files in the Tests dir are just showcase over and over again.


src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -120,57 +121,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $filesToWorkOn = $this
+            ->getContainer()
+            ->get('graviton_generator.definition.loader')
+            ->load($input->getOption('json'))
+        ;
 
         // bundles in mongodb?
-        $filesToWorkOn = array_merge($filesToWorkOn, $this->getDefinitionsFromMongoDb());
+        // @todo this should move to loader
+        foreach ($this->getDefinitionsFromMongoDb() AS $mongoDef) {
```
lapistano: as vs AS

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -261,7 +220,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $arguments = array(
```
lapistano: for me it looks pretty like the argument definition for sub-resources and could be consolidated.

src/Graviton/GeneratorBundle/Definition/Loader/Loader.php
```diff
@@ -0,0 +1,55 @@
...
+ * - scan
+ * - mongodb
+ */
+
+namespace Graviton\GeneratorBundle\Definition\Loader;
+
+use Graviton\GeneratorBundle\Definition\Loader\Strategy\StrategyInterface;
+
+/**
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
```
lapistano: didn't we agree on class level docblocks?

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/DirStrategy.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return boolean
+     */
+    public function accepts($input)
+    {
+        return is_dir($input);
+    }
+
+    /**
+     * load
```
lapistano: Tries to load every file in the given directory with the extension *.json.

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
hairmare: yupp, pending push ;)

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
```diff
@@ -75,6 +75,16 @@ public function getId()
     }
```
lapistano: I  like to propose a proper error handling for json_decode() (https://github.com/libgraviton/graviton/blob/feature/file-loading-in-dymanic-generator/src/Graviton/GeneratorBundle/Definition/JsonDefinition.php#L64) not to get a PHP Fatal somewhere... see http://php.net/manual/en/function.json-last-error.php

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
hairmare: yeah, I'm not touching those in this PR though :D

src/Graviton/GeneratorBundle/Definition/Loader/Loader.php
hairmare: I was doing comment in file level and ``@tags`` in class now, I think the sniffer doesn't really care anymore.

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/DirStrategy.php
hairmare: thanks, in next push

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
lapistano: so I propose to create a story for that not to be forgotten ;)

narcoticfresh:
@hairmare 
* The thing about ignoring files starting with '_' is just pure convenience.. it allows us to easily 'disable' loads by renaming them.. ml3k does the same.. and to be able to put stuff inside the directory structure (like docs and examples) in a way they don't interfere with anything (directories)..

IMHO no load filename should start with '_', that would mean it's "id" property also needs to start with '_' which means the generated bundle would start with '_'

* 'srcDir' is used there mainly because the default value (as defined in the Input options) for that option is the srcDir we want to start scanning. this is only used in not-wrapper-context (ie no 'vendor' above)

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
lapistano: or use a assertion class (or parts of it) https://github.com/beberlei/assert

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * load
+     *
+     * @param string|null $input input from command
+     *
+     * @return JsonDefinition[]
+     */
+    public function load($input)
+    {
+        $testMode = strpos($input, '/Tests/') !== 0;
```
lapistano: looks almost the same as in DirStrategy() .. consolidate?

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
hairmare: Again, the code you found looks just as I expected. I'm sure symfony has something we can leverage to do this proper, but it's a different battle.

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
hairmare: yeah, that one only recently got sane enough to kinda look the same, will do

hairmare:
@narcoticfresh  regarding the ``srcDir``

So basically if I just do the vendor sniffing and go for %kernel.root_dir% otherwise we'd be safe?

If it really needs us to be able to pass a dir that needs checking for resource/definitions subdir I would re-do that as a seperate flag. Like maybe ``--scan-json-dir`` or whatnot.

narcoticfresh:
@hairmare Yes i think that's safe.. vendor sniffing and fall back to the root_dir if not.. I see that's not so good as-is..

Regarding `--scan-json-dir`, IMHO that's not useful. Point is, at the start you *needed* to pass want you want generate. But in fact, letting the generator search is the de-facto *standard* behavior (most invocations let it scan) - so IMHO if you don't pass anything, just let it scan.. that's a good default behavior as it's the most used case..

hairmare:
Ok. So I'll do one more small refactor, satisfy travis and scurtinizer and them remove the wip.

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
hairmare: this is now refactored to share the looping code

hairmare:
wip removed, @narcoticfresh do we need to test this with something real before merging?

I added a bunch of tests and also did some dev-check'ing so far.

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
lapistano: it is #differentBattle 
and yes the JsonDecoder (JMSSerializer) covers this as well.

narcoticfresh:
@hairmare If i just run it now, `$filesToWorkOn` has 2 files:
* Showcase.json
* test.json from Tests/

IMHO this is wrong, everything in Tests/ shouldn't be generated by the generator as ran from the cli..

narcoticfresh:
I checked out this branch in the wrapper context and it seems to be still fine - except as said, that test.json (and everything in */Tests/*) shouldn't be found..

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
```diff
@@ -0,0 +1,72 @@
...
+        return new \RegexIterator(
+            $iterator,
+            '/.*\/resources\/definition\/[^_].+\.json$/i',
+            \RegexIterator::GET_MATCH
+        );
+    }
+
+    /**
+     * @param string|null $input input from command
+     * @param string|null $input input from command
```
lapistano: according to usage $file should be an array ... the first char of a string can never be 'Test' ;)

lapistano: I really like type hints in that scope ;)

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return is_null($input);
+    }
+
+    /**
+     * @param string|null $input input from command
+     *
+     * @return \RegexIterator
+     */
+    protected function getIterator($input)
```
lapistano: $input is not used.

hairmare:
@narcoticfresh argh, that damn Tests file. It should get filtered but it doesnt. I'll recheck.

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
hairmare: fixed in next push

src/Graviton/GeneratorBundle/Definition/Loader/Strategy/ScanStrategy.php
hairmare: I removed the args in this case. It's not like their meant to be public API.

I next push...

hairmare: Removing didn't work. The args need to be defined since they need to implement the same interface as the parent class.

I originally had this as.

```php
return (!empty($input) && !empty($file));
```

But then decided that I'd rather not do it that way.

Another classical solution to this problem would be:

```php
unset($input);
unset($file);
return true;
```

Do you have any idea what best-practices in area would be?

narcoticfresh:
Seems to be fine now, my wrapper could generate everything as before, test.json doesn't get picked up anymore..

So I guess we can merge this?

hairmare:
Ok to merge from my point of view.

PR #29, opened 2015-02-23
low hanging swagger fruits

narcoticfresh:
those are really low hanging swagger fruits ;-)

* re-enable OPTIONS requests
* create a schema file for SchemaModel to show that as expected response from the OPTIONS request
* sort the resources by path name (makes much for sense)
* change the grouping of the resources the same as we agreed for Graviphoton, so we have less confusing long names (FinancingAmortization group just is group Financing now, with all it below that)
* restructure the class a little bit..

*Please* I know all this is shitty, we established that, we have some tasks on the way for that - just focus on the fruits ;-)


src/Graviton/SchemaBundle/Model/SchemaModel.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Graviton\SchemaBundle\Model;
 
+use MyProject\Proxies\__CG__\OtherProject\Proxies\__CG__\stdClass;
```
hairmare: whats this?

src/Graviton/SwaggerBundle/Service/Swagger.php
```diff
@@ -255,4 +183,145 @@ private function getBasicStructure()
...
+
+    /**
+     * Returns the definition of a 'schema' response
+     *
+     * @return stdClass Schema
+     */
+    protected function getSchemaSchema()
+    {
+        $schemaModel = new SchemaModel();
+        return $schemaModel->getSchema();
```
hairmare: You should inject the results of ``getSchema()`` from the DIC and get rid of this method.

narcoticfresh: How should I get that by the DIC?

Thing is, in this case, I need the 'raw' schema, basically the pure contents of the file as SchemaUtils is not able to run it's normal routine involving the doctrine metadata..

hairmare:
All in all a large improvement over how the service was before. Scrutinizer is a bit slow today, but I'm sure he'd be happy ;)

narcoticfresh:
I hope he's happier then before ;-)

I removed the WTF line that my IDE put there.. the other point, well, i don't get it.. ;-)

src/Graviton/SwaggerBundle/Service/Swagger.php
hairmare: Do this in ``services.xml`` in the ``graviton.rest.apidoc`` service:

```xml
            <call method="setSchema">
                <argument type="service" id="graviton.schema.model.schemamodel"></argument>
            </call>
```

And then just call ``getSchema`` that one time you need it.

You could also do setter-injection since the constructor has no params yet.

hairmare:
inject ``SchemaModel`` from the DIC and I'll merge...

PR #18, opened 2015-02-10
update codesniffer and fix CI issues

hairmare:
This updates the ENTB codesniffer to its latest version. It fixes it's whole integration problems and bumps it to a rather current version of phpcs.

It also relaxes the ENTB code-style so there should be no surprises.

This also has the scrutinizer conditions removed. I've set up some more relaxed ones through the web interface, but will re-tune them once we remove some wtfs.

The changes I did are mostly clean up and not comprehensive yet. I cleaned up the ``@author`` and ``@link`` annotations and removed some legacy stuff like ``@package``.

There will probably be more to be done at a later stage when we've decided on a phpdoc,sami, ...  like solution to go forward with but I'm not tackling that in this pr.

All changes really worth reviewing are in the dotfiles, ``composer.*`` and ``dev-check.sh``. The rest is just bringing our docblocks up to speed and fixing stuff.

PR #30, opened 2015-02-23
externalize getVersion()

narcoticfresh:
We had a version determination in the MainController. this is clearly the wrong place..

So this contains
* A new CoreUtils service containing a getVersion() method - yes, using the same code as before
* Let MainController use that service

This is preparation so swagger and others can display the same version to the user..

### DIFF
--- a/src/Graviton/CoreBundle/Controller/MainController.php
+++ b/src/Graviton/CoreBundle/Controller/MainController.php
@@ -65,11 +65,7 @@ class MainController implements ContainerAwareInterface
         );
 
         $response->headers->set('Link', (string) $links);
-
-        # @todo don't find the composer file like so, use packagist to find and parse it if possible
-        $composerFile = __DIR__.'/../../../../composer.json';
-        $composer = json_decode(file_get_contents($composerFile), true);
-        $response->headers->set('X-Version', $composer['version']);
+        $response->headers->set('X-Version', $this->container->get('graviton.core.utils')->getVersion());
 
         $mainPage = new \stdClass;
         $mainPage->message = 'Please look at the Link headers of this response for further information.';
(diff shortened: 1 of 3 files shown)


src/Graviton/CoreBundle/Controller/MainController.php
```diff
@@ -65,11 +65,7 @@ public function indexAction()
         );
 
         $response->headers->set('Link', (string) $links);
-
-        # @todo don't find the composer file like so, use packagist to find and parse it if possible
-        $composerFile = __DIR__.'/../../../../composer.json';
-        $composer = json_decode(file_get_contents($composerFile), true);
-        $response->headers->set('X-Version', $composer['version']);
+        $response->headers->set('X-Version', $this->container->get('graviton.core.utils')->getVersion());
```
lapistano: I propose to use a kernel response filter for that .. (I already implemented this in GRV-1347), But I guess for now core.utils is fine

narcoticfresh:
what is a kernel response filter? can we merge this - as you said it's fine for now?

if you want a kernel response filter - whatever that is - you need to point me to some sources, not GRV's ;-)

lapistano:
@narcoticfresh  I just created a PR for that: https://github.com/libgraviton/graviton/commit/50b495086f35dba3bae41aee40b5d6f762326773

hairmare:
You also might want to call ``CoreUtils`` just ``VersionUtils`` or similar. Chances are we'll have more versiony stuff to do anyhow.

lapistano:
@narcoticfresh as I said, I think it is fine for now. It will be moved to a kernel response filter once the other PR is merged ;)

narcoticfresh:
@lapistano It's fine that a Listener provides the X-Version header - but the 'heavy lifting' *needs* to be done in a service so other parts (as swagger or other services) can also easily get the current version..

hairmare:
Composer has the following on this issue.

> Optional if the package repository can infer the version from somewhere, such as the VCS tag name in > the VCS repository. In that case it is also recommended to omit it.

They then go on to say.

> Specifying the version yourself will most likely end up creating problems at some point due to human error.

Which is all very much true. I should now, I just broke #18 due to this.

Can we get the necessary info from ``composer.lock``? Should we grab it from git?

I've been looking for 3rd party solutions to this but seem to be running into a brick wall on every attempt.

narcoticfresh:
Please.. This PR just wants to put the version determination into a own simple class - we don't need to solve the whole versioning problem here. Once externalized, we can always change the way the version is determined.

PR #31, opened 2015-02-23
feature/grv 1347 reactivate sf2 view controller and templates

lapistano:


src/Graviton/RestBundle/Listener/XVersionResponseListener.php
```diff
@@ -0,0 +1,93 @@
...
+    private $file;
+
+
+    /**
+     * @param LoggerInterface $logger
+     */
+    public function __construct(LoggerInterface $logger, $file = '')
+    {
+        $this->logger = $logger;
+        $this->file = !empty($file) ? $file : __DIR__ . '/../../../../composer.json';
```
narcoticfresh: As said.. getting and returning the current version should be in a easy to use Service class so others can use it as well..

app/Resources/views/base.json.twig
```diff
@@ -0,0 +1,3 @@
+{% block main %}
+    <h1> this is to be replaced </h1>
```
hairmare: Wouldn't it make sense to just ``{{ response|raw }}`` here? Otherwise the same twig code is repeated over and over with no clear benefits.

phpunit.xml.dist
```diff
@@ -28,6 +28,8 @@
         <directory>src/*/*Bundle/*Bundle/Tests</directory>
         <directory>src/*/*Bundle/Resources</directory>
         <directory>src/*/Bundle/*Bundle/Resources</directory>
+        <directory>src/*/Bundle/DependencyInjection</directory>
```
hairmare: I'm not really sure about these. They do actually get test coverage when the cache is empty and I feel like it's to easy to hide stuff from travis/scrutinizer in here when these are removed.

src/Graviton/CoreBundle/Controller/MainController.php
```diff
@@ -5,11 +5,12 @@
...
 
-use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 use Graviton\RestBundle\HttpFoundation\LinkHeader;
 use Graviton\RestBundle\HttpFoundation\LinkHeaderItem;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Router;
```
hairmare: It's rather hard to read the real changes here. Could you tell your IDE to stop reordering use flags or is there a benefit to this?

src/Graviton/CoreBundle/Controller/MainController.php
```diff
@@ -44,55 +45,101 @@ public function setContainer(ContainerInterface $container = null)
...
 
-        $restUtils = $this->container->get('graviton.rest.restutils');
-        $optionRoutes = $restUtils->getOptionRoutes();
+        $response->setContent(json_encode($mainPage));
+        $response->setStatusCode(Response::HTTP_OK);
+        $response->headers->set('Link', $this->prepareLinkHeader($router));
 
+        // todo: make sure, that the correct content type is set.
+        // todo: this should be covered by a kernel.response event listener?
+        $response->headers->set('Content-Type', 'application/json');
```
hairmare: Yeah, this certainly won't stay here. I don't think it works with twig, but have we considered doing something ala ``Symfony/Component/HttpFoundation/JsonResponse``?

src/Graviton/CoreBundle/Controller/MainController.php
```diff
@@ -44,55 +45,101 @@ public function setContainer(ContainerInterface $container = null)
...
+    {
+        $sortArr = array();
         $services = array_map(
             function ($routeName) use ($router) {
                 list($app, $bundle, $rest, $document) = explode('.', $routeName);
                 $schemaRoute = implode('.', array($app, $bundle, $rest, $document, 'canonicalSchema'));
 
                 return array(
-                    '$ref' => $router->generate($routeName, array(), true),
+                    '$ref'    => $router->generate($routeName, array(), true),
```
hairmare: AFAIK ``phpcbf`` reverts these again...

src/Graviton/CoreBundle/Controller/MainController.php
```diff
@@ -44,55 +45,101 @@ public function setContainer(ContainerInterface $container = null)
...
+                $router->generate('graviton.core.rest.app.all', array(), true),
+                array(
+                    'rel'  => 'apps',
+                    'type' => 'application/json'
+                )
+            )
+        );
 
-        return $response;
+        return (string) $links;
```
hairmare: Are you sure you don't want to let the Response do the casting? That would be more like the CookieHeader thing does.

src/Graviton/CoreBundle/Service/CoreUtils.php
```diff
@@ -5,44 +5,40 @@
...
     {
-        /**
-         * @todo don't find the composer file like so, use packagist to find and parse it if possible
-         * @todo if we're in a wrapper context, use the version of the wrapper, not graviton
-         */
-        $composerFile = __DIR__.'/../../../../composer.json';
-        $composer = json_decode(file_get_contents($composerFile), true);
-        return $composer['version'];
+        //@todo if we're in a wrapper context, use the version of the wrapper, not graviton
+        $composerFile = !empty($composerFile) ? $composerFile : __DIR__ . '/../../../../composer.json';
```
hairmare: This should use symfony/finder and the ``serializer`` service that returns a ``Symfony\Component\Serializer\Encoder\DecoderInterface``. Maybe drop a todo?

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
```diff
@@ -57,7 +57,7 @@ public function testAppsLink()
 
         $this->assertContains(
             '<http://localhost/core/app>; rel="apps"; type="application/json"',
-            explode(',', $response->headers->get('Link'))
+            $response->headers->get('Link')
```
hairmare: Will this still work with multiple Link headers? I always ``explode`` them to be  sure to have an array.

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
```diff
@@ -97,4 +97,99 @@ function ($service) {
...
+    public function testDetermineServices()
+    {
+        $services = [
+            [
+                '$ref'    => "http:\/\/localhost\/core\/product",
+                "profile" => "http:\/\/localhost\/schema\/core\/product\/collection"
+            ],
+            [
+                '$ref'    => "http:\/\/localhost\/core\/app",
+                "profile" => "http:\/\/localhost\/schema\/core\/app\/collection"
```
hairmare: You should use single quotes in these URLs for readability.

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
```diff
@@ -97,4 +97,99 @@ function ($service) {
...
+
+        $this->assertEquals(
+            [
+                [
+                    '$ref'    => "http:\/\/localhost\/core\/app",
+                    "profile" => "http:\/\/localhost\/schema\/core\/app\/collection"
+                ],
+                [
+                    '$ref'    => "http:\/\/localhost\/core\/product",
+                    "profile" => "http:\/\/localhost\/schema\/core\/product\/collection"
```
hairmare: Same as above, use single quotes for large strings with alot of slashes.

src/Graviton/CoreBundle/Tests/Controller/ProductControllerTest.php
```diff
@@ -97,6 +97,9 @@ public function testProductIsCached()
 
         $this->assertInternalType('string', $etag);
 
+        // client ha to be rebuild since the AppKernel will be resetted after a request
```
hairmare: client **has** to be rebuild since the AppKernel will be **reset** after a request

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
```diff
@@ -35,8 +35,8 @@ public function testVersionHeader()
         $client = static::createRestClient();
         $client->request('GET', '/');
 
-        $version = json_decode(file_get_contents(__DIR__.'/../../../../../composer.json'), true);
-        $version = $version['version'];
+        $composer = json_decode(file_get_contents(__DIR__ . '/../../../../../composer.json'), true);
+        $version = $composer['version'];
```
hairmare: You could mock the new ``CoreUtils`` here instead of depending on the ``composer.json`` that should have its version removed at some point.

src/Graviton/ExceptionBundle/Exception/NoInputException.php
```diff
@@ -21,8 +21,10 @@
      *
      * @param string     $message Error message
      * @param \Exception $prev    Previous Exception
+     *
+     * @return \Graviton\ExceptionBundle\Exception\NoInputException
```
hairmare: The short form would also be ok here.

src/Graviton/ExceptionBundle/Listener/ExceptionListener.php
```diff
@@ -27,53 +27,6 @@ class ExceptionListener
...
-                $exception->getHeaders(),
-                $headers
-            );
-        }
-
-        $response->headers->replace($headers);
-        $response->setData($message);
-
-        $event->setResponse($response);
-        $response->send();
```
hairmare: Please kill the rest of this file also. As it stands this contains to much obsolete uncalled code.

src/Graviton/I18nBundle/Tests/Controller/LanguageControllerTest.php
```diff
@@ -70,10 +70,16 @@ public function testAddAndUseNewLanguage()
 
         $this->assertEquals('en', $response->headers->get('Content-Language'));
 
+        // client ha to be rebuild since the AppKernel will be resetted after a request
```
hairmare: see my earlier comment on how to s/// these

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -39,25 +39,25 @@ class RestController implements ContainerAwareInterface
...
-     * {@inheritdoc}
-     *
-     * @param \Symfony\Component\DependencyInjection\ContainerInterface $container service_container
+     * Get the container object
      *
-     * @return void
+     * @return \Symfony\Component\DependencyInjection\ContainerInterface
      */
-    public function setContainer(ContainerInterface $container = null)
+    public function getContainer()
```
hairmare: reordering methods like this makes this file very hard to review. Same as with the reordered use statements but even harder.

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -70,13 +70,113 @@ public function getContainer()
...
+            ->setStatusCode(Response::HTTP_OK);
 
         $record = $this->findRecord($id);
 
         $response->setContent($this->serialize($record));
 
-        return $response;
+        return $this->render(
+            'GravitonRestBundle:Main:index.json.twig',
+            array('response' => $response->getContent()),
```
hairmare: Does it make sense to get this from the response rather than directly from the data? We could just add in the data and then serialize in twig for more lightweight action methods.

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -93,11 +193,27 @@ public function allAction()
...
-                                );
+        $response = $this->getResponse()
+            ->setStatusCode(Response::HTTP_OK)
+            ->setContent(
+                $this->serialize($model->findAll($this->getRequest()))
+            );
+
+        return $this->render(
+            'GravitonRestBundle:Main:index.json.twig',
+            array('response' => $response->getContent()),
```
hairmare: See above on using the response for passing a string. We should serialize in twig and not pass a string through the response to get it into twig. Applies to all other actions in this file as well.

src/Graviton/RestBundle/Listener/XVersionResponseListener.php
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        /** @var \Symfony\Component\HttpFoundation\Response $response */
+        $response = $event->getResponse();
+        $response->headers->set(
+            'X-Version',
+            $this->coreUtils->getVersion()
+        );
+    }
+}
```
hairmare: nice

src/Graviton/RestBundle/Resources/config/services.xml
```diff
@@ -1,151 +1,174 @@
...
+        </service>
+
+        <!-- X-Version response listener -->
+        <service id="graviton.rest.listener.xversionresponselistener"
+                 class="%graviton.rest.listener.xversionresponselistener.class%">
+            <argument type="service" id="graviton.core.utils" />
+            <argument type="service" id="logger" />
+            <tag name="kernel.event_listener" event="graviton.rest.response" method="onKernelResponse"/>
+        </service>
+    </services>
```
hairmare: This one is almost impossible to review. Could you summarize the changes?

src/Graviton/CoreBundle/Controller/MainController.php
lapistano: My idea was to remove it once the response filter is created setting the correct response header depending on the request Accept header.

hairmare: I see, I didn't have that one in mind earlier.

src/Graviton/CoreBundle/Service/CoreUtils.php
lapistano: this should use the CoreUtils  ... which will tackle this.

hairmare: this is the CoreUtils, and since you where already dropping todos here left and right ;)

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
lapistano: assertContains() finds the first appearance of the given string in a string or array. See https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertContains.example2

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
lapistano: #done

src/Graviton/ExceptionBundle/Listener/ExceptionListener.php
lapistano: this will be done in an other story (next sprint) :)

src/Graviton/RestBundle/Resources/config/services.xml
lapistano: lines 16 && 166ff

app/Resources/views/base.json.twig
lapistano: afaik there must be a dedicated template to be addressed in the render() method. So just to be sure I created the base template and let each controller handle the output by it's own .. if you see a better solution, I am eager to hear it.

src/Graviton/CoreBundle/Controller/MainController.php
lapistano: #done

src/Graviton/CoreBundle/Controller/MainController.php
lapistano: the return value will directly be used as value for the header field.

src/Graviton/CoreBundle/Service/CoreUtils.php
lapistano: I was in the wrong context ;)

lapistano: nevertheless I agree. I propose to do this in another PR.

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
hairmare: ok, all good then

src/Graviton/CoreBundle/Tests/Controller/MainControllerTest.php
hairmare: thanks

src/Graviton/RestBundle/Resources/config/services.xml
hairmare: Thanks, looks good

app/Resources/views/base.json.twig
hairmare: I was thinking that the dedicated templates could just be like so:
```
{% extend "base.json.twig" %}
```
I also feel the message "this is to be replaced" is not really specific enough. It should point to the whole thing being our fault.

composer.json
```diff
@@ -58,7 +58,8 @@
         "phpunit/phpunit": "~4.6",
         "liip/functional-test-bundle": "~1.0",
         "squizlabs/php_codesniffer": "~2.2",
-        "libgraviton/codesniffer": "~1.3"
+        "libgraviton/codesniffer": "~1.3",
+        "lapistano/proxy-object": "~2.1"
```
hairmare: Could you mention proxy-object in TestBundles README and maybe add a link and a short one-line intro?

hairmare:
We need to refactor RestController again soon as scrutinizer has raised some IMHO legit issues. I'm merging this nevertheless since it also does lots of good :)

src/Graviton/RestBundle/Controller/RestController.php
hairmare: as discussed, we might touch this later, but we're leaving it for now

phpunit.xml.dist
hairmare: Look here for an example of why testing DependencyInjection make sense: https://github.com/libgraviton/GravitonVcapBundle

That bundle only has business logic in the DI stuff. I still need to do GravitonVcapBundle prod to make sure (I plan on announcing it on vcap-dev or so when it's done though).

PR #34, opened 2015-02-24
small refactor on resource generator

hairmare:
Puts the field mapping code into its own ``mapField`` method and adds tests for the new method.

It also decouples ``ResourceGenerator`` from ``InputInterface`` by moving ``JsonDefinition`` creation to the calling command. This is a first step towards moving the definition creation towards the dic.

Also adds some todos for future refactor potential and cleans up some docs and hinting things.


hairmare:
I fixed tests by activating ``E_ALL`` reporting due to the reason for the last error being a missing variable that would have been caught by ``E_STRICT``.

Also fixed the underlying issue that was introduced in a last minute refactor (a71886e).

@narcoticfresh could you point me in the direction of a json that should not get a controller? I'm having a hard time finding one to smoke-screen the no-controller option.

narcoticfresh:
@hairmare An example would be 'CustomerContact.json' in the fi**ova (to avoid legal names) loadPackage (inherited in the s*kb-* packages). This has no service URI defined and thus should not have a controller.. Or Balance.json is another..

CustomerContact may change as soon as I merged a PR in Stash..

hairmare:
Thanks for the examples, will look in the f\*ing F\* that s\* are using :trollface:

hairmare:
I did some smoke-screening and everything seems to look ok now. Balance.json is not generating a Controller but, for instance, CustomerNote.json is generating one.

AFAIK this is the intended behaviour.

narcoticfresh:
hm, something is quite wrong.. i checked it out on my wrapper, suddenly all the paths to the generated services changed.. ;-/

i.e. a have a url http://localhost:8000/financingmortgagecost/financingmortgagecost, which differs from the one defined in the json file..

so this doesn't work as expected anymore..

narcoticfresh:
AND IMPORTANT: i have a service /balance/balance now - this one one (as we said) that shouldn't be generated as it hasn't have a service uri defined thus should not have a controller..

hairmare:
This has been taken care of with #117 #118 #119.

PR #37, opened 2015-02-25
Implemented ability to be more verbose when authenticating a contract.

lapistano:


hairmare:
You still have some phpcs dues to pay.

hairmare:
Other than that, lgtm.

src/Graviton/SecurityBundle/Authentication/AirlockAuthenticationKeyAuthenticator.php
```diff
@@ -122,10 +124,10 @@ public function supportsToken(TokenInterface $token, $providerKey)
      * called by authentication listeners inheriting from
      * AbstractAuthenticationListener.
      *
-     * @param \Symfony\Component\HttpFoundation\Request                          $request   original request
-     * @param \Symfony\Component\Security\Core\Exception\AuthenticationException $exception exception from auth attempt
+     * @param Request                 $request   original request
+     * @param AuthenticationException $exception exception from auth attempt
      *
-     * @return Response The response to return, never null
+     * @return Response|null
```
hairmare: Why the ``null``? From a quick glance at the method it seems to me that this can just return the Response.

lapistano: Reading the source proves the docBlock wrong ;) 
see \Symfony\Component\Security\Http\Authentication\SimpleAuthenticationHandler::onAuthenticationSuccess()

lapistano: Also the only purpose of the 'custom' onAuthenticationSuccess() is to log the successful authentication and not to redirect or rewrite the current response.

hairmare: Ah I think I get it. You're not returning ``null` anywhere since we need to indicate a network auth fail and the default fail handler won't be of much help in these cases.

I'll merge the PR in a bit...

PR #38, opened 2015-02-25
fix CORS headers on OPTIONS request with id

hairmare:
This is based on a case found by the frontend team. For some reasons GravitonRestBundle was not generating rules for OPTIONS requests with id. This lead to them not getting proper CORS headers in the responses.

I'm testing with a small test that is based on getting OPTIONS from ``/i18n/translatable/i18n-de-German`` since this is the module where the problems where first identified.


narcoticfresh:
Looks good to me..

PR #39, opened 2015-02-27
simple ability to add more bundles to dynamic bundlebundle via parameter..

narcoticfresh:
read param (normally empty) and incorporate class names from there as additional bundles into dynamic bundlebundle

### DIFF
--- a/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
@@ -407,7 +407,7 @@ class GenerateDynamicBundleCommand extends ContainerAwareCommand
 
         // add optional bundles by param..
         $additions = json_decode(
-            $this->getContainer()->getParameter('generator.bundlebundle.additions'),
+            $this->getContainer()->getParameter('generator.bundlebundle.additions', false),
             true
         );
         if (is_array($additions)) {
--- a/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
@@ -404,6 +404,16 @@ class GenerateDynamicBundleCommand extends ContainerAwareCommand
     private function generateBundleBundleClass()
     {
         $dbbGenerator = new DynamicBundleBundleGenerator();
+
+        // add optional bundles by param..
+        $additions = json_decode(
+            $this->getContainer()->getParameter('generator.bundlebundle.additions'),
+            true
+        );
+        if (is_array($additions)) {
+            $dbbGenerator->setAdditions($additions);
+        }
+
         $dbbGenerator->generate(
             $this->bundleBundleList,
             $this->bundleBundleNamespace,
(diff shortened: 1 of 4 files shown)


src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -404,6 +404,16 @@ public function getGeneratedValidationXmlPath($namespace)
     private function generateBundleBundleClass()
     {
         $dbbGenerator = new DynamicBundleBundleGenerator();
+
+        // add optional bundles by param..
+        $additions = json_decode(
+            $this->getContainer()->getParameter('generator.bundlebundle.additions'),
```
lapistano: I'd add a default return value to avoid PHP Notices: 

$this->getContainer()->getParameter('generator.bundlebundle.additions', array())

lapistano: The parameters »generator.bundlebundle.additions« has to be added to the /app/config/parameters.yml.dist to satisfy the Incenteev Parameters Builder to make sure the parameters.yml contains the parameter after the composer install/update scripts were run.

$ echo "generator.bundlebundle.additions: ~" >> app/config/parameters.yml.dist

narcoticfresh: Adding it there will make it necessary to define it - this should not be necessary.. if it doesn't exist (which is perfectly fine), we'll just ignore it..

narcoticfresh:
I added it with a default value of false, get it with fallback value false.. shouldn't be empty array if nothing is actually set..

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
lapistano: My point is that it will be forgotten, if not set (even to ~ (aka NULL)) in parameters.yml.
Another way is to add the parameter to the ReadMe.md, which I would be fine with.

narcoticfresh:
we'll discard this anyway again when the proper bundle register thing is done..

lapistano:
am just waiting for the build to be successful be successful before I merge

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -404,6 +404,16 @@ public function getGeneratedValidationXmlPath($namespace)
     private function generateBundleBundleClass()
     {
         $dbbGenerator = new DynamicBundleBundleGenerator();
+
+        // add optional bundles by param..
+        $additions = json_decode(
```
hairmare: I don't get why we need to have json encoded data in parameters. Couldn't we just store this as array params?

http://symfony.com/doc/current/components/dependency_injection/parameters.html#component-di-parameters-array

narcoticfresh:
Because this will only be set via ENV vars in wrapper context to inject bundles.. and as JSON it's easy to define array in bash as ENV vars  ;-)

Anyway, all this is temporary until we get proper registration possibilities..

lapistano:
@hairmare  see: https://issue.swisscom.ch/browse/GRV-1230

hairmare:
Not much to see there ;)

The ENV reasoning is ok with me, I'm doing something similar on cf with ``VCAP_SERVICES``.

Maybe there should be a comment somewhere that explains these types of vars. I'm quite certain I didn't explain vcap thing proper (addressed in #40).

lapistano:
I am on it ;) #storyWriting

PR #40, opened 2015-02-27
use Incenteev\\ParameterHandler\\ScriptHandler::buildParameters

hairmare:
The script was already properly configured in composer.json so I see no reason not to use it as intended.

In the same spirit as the recent composer cleanup I'm also removing a boatload of parameters that we don't need anymore and add some docs to ``vcap.services`` which is not self explanatory otherwise.


hairmare:
Yeah as discussed, we'll keep this one open a bit to make sure everyone knows it's implications and has a chance to backup their ``parameters.yml`` before we merge it.

hairmare:
@narcoticfresh Have looked at this? It might disrupt your workflow.

app/config/parameters.yml.dist
```diff
@@ -1,25 +1,17 @@
...
-    # database_path: "%kernel.root_dir%/data.db3"
-
     mongodb.default.server.db: db
     mongodb.default.server.uri: ~
 
-    graviton.log.elasticsearch.host: ~
     graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
     graviton.authentication.user_provider.model: ~
 
+    generator.bundlebundle.additions: false
```
narcoticfresh: this shouldn't be here at all (`generator.bundlebundle.additions`) - every param that influences the generator should not be set at all in the files as *existing* params cannot be overriden via ENV

narcoticfresh:
@hairmare well i don't think the implication is too big (apart from the `generator.bundlebundle.additions` removal I need)..

*but* two things in my mind
* IMHO the default connection of doctrine to point it to sqlite should be in the `*.dist` already I suggest. Problem is, if this is *not set*; Doctrine defaults to mysql and also demands `php_mysql` ext to be installed. It makes more sense to fix that by setting sqlite on dbal..
* The drawback to remove it would maybe be on cloud push - if somebody pushes without a `parameters.yml` (which shouldn't be done as I understand), the invocation of `buildParameters()` by composer would open an interactive prompt to input the new params - what obviously would fail on the cloud..

hairmare:
The ``generator.bundlebundle.additions`` removal is an error i need to fix.

On the other two points:
* The default connection as it stands now does not depend on mysql. It does depend on ``pdo`` and we should probably put that into ``composer.json``.
* In the long run we need to find a solution for these. I would prefer adding them to ``.cfignore`` but we then need a different solution for pushing overriden configs to the cloud. Right now you can push without a ``parameters.yml`` and incenteev will create one in non-interactive mode ([gravity-platform/registry](https://github.com/gravity-platform/registry) is an example of this since it pushes from travis).

app/config/parameters.yml.dist
hairmare: hmmm... good to know, I had not realized that being defined make parameters non-overridable like that. It looks like I'll have to revisit the ``vcap.services`` that actually wants the same behaviour.

hairmare:
@lapistano @narcoticfresh I don't think we're blocked on this anymore.

I'm unassigning myself and hoping one of you merges this soonish.

PR #43, opened 2015-03-02
fix multi version constraint in composer.json

hairmare:
composer should actually support both versions (ie ">=0.13 <2" and ">=0.13,<2", but the version without a comma is broken in composer-1.0.0-alpha9. For us the easiest fix is to just not use the non working syntax
and use a form that works in the composer version that the build pack installs.


hairmare:
@narcoticfresh this should fix your issues ;)

I turned out to be easier to fix on our side since all that the buildpack is lacking is a new composer release that contains the fix.

hairmare:
fixed here, debug help there: cloudfoundry/php-buildpack#47

PR #45, opened 2015-03-03
CompilerPass to find services needed for generating routes

hairmare:
Load routes tagged as ``graviton.rest`` during container compilation phase by using a compiler pass.

We'll still need to identify all the places where we need to clear the cache after this change. One area is certainly after generating bundles. I'm working on adding that to ``graviton:generate:`` on a branch as well.

For the time being we'll have to run ``app/console cache:clear`` after running the generators.

With this SYMFONY_ENVs aren't broken anymore. I was able to run graviton under the prod env.


lapistano:
lgfms! 
The need of running cache:clear after the bundle and entity generation outside the dev env makes totally sense, when taking into account that the dev env does disable the cache completely and caching is welcome in any other environment ;)

PR #46, opened 2015-03-03
clear cache after generating resource

hairmare:
This might help with cache problems introduced by libgraviton/graviton#45.

We still need to find out of those really are an issue and if this is best put here. I'm also not sure if I should make this an option or not.

Feedback welcome ;)

### DIFF
--- a/src/Graviton/GeneratorBundle/Command/GenerateResourceCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/GenerateResourceCommand.php
@@ -10,6 +10,7 @@ use Sensio\Bundle\GeneratorBundle\Command\GenerateDoctrineEntityCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\ArrayInput;
 
 /**
  * generator command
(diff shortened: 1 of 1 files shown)


src/Graviton/GeneratorBundle/Command/GenerateResourceCommand.php
```diff
@@ -64,6 +70,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
+        $commandInput = new ArrayInput(array(
+            'command' => 'cache:clear',
+            '--no-warmup' => true
+        ));
+        $command = $this->getApplication()->find('cache:clear');
+        if ($command->run($commandInput, $output) == 0) {
+            $output->isVerbose() && $output->writeln(
+                'cache cleared'
+            );
```
lapistano: I find the formatting quite hard to read.

```php
if ($command->run($commandInput, $output) == 0) {
    $output->isVerbose() && $output->writeln('cache cleared');
}
```
would be easier for me to read.

hairmare: changed

hairmare:
@narcoticfresh do we need this? I would add some tests and remove WIP if its needed but the need seems to have gone away...

hairmare:
Not needed...

PR #47, opened 2015-03-04
Feature/security logger to listener

lapistano:
just to be sure not to forget logging if other authentication providers are setup.


narcoticfresh:
@lapistano Looks fine to me.. 

Just for curiosity - now you removed the `onAuthenticationSuccess` stuff from `AirlockAuthenticationKeyAuthenticator` - where is that stuff now? ;-)

lapistano:
@narcoticfresh it moved to the AuthenticationLogger.
Despite the fact onAuthenticationSuccess interface is able to handle a NULL returned by the callback, it asks for a Response object to be returned, but in our case there is just some logging to be done. 
Therefore I moved the logging to a dedicated Eventlistener and removed both logging actions from the specific event callbacks in the Authenticator.

narcoticfresh:
ah, seen now, it was just logging anyway.. I see.. ACK

app/config/parameters.yml
```diff
@@ -16,7 +16,7 @@ parameters:
     mongodb.default.server.db: db
     mongodb.default.server.uri: 'mongodb://localhost:27017'
     graviton.log.elasticsearch.host: 'http://elasticsearch.beta.scapp.io:80'
-    graviton.security.authentication.strategy: graviton.security.authentication.strategy.header
+    graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
```
hairmare: This is why we need #40 (compare this with parameters.yml.dist)

PR #48, opened 2015-03-04
add field 'order' to /core/app

narcoticfresh:
* add 'order' field to /core/app
* add new fixture for app 'tablet' - this is the one the /configuration/module fixtures will point to
* changed tests to reflect changes..

### DIFF
--- a/src/Graviton/CoreBundle/DataFixtures/MongoDB/LoadAppData.php
+++ b/src/Graviton/CoreBundle/DataFixtures/MongoDB/LoadAppData.php
@@ -27,19 +27,11 @@ class LoadAppData implements FixtureInterface
      */
     public function load(ObjectManager $manager)
     {
-        $helloApp = new App;
-        $helloApp->setId('hello');
-        $helloApp->setTitle('Hello World!');
-        $helloApp->setShowInMenu(true);
-        $helloApp->setOrder(1);
-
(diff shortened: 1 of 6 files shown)


narcoticfresh:
I'm not sure if 'tablet' is a so good choice.. but we need something to reflect our app "ev0j4", but name neutral (IMHO).. so 'tablet' is our first app that will contain modules..

hairmare:
I'm OK with the order field but I don't see any value in you changes with the additional feature. It might have been better to just rename the "Hello World" fixture to "Tablet" if this is really needed.

The reason this really bothers me is that it is making the ``AppControllerTest`` longer without adding to what is covered by it.

I'm OK with merging, but we really need to start keeping test data used by external devs seperate from fixture data used for unit tests.

narcoticfresh:
Yes, we need to seperate the fixtures somehow in the future..
Additonally, consumers will need 'explicit' test fixtures (brand and product names in apps and modules) which we can't or shouldn't put in github.. I'm unsure how we should address this..

narcoticfresh:
Hm, I just see that both DELETE and PUT tests use the 'hello world' resource that would now become Tablet.. So all those tests new to work with the 'tablet' app.. is that OK?

hairmare:
totes ok

PR #50, opened 2015-03-04
add /core/module

narcoticfresh:
this is the new 'module' service as requested in the issue..
I simplyfied the naming of fields according to what we have already.. So..

Name as in issue | Name in service
------------ | -------------
"String key" | key (values also camelCased)
"Core app" | appId
localizedText | name


narcoticfresh:
I thought CoreBundle is the right place as it's a companion to /app/core - agree?

hairmare:
I also think it belongs in CoreBundle. I'm still not happy with the ``/configuration`` namespace in the URL though.

At some point I was considering that we might want to express this kind of thing like ``/core/app/<my-great-app>/modules`` but I'm not sure we currently support such patterns.

What do you think about doing ``/core/config/modules`` or so? I prefer having ``/core/`` in it just so it's clear which bundle it's from. It also nicely namespaces all the paths that are reserved for people using graviton as a server generator.

hairmare:
Oh and thanks for renaming those fields. They where horrible ;)

narcoticfresh:
@hairmare as discussed; /core/module it is..

PR #51, opened 2015-03-04
GRV-1308: deleted wrapper service for Request causing request...

lapistano:
...s to fail/be empty occasionally when SF2 issued subrequests


src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -425,13 +425,14 @@ public function patchAction($id)
...
      *
-     * @param string $id ID of record
+     * @param string  $id      ID of record
+     * @param Request $request Current http request
      *
+     * @throws SerializationException
      * @return \Symfony\Component\HttpFoundation\Response $response Result of the action
      */
-    public function optionsAction($id = null)
+    public function optionsAction($id = null, Request $request)
```
hairmare: I suspect, defining $id as null should be done in the router in this case.

hairmare:
I hope you where already done. It worked for me and looked good :D

PR #52, opened 2015-03-04
generator composer integration

narcoticfresh:
* Make symfony command to clean GravitonDyn
* Wrap our 2 commands in a composer ScriptHandler
* Integrate that into composer.json
* Remove that from dev-cleanstart.sh

### DIFF
--- a/src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
@@ -1,4 +1,7 @@
 <?php
+/**
+ * cleans dynamic bundle directory
+ */
 
 namespace Graviton\GeneratorBundle\Command;
 
--- a/src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
@@ -5,7 +5,7 @@
 
 namespace Graviton\GeneratorBundle\Command;
 
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\Output;
@@ -19,8 +19,22 @@ use Symfony\Component\Process\Process;
  * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
  * @link     http://swisscom.ch
  */
-class CleanDynamicBundleCacheCommand extends ContainerAwareCommand
+class CleanDynamicBundleCacheCommand extends Command
 {
+
+    /**
+     * kernel
+     *
+     * @var \Graviton\AppKernel
+     */
+    private $kernel;
+
+    /**
+     * filesystem
+     *
+     * @var \Symfony\Component\Filesystem\Filesystem
+     */
+    private $filesystem;
 
     /**
      * {@inheritDoc}
@@ -38,6 +52,30 @@ class CleanDynamicBundleCacheCommand extends ContainerAwareCommand
     }
 
     /**
+     * set kernel
+     *
+     * @param mixed $kernel kernel
+     *
+     * @return void
+     */
+    public function setKernel($kernel)
+    {
+        $this->kernel = $kernel;
+    }
+
+    /**
+     * set filesystem
+     *
+     * @param mixed $filesystem filesystem
+     *
+     * @return void
+     */
+    public function setFilesystem($filesystem)
+    {
+        $this->filesystem = $filesystem;
+    }
+
+    /**
      * {@inheritDoc}
      *
      * @param InputInterface  $input  input
@@ -48,13 +86,10 @@ class CleanDynamicBundleCacheCommand extends ContainerAwareCommand
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         // @todo it was suggested this may/should be moved to app/cache..?
-        $dynamicDir = $this->getContainer()->get('kernel')->getRootDir().'/../src/GravitonDyn/';
+        $dynamicDir = $this->kernel->getRootDir().'/../src/GravitonDyn/';
 
-        $process = new Process('rm -Rf '.escapeshellarg($dynamicDir));
-        $process->run();
-
-        if (!$process->isSuccessful()) {
-            throw new \RuntimeException($process->getErrorOutput());
+        if ($this->filesystem->exists($dynamicDir)) {
+            $this->filesystem->remove($dynamicDir);
         }
     }
 }
--- a/src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Graviton\GeneratorBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\Output;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Process\Process;
+
+/**
+ * Deletes the GravitonDyn/ folder..
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class CleanDynamicBundleCacheCommand extends ContainerAwareCommand
+{
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this->setName('graviton:clean:dynamicbundles')
+             ->setDescription(
+                 'Removes the folder with the generated dynamic bundles'
+             );
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @param InputInterface  $input  input
+     * @param OutputInterface $output output
+     *
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        // @todo it was suggested this may/should be moved to app/cache..?
+        $dynamicDir = $this->getContainer()->get('kernel')->getRootDir().'/../src/GravitonDyn/';
+
+        $process = new Process('rm -Rf '.escapeshellarg($dynamicDir));
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            throw new \RuntimeException($process->getErrorOutput());
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
```diff
@@ -0,0 +1,60 @@
...
+use Symfony\Component\Process\Process;
+
+/**
+ * Deletes the GravitonDyn/ folder..
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class CleanDynamicBundleCacheCommand extends ContainerAwareCommand
```
hairmare: This would be much cleaner if it did not extend ``ContainerAwareCommand``.

src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @param InputInterface  $input  input
+     * @param OutputInterface $output output
+     *
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        // @todo it was suggested this may/should be moved to app/cache..?
+        $dynamicDir = $this->getContainer()->get('kernel')->getRootDir().'/../src/GravitonDyn/';
```
hairmare: You would then need to inject a kernel using the injection containers xml configs for getting this done.

The command should also be tagged as a command there.

src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
```diff
@@ -0,0 +1,60 @@
...
+
+        $process = new Process('rm -Rf '.escapeshellarg($dynamicDir));
```
hairmare: While your at it you could also inject a ``Process`` instance and use its ``setCommandLine`` method to set the command.

This would make this rather dangerous method nicely testable. Being dangerous as it is it really needs that testing.

hairmare:
Could you write some tests for the command. Also you might what to consider replacing ``rm -rf`` with ``rm -r``.

In most cases ``-r`` should be more than enough and I think other cases warrant user intervention. We should probably start checking the umask before generation to ensure this and let the user make an informed decision.

Maybe ``-f`` this could be an optional flag to the command.

The Composer ScriptHandler and integration is looking good as it stands.

hairmare:
@narcoticfresh let me know what you think about the ``-f`` thing and I'll write up a GRV as needed.

hairmare:
nevermind about the ``-f`` thing, i thought i through and my mind seems to have been throwing blanks when I came up with it being an issue.

narcoticfresh:
> This would be much cleaner if it did not extend `ContainerAwareCommand.`

Cleaner? Isn't `ContainerAwareCommand` exactly for that case - to use it, when I need the container? What's the point of removing the inherit, then have to go through lengths to get the container again?

How to "inject a kernel using the injection containers" without a container, I'm not sure..

Most functions btw (also like `execute()`) i borrowed/nicked directly from the Sensio DistributionBundle ScriptHandler..

narcoticfresh:
This PR will need to wait until time is around to do it..

narcoticfresh:
To sum it up, I think we should leave the `ContainerAware` to simplify stuff.. if not, I need a example how to inject stuff without a container, I'm unsure how to do this..
I'm also unsure how to inject a `Process` in  the first place, where to get it from?

And finally, I'm not sure how to write a test for a Command that only deletes a dir()..

src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
lapistano: Why using the Process component? I don't see a benefit here. 
Instead you could use the FileSystem compoment (http://symfony.com/doc/current/components/filesystem/introduction.html) to delete the directory. 
Nevertheless it is testable in both cases using test doubles to stub the behaviour of either the Process or FileSystem class handling the deletion of the directory.

hairmare:
It's actually a bit of a ruse that being ``ContainerAware`` make things simpler. As soon as you inject the container you get a dependency of the most coupled component in the system. 

To inject services into a command you would define the command as follows.

```xml
        <service id="graviton_generator.command.cleandynamicbundle"
            class="Graviton\GeneratorBundle\Command\CleanDynamicBundleCacheCommand">
            <tag name="console.command" />
        </service>
```

Now you can remove ``ContainerAwareCommand`` since the container knows how to find the container on it's own (more details are in [the cookbook](http://symfony.com/doc/current/cookbook/console/commands_as_services.html)).

Next you would implement the setters ``setProcess(Process $process)`` and ``setKernel(Kernel $kernel)`` so you can do setter injection.

To finish up you would then define new services and inject them into the now non-``ContainerAware`` command.

```xml
        <!-- services to inject -->
        <service id="graviton_generator.command.cleandynamicbundle.kernel" alias="kernel"/>
        <service id="graviton_generator.command.cleandynamicbundle.process"
            class="Symfony\Component\Process\Process">
        </service>

        <!-- service to inject into -->
        <service id="graviton_generator.command.cleandynamicbundle"
            class="Graviton\GeneratorBundle\Command\CleanDynamicBundleCacheCommand">
            <!-- setter injection -->
            <call method="setKernel">
                <argument type="service" id="kernel"/>
            </call>
            <call method="setProcess">
                <argument type="service" id="graviton_generator.command.cleandynamicbundle.process"/>
            </call>
            <tag name="console.command" />
        </service>
```

Voila. You know have command that depends on a kernel and a process instance rather than depending on the container in it's whole.

In testing you can now to ``$this->getMock('Symfony\Component\Process\Process');`` and pass that into ``setProcess`` instead of the real fs to validate if your code calls it with the intended ``setCommandLine`` parameters.

narcoticfresh:
@hairmare Thanks.. I wasn't aware those can be tagged and that services.xml will be honored in the command context..

src/Graviton/GeneratorBundle/Resources/config/services.xml
```diff
@@ -29,5 +29,42 @@
...
+        <service id="graviton_generator.command.generatedynamicbundle.process"
+                 class="Symfony\Component\Process\Process">
+            <argument type="string" key="null" />
+        </service>
+
+        <service id="graviton_generator.command.generatedynamicbundle"
+                 class="Graviton\GeneratorBundle\Command\GenerateDynamicBundleCommand">
+            <call method="setContainer">
+                <argument type="service" id="graviton_generator.command.generatedynamicbundle.container"/>
+            </call>
```
hairmare: You don't seem to do anything with the service container you are injecting here. It could possibly be removed.

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -341,9 +378,14 @@ private function executeCommand(array $args, OutputInterface $output)
             )
         );
 
-        passthru($cmd, $exitCode);
+        $this->process->setCommandLine($cmd);
+        $this->process->run();
```
hairmare: Have you considered calling the ``generate:dynamicbundle`` command through the AppKernel like [so](http://symfony.com/doc/current/components/console/introduction.html#calling-an-existing-command)? It works like [here](https://github.com/libgraviton/graviton/pull/46/files).

src/Graviton/GeneratorBundle/Resources/config/services.xml
narcoticfresh: @hairmare This is the new part of the `GenerateDynamicBundle` command, not the clean! There, the container is used quite significantly..

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
narcoticfresh: @hairmare This is not scope of this PR as this is `GenerateDynamicBundleCommand`.. I thought it needs to be as `Process` as the `Kernel` needs to be refreshed after each bundle(?)..

src/Graviton/GeneratorBundle/Resources/config/services.xml
```diff
@@ -29,5 +29,42 @@
...
+            <call method="setKernel">
+                <argument type="service" id="graviton_generator.command.cleandynamicbundle.kernel"/>
+            </call>
+            <call method="setFilesystem">
+                <argument type="service" id="graviton_generator.command.cleandynamicbundle.filesystem"/>
+            </call>
+            <tag name="console.command" />
+        </service>
+
+        <service id="graviton_generator.command.generatedynamicbundle.container" alias="service_container"/>
```
lapistano: I see a practice like this a lot in G2, what is the benefit instead of using 'service_container' directly? 
Even worse, stuff like this did cause strange behaviors in the past when the request service was 'wrapped' like this.

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
hairmare: I was thinking generatedynamicbundle already wraps the individual calls to the generator command so we might be safe there.

src/Graviton/GeneratorBundle/Composer/ScriptHandler.php
```diff
@@ -0,0 +1,130 @@
...
+     */
+    protected static function executeCommand(CommandEvent $event, $consoleDir, $cmd, $timeout = 300)
+    {
+        $php = escapeshellarg(self::getPhp(false));
+        $console = escapeshellarg($consoleDir.'/console');
+        if ($event->getIO()->isDecorated()) {
+            $console .= ' --ansi';
+        }
+
+        $process = new Process($php.' '.$console.' '.$cmd, null, null, null, $timeout);
```
lapistano: escapeshellarg($cmd) ??

src/Graviton/GeneratorBundle/Resources/config/services.xml
hairmare: Ah I see, so basically the Generator asks for the commands container. In that case we can leave it in (for now). Could you document this in the file so we remember when we look at it and it's unclear why it uses the container.

narcoticfresh: I added a todo.. it uses it quite alot.. to get the json file helper, to get params, to get the kernel..

src/Graviton/GeneratorBundle/Composer/ScriptHandler.php
narcoticfresh: Done..

narcoticfresh:
@hairmare Think is ready to merge..

hairmare:
I think so to :)

src/Graviton/GeneratorBundle/Command/CleanDynamicBundleCacheCommand.php
```diff
@@ -0,0 +1,95 @@
...
+     * {@inheritDoc}
+     *
+     * @param InputInterface  $input  input
+     * @param OutputInterface $output output
+     *
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        // @todo it was suggested this may/should be moved to app/cache..?
```
hairmare: we haz issue: https://issue.swisscom.ch/browse/GRV-1448

hairmare:
I think this takes care of [GRV-1441](https://issue.swisscom.ch/browse/GRV-1441) and am closing it as well.

I wasn't able to get rid of ``rm -rf vendor/graviton/graviton/src/GravitonDyn`` in gravity-platform/registry@e849045b73ac40954d89f3e4091b114afd279f2e but, well do that in [GRV-1448](https://issue.swisscom.ch/browse/GRV-1448)

PR #53, opened 2015-03-04
throw at least some errors in graviton:generate:dynamicbundles

hairmare:
These don't change the fact that we need to implement some kind of validation
and logging in these areas of our code.

If one happens to have a small error in a definition it should not take that much
debugging to figure it out :)

I think we already have a GRV on this, for now this is what happend due to a typo
of mine that i realized to late.

### DIFF
--- a/src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
+++ b/src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
@@ -60,6 +60,10 @@ class JsonDefinition
         }
 
         $this->doc = json_decode(file_get_contents($this->filename));
+
+        if (empty($this->doc)) {
+            throw new \RuntimeException(sprintf('Could not load %s', $filename));
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
```diff
@@ -60,6 +60,10 @@ public function __construct($filename)
         }
 
         $this->doc = json_decode(file_get_contents($this->filename));
+
+        if (empty($this->doc)) {
```
lapistano: if (empty($this->doc) && JSON_ERROR_NONE !== json_last_error()) [...]

see http://php.net/manual/de/function.json-last-error.php

hairmare: In this case I don't even care about json anymore. Even if we where to load a file containing ``null``  this would still me more than invalid enough.

We should rewrite this to use ``JsonDecode`` for the json validation step and also start doing validation on the documents contents before they end up in this definition.

I think this should be done somewhere in the loader. Going forward we should also remove the actual file reading from this class and get it done by the loader as well.

All the loader currently does is find the needed files. That needs changing at some point.

lapistano: d'accord

PR #54, opened 2015-03-05
swagger should come with https support

hairmare:


hairmare:
@narcoticfresh I think this small change should not break anything and fix swagger.nova.scapp.io in the http department.

We should start thinking about ssl-everywhere as thats coming soon!

PR #55, opened 2015-03-05
Implemented GRV-1418: voter to grant access to own contract info

lapistano:


phpunit.xml.dist
```diff
@@ -29,6 +29,8 @@
         <directory>src/*/*Bundle/Resources</directory>
         <directory>src/*/Bundle/*Bundle/Resources</directory>
         <directory>src/*/Bundle/DependencyInjection</directory>
+        <directory>src/Graviton/TestBundle</directory>
```
hairmare: ignoring this bundle is a bad idea since we won't be able to see what parts of our test harness consist of dead, never executing code.

hairmare:
This is rather demanding to review. I'll have to read over it again later today and also check it out and run it to make sure i grok it

Have you considered adding some words of encouragement to the bundles README. The tbd mark in there looks like its waiting for it.

phpunit.xml.dist
lapistano: so we will life with the fact that it reduces the coverage percentage. I guess it does not make any sense to write tests for it.

lapistano:
@hairmare we could go over it together in you want.

phpunit.xml.dist
hairmare: Yes we should be able to live with the reduced coverage (if its much). That is why we're only aiming for >80% coverage and not for more.

The bundle does have things worth ignoring, but if feel they should be ignored inline with annotations to ensure that we can test the module individually at some stage.

Also, It is possible do add stuff that is used for acceptance tested that should itself be fully unit-tested.

I feel we are already ignoring too much of it due to us ignoring the DI component (ie. I want an inactive bundle with just a DI component to show up as unused code in coverage).

lapistano: removed it

hairmare: Thanks. I'll be in Gum later today, maybe we can go over the rest then.

src/Graviton/SecurityBundle/Voter/OwnContextVoter.php
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Return an array of supported classes. This will be called by supportsClass
+     *
+     * @return array an array of supported classes, i.e. array('Acme\DemoBundle\Model\Product')
+     */
+    protected function getSupportedClasses()
+    {
+        return array(
+            'GravitonDyn\AccountBundle\Document\Account',
+            'GravitonDyn\CustomerBundle\Document\Customer',
```
hairmare: We should replace this with an ``AccountInterface`` and a ``CustomerInterface``. Those should not be in ``GravitonDyn`` but they should be implemented by the two classes in ``GravitonDyn``.

Otherwise this is unfair to someone that does not have the ``ml3k-loaderconfig`` stuff.

src/Graviton/SecurityBundle/Voter/OwnContextVoter.php
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Return an array of supported attributes. This will be called by supportsAttribute
+     *
+     * @return array an array of supported attributes, i.e. array('CREATE', 'READ')
+     */
+    protected function getSupportedAttributes()
+    {
+        return array(
+            'VIEW',
```
hairmare: You only seem to test for ``VIEW``. Does the voter disallow other Attributes? I assume we will disallow them in other voters.

src/Graviton/SecurityBundle/README.md
```diff
@@ -59,7 +59,28 @@ The service referenced in the parameter must implement the »\Graviton\RestBundl
...
+Voter
+~~~~~
+- OwnContextVoter
+Acting on either the Account (GravitonDyn\AccountBundle\Document\Account) or the 
+Customer (GravitonDyn\CustomerBundle\Document\Customer) object, this voter determines depending
+on the provided authentication token, if the current user is granted to proceed or not.
+
+Example:
+
+```
```
hairmare: You could fence this as PHP block just for the wow effect. GHM ftw :dancers:

lapistano: it is already in GRV-1419 ;)

src/Graviton/SecurityBundle/Voter/OwnContextVoter.php
lapistano: nope, the voter does only get into action, if the attribute is VIEW .. I thought this would be ok for now.

PR #56, opened 2015-03-05
consultant contact api

narcoticfresh:
This is [WIP]!

In the ConsultantContact API, the goal is to save resources that will later be sent as E-Mails in an async fasion..
I want to use this PR to solve a fundamental problem. Or, as soon as we agree on how to solve it, we'll do that in a different branch and merge it back here to finalize..

### DIFF
--- a/src/Graviton/PersonBundle/Controller/ConsultantContactController.php
+++ b/src/Graviton/PersonBundle/Controller/ConsultantContactController.php
@@ -1,0 +1,17 @@
+<?php
+/** * controller for consultant contact */
+
+namespace Graviton\PersonBundle\Controller;
+
+use Graviton\RestBundle\Controller\RestController;
+
+/**
+ * ConsultantContactController
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class ConsultantContactController extends RestController
+{
+}
--- a/src/Graviton/PersonBundle/Resources/definition/ConsultantContact.json
+++ b/src/Graviton/PersonBundle/Resources/definition/ConsultantContact.json
@@ -1,0 +1,72 @@
+{
+  "id": "ConsultantContact",
+  "description": "An API allowing to contact consultants.",
+  "service": {
+    "readOnly": false,
+    "routerBase": "/person/contact/consultant/",
+    "baseController": "\\Graviton\\PersonBundle\\Controller\\ConsultantContactController",
+    "fixtures": []
+  },
+  "target": {
+    "indexes": [],
+    "relations": [],
+    "fields": [
+      {
+        "name": "id",
+        "type": "varchar",
+        "title": "ID",
+        "description": "Unique identifier for an item."
+      },
+      {
+        "name": "senderName",
+        "type": "varchar",
+        "title": "Sender name",
+        "description": "Name of the contact initiator",
+        "required": true
+      },
+      {
+        "name": "senderEmail",
+        "type": "varchar",
+        "title": "Sender email",
+        "description": "E-Mail of the sender"
+      },
+      {
+        "name": "senderPhone",
+        "type": "varchar",
+        "title": "Sender Phone",
+        "description": "Phone of the sender"
+      },
+      {
+        "name": "comment",
+        "type": "varchar",
+        "title": "Comment",
+        "description": "Comment",
+        "required": true
+      },
+      {
+        "name": "desiredContactDate",
+        "type": "datetime",
+        "title": "Desired Contact Date",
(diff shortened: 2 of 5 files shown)


narcoticfresh:
@hairmare @lapistano 

I need to discuss how we can solve the following:
* Now and then, we'll have fields in a Document (like ConsultantContact) that need custom processing. An example would be the "comment" field that needs to go through HtmlPurifier.. 

Proposed solution
* IMHO the best way to deal with this is to define custom "types" (in this PR I created a Document, that may be the wrong place) like "PurifiedHtml" and use this as the field type inside the Document.
* Then, we can implement a Serializer Handler that properly applies the work on the input on (de)serialization for this Type - in that case, it will take the user input on serialization, pipes it through HtmlPurifier and the process continues.

Advantages
* Using the serializer for these use cases IMHO solves a lot. It's a good place to do it and it allows us to use those types in the service JSON definition directly, having max re-usability.
* It just makes sense - purifiying a user HTML snippet with the serializer is like formatting a DateTime, having a place to store the logic and having the ability to raise (de)SerializerExceptions..

My Problem
* Currently, if (as seen in this PR) I refer to this custom `PurifiedHtml` type, the generator generates Doctrine relations (simple assoc) that will trigger errors in `JsonInput` when checking assoc (as incoming value is (and needs to be) a string). So it seems we'll need to enhance the generator to cover this.

Can we agree to solve this requirement on the serializer level? Are there other ways to do this? 
I think writing custom controllers that loop through the properties are not a valid solution..

If we would do it this way, other types (like `Email`) could be defined easily were we can validate in the serializer and throw an Exception if the email is invalid for example.. It opens a lot of doors..

hairmare:
We're talked about using twig more. With twig I would probably just put a json in the right place and call htmlpurifier there.

I'm not sure about having a loop containing an if in a template though (as thats how I would solve it). My template would use serializer for most fields but filter the field that needs purifying through htmlpurifier first.

The generator might somehow be told to generate a template based on what kind of field it is currently rendering.

Maybe we should look into extending doctrines types rather than keeping HtmlPurifier in Documents.

narcoticfresh:
Did we agree to save the sanitized input right? We won't save the un-purified string and purify at output time? 

I think it's better to do this stuff before saving and save it correctly sanitized - thus Serializer would be the best point to do it..

It would also be quite simple IMHO, but we'll need to generate the doctrine models (xml files for them) a little bit differently..

hairmare:
Your probably right, input it will be.

I looked into them quickly and it seems that using [doctrine custom types](http://doctrine-mongodb-odm.readthedocs.org/en/latest/reference/basic-mapping.html#custom-mapping-types) would be a nice fit for this. 

I found [a blog](http://php-and-symfony.matthiasnoback.nl/2012/09/symfony2-mongodb-odm-creating-custom-types-with-dependencies/) detailing one way to do this in symfony.

doctrine/DoctrineMongoDBBundle#75 tells us that there is no official way (yet).

narcoticfresh:
Well, I would even say we don't have to bother Doctrine at all.. 
In most cases, a type like 'purifiedHtml' or 'email' can just have a Doctrine type of String..

The Document class would have an instance type of the class - but in MongoODM config xml we would just notate it as string.. I think that would function nicely..

IMHO we don't need to bother Doctrine as the resulting value to safe is a primitive, ie a String or int - it's just Serializer that needs to invoke the correct handler in order to get that primitive..

hairmare:
I'm not sure how we validate emails yet. Even they might benefit from a custom type for validation.

narcoticfresh:
IMHO it's the same task as validating a DateTime input from the client and raising throwing a Serialization error if it's not a valid email. With our custom serializer handler we can verify it how we want with a flexible complexity..

narcoticfresh:
@hairmare @lapistano 

I'm still looking where to actually put the purifying code to the fields that need it.. I think my initial approach of using the Serializer is flawed. As it turns out, I could use the Serializer to apply the purify by setting a special Classtype on the field - but then it needs to be a class later (as opposed to a primitive what it should be) as well so the Serializer can serialize it again - this is not what we're looking for..

What we actually need is some kind of 'transformation' (not to be confused with email validation) like supported by Symfony in forms: http://symfony.com/doc/current/cookbook/form/data_transformers.html

I couldn't find how to use them without using Form. The `HTMLPurifierBundle` also registers the purify stuff as Form Transformer.

**So the question remains**
* Before we deserialize user content into our Document, some component needs to be able to transform some fields on the Input depending on some configuration. What component could that be?

@lapistano Do you see any lightweight way of using that Transformer stuff without having to deal too much with Forms? 

(The "How to validate email" is different - here we just need the generator to generate additional constraints IMHO..)

narcoticfresh:
I realized that ConsultantContact doesn't belong in here, as Consultant is also not here.. Will do this in a outside bundle alongside Consultant..

PR #57, opened 2015-03-06
fix install flow

narcoticfresh:
I received the following error after `composer install` and loading the fixtures afterwards:

```
[JMS\Serializer\Exception\RuntimeException]                                 
  You must define a type for GravitonDyn\ModuleBundle\Document\Module::$key.
```

Changing the order of install commands as in this PR fixes this. I'm not sure why, but it does it ;-)
The error currently prevents the fixtures from getting loaded, so we need this fix asap..

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,11 +69,11 @@
             "sh src/Graviton/SwaggerBundle/Resources/scripts/swagger-copy.sh",
             "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
+            "Graviton\\GeneratorBundle\\Composer\\ScriptHandler::cleanDynamicBundles",
+            "Graviton\\GeneratorBundle\\Composer\\ScriptHandler::generateDynamicBundles",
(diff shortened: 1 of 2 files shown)


hairmare:
This works due to the scripts now being after ``Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache``.

narcoticfresh:
Actually, it seems they need to be *before* the `clearCache`. If they're before, the mentioned error orcurrs and fixtures cannot be loaded..(?)

hairmare:
When u see it. This is where I ended up having them in the registry as well. You could in theory do ``Graviton\\GeneratorBundle\\Composer\\ScriptHandler::cleanDynamicBundles`` earlier so ``src/GravitonDyn`` doesn't find it's way into the bootstrap cache on the first ``buildBootstrap `` run (I have it first in the registry and it works).

narcoticfresh:
Ehm, I have no idea what registry u mean.. Anyway, what is wrong with the order as proposed in this PR?

hairmare:
Sorry, I had linked my [smoke-screen registry thing](https://github.com/gravity-platform/registry) in another pr.

PR #59, opened 2015-03-09
remove swagger-ui dependency and repo

hairmare:
Another step at making our composer runs faster. This drops swagger-ui at the graviton level. We will be keeping/re-adding it as needed in offline instances.
For the moment it does not break anything in those but I do plan on removing swagger-copy.sh when no one uses it anymore.


narcoticfresh:
let's merge #61 first, merge develop back in here and see then..

hairmare:
merged and remerged back here, mostly pointless because the only conflict was your removal of swagger-copy.sh.
I'm assuming you're updating wrappers as we speak?

narcoticfresh:
the commands need to be removed also (the new swagger composer calls).. don't see it in the diff..

narcoticfresh:
I will update the wrappers - i just realized that it will pick the wrong `web` path there I assume..

hairmare:
sorry about the command, that must have merged cleanly and I was a bit to fast.

narcoticfresh:
and the lock file needs an update too I suppose as composer will complain..

hairmare:
I feel like there's still something lurking here. Does the generator really write to ``web/``?

narcoticfresh:
Not Generator, but SwaggerCommands..

hairmare:
swagger generator :)

PR #60, opened 2015-03-10
remove container dependency from listeners

hairmare:
This should help a bit with our phpunit performance issues. Some of them seem to stem from the
fact the container builder does multiple passes on ContainerAware things. This gets rid of a small
part of them. It won't make phpunit fast, but it is a step in the right direction.

File this under cleaning up legacy code from the olden days when injecting the container was still
the rage.

### DIFF
--- a/src/Graviton/RestBundle/Resources/config/services.xml
+++ b/src/Graviton/RestBundle/Resources/config/services.xml
@@ -107,18 +107,14 @@
         <!-- Paging listener -->
         <service id="graviton.rest.listener.paginglinkresponselistener"
                  class="%graviton.rest.listener.paginglinkresponselistener.class%">
-            <call method="setRouter">
-                <argument type="service" id="router"></argument>
-            </call>
+            <argument type="service" id="router"></argument>
             <tag name="kernel.event_listener" event="graviton.rest.response" method="onKernelResponse"/>
         </service>
 
         <!-- Self link response listener -->
         <service id="graviton.rest.listener.selflinkresponselistener"
                  class="%graviton.rest.listener.selflinkresponselistener.class%">
-            <call method="setRouter">
-                <argument type="service" id="router"></argument>
-            </call>
+            <argument type="service" id="router"></argument>
             <tag name="kernel.event_listener" event="graviton.rest.response" method="onKernelResponse"/>
         </service>
 
--- a/src/Graviton/RestBundle/Resources/config/services.xml
+++ b/src/Graviton/RestBundle/Resources/config/services.xml
@@ -107,8 +107,8 @@
         <!-- Paging listener -->
         <service id="graviton.rest.listener.paginglinkresponselistener"
                  class="%graviton.rest.listener.paginglinkresponselistener.class%">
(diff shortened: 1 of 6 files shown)


src/Graviton/RestBundle/Resources/config/services.xml
```diff
@@ -116,8 +116,8 @@
         <!-- Self link response listener -->
         <service id="graviton.rest.listener.selflinkresponselistener"
                  class="%graviton.rest.listener.selflinkresponselistener.class%">
-            <call method="setContainer">
```
lapistano: why not:
<argument id="router" type="service" />

instead of using the setter/getter injection?

src/Graviton/SchemaBundle/Resources/config/services.xml
```diff
@@ -9,14 +9,14 @@
   </parameters>
   <services>
     <service id="graviton.schema.listener.schematyperesponse" class="%graviton.schema.listener.schematyperesponse.class%">
-      <call method="setContainer">
-        <argument type="service" id="service_container"></argument>
+      <call method="setRouter">
```
lapistano: why not:
```<argument id="router" type="service" />```

instead of using the setter/getter injection?

File: src/Graviton/SchemaBundle/Resources/config/services.xml [hunk 3]
```diff
@@ -9,14 +9,14 @@
...
+        <argument type="service" id="router"></argument>
       </call>
       <tag name="kernel.event_listener" event="kernel.response" method="onKernelResponse"/>
     </service>
     <service id="graviton.schema.listener.canonicalschemaresponse" class="%graviton.schema.listener.canonicalschemaresponse.class%">
-      <call method="setContainer">
-        <argument type="service" id="service_container"></argument>
+      <call method="setRouter">
```
lapistano: see comment above

src/Graviton/RestBundle/Resources/config/services.xml
hairmare: I mostly prefer setter injection. It's not that it really makes a difference. To me it makes the code more reusable since its easier to make deps optional at later stages.

I don't mind about shortening the xml because it's verbose anyway.

I think the ``$this->getMockBuilder()->disableOriginalContructor()`` kludge in PHPUnit supports me in this.

lapistano: for me setter injection is almost the same as using the container directly, it hides the dependencies of an object.

The ```disableOriginalConstructor()``` method of phpunit's  MockBuilder should only be used when ever necessary - it is a nasty unserialize() hack ;) 
Further when creating a test double for the class (at least for me) it does not matter to provide the doubles via constructor or setter. 
I still vote for constructor injection ;)

hairmare: I see, makes sense, change pushed.

lapistano:
[~hairmare] is this ready to be merged?

hairmare:
@lapistano yupp it is

PR #61, opened 2015-03-10
swagger commands

narcoticfresh:
contains

* create a base class for composer scripthandlers
* use that as base class for both swagger and generator scripthandlers
* implement symfony commands to copy swagger-ui to web & generate swagger.json
* use those in scripthandlers
* delete controller and views in SwaggerBundle as we serve it static now


narcoticfresh:
A reminder: I had to remove `HTTP_HOST` (root element `host`) from swagger.json in order to generate it like this. I checked swagger spec

> The host (name or ip) serving the API. This MUST be the host only and does not include the scheme nor sub-paths. It MAY include a port. **If the host is not included, the host serving the documentation is to be used (including the port).** The host does not support path templating.

And indeed, swagger-ui has no problem and it works as expected..

src/Graviton/CoreBundle/Composer/ScriptHandlerBase.php
```diff
@@ -0,0 +1,98 @@
...
+     * Executes a command
+     *
+     * @param CommandEvent $event      Event
+     * @param string       $consoleDir Console dir
+     * @param string       $cmd        Command
+     * @param int          $timeout    Timeout
+     *
+     * @return void
+     */
+    protected static function executeCommand(CommandEvent $event, $consoleDir, $cmd, $timeout = 300)
```
hairmare: This method would actually warrant test don't you think?

src/Graviton/CoreBundle/Composer/ScriptHandlerBase.php
```diff
@@ -0,0 +1,98 @@
...
+
+
+    /**
+     * Finds the path to the php executable
+     *
+     * @param bool $includeArgs include args
+     *
+     * @return bool|false|null|string path
+     */
+    protected static function getPhp($includeArgs = true)
```
hairmare: This one as well.

src/Graviton/SwaggerBundle/Service/Swagger.php
```diff
@@ -191,7 +191,6 @@ private function getBasicStructure()
             'title' => 'Graviton REST Services',
             'description' => 'Testable API Documentation of this Graviton instance.'
         );
-        $ret['host'] = $_SERVER['HTTP_HOST'];
```
hairmare: I think you've hit gold on this one.

src/Graviton/SwaggerBundle/Resources/views/Main/index.json.twig
```diff
@@ -1 +0,0 @@
-{% extends 'base.json.twig' %}
```
hairmare: Why doesn't the generator use twig anymore?

narcoticfresh: Hm? This is SwaggerBundle and that hasn't got any Controllers anymore.. So it needs no view..

src/Graviton/CoreBundle/Composer/ScriptHandlerBase.php
narcoticfresh: All those functions are copies from SensioDistributionBundle (https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Composer/ScriptHandler.php)

If you want to test that again, let's open an separate issue for that

src/Graviton/SwaggerBundle/Resources/views/Main/index.json.twig
hairmare: templates can be used to generate more than just views for http in controller. A lot of the resource generators do it that way.

src/Graviton/CoreBundle/Composer/ScriptHandlerBase.php
hairmare: And they are just as crappy over there ;)

Maybe we should just start phasing out the distribution bundle and replace these with something external that is meant to be used in this fashion. I'm sure someone already wrote these as a composer package.

Copying an internal API does not make an excuse not to test it locally, but I'm not blocking this pr on these grounds.

src/Graviton/SwaggerBundle/Resources/views/Main/index.json.twig
hairmare: I'm ok with not using twig on this one but would very much like to use a proper serialization component on this one, I think I've got a story on that already though.

src/Graviton/CoreBundle/Composer/ScriptHandlerBase.php
hairmare: I was also mentioning this since this is the first time that these methods are being used by a subclass and therefore expose an internal API that, in theory, needs testing.

I'm letting it go due to the fact that these get executed on each travis build anyway and will fail early and hard if they really have issues.

PR #63, opened 2015-03-10
swagger dist path

narcoticfresh:
The path to the `vendor` dir differs in wrapper context, this fixes that..

AFAIK there is no programmatic way to know if we're the main package or a dependency, so this is a no-fuss way..


hairmare:
I don't really like it but I'm merging it. At this point we're almost so far that we could move this to the general wrapper btw.

hairmare:
Also this should probably just be an option to the command?

narcoticfresh:
Well, IMHO making it an option would just move the `strpos`-like check to the Composer ScriptHandler, not making it nicer.. We need an API to know if we're inside a vendor or not to make it really better..

PR #64, opened 2015-03-10
return cors header for swagger.json

hairmare:
Now that we host it for itself we need to make it reachable from https://swagger.nova.scapp.io/.

This fixes that for cf and will need adding to all wrappers that target cf. We still need to solve it for wrappers that push elsewhere (ie. docker et al).

### DIFF
--- a/.bp-config/nginx/server-locations.conf
+++ b/.bp-config/nginx/server-locations.conf
@@ -19,6 +19,10 @@ location / {
     try_files $uri $uri /app_dev.php?$args;
 }
 
+location /swagger.json {
+    add_header 'Access-Control-Allow-Origin' "*";
+}
+
 location ~* \.php$ {
     include         fastcgi_params;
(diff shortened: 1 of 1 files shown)


.bp-config/nginx/server-locations.conf
```diff
@@ -19,6 +19,10 @@ location / {
     try_files $uri $uri /app_dev.php?$args;
 }
 
+location /swagger.json {
+    add_header 'Access-Control-Allow-Origin' "*";
```
lapistano: is it wise to allow the cross side access to any domain?  Can't we be more specific on this one?

hairmare: We probably can an should also. Since this is only for ``swagger.json`` and we don't have a way to store a whitelist yet I'm doing it like this. When graviton learns to store a whitelist we should be able to integrate that here as well.

PR #65, opened 2015-03-11
fix --no-dev composer install cases

hairmare:
Also fixes wrapper cases that don't themselves have phpcs installed.

I'm doing it like so since that was recommended on the composer issue
that talked about adding a "scripts-dev" node to composer.json. I can't
seem to find that again though.

This also fixes cloud pushes directly from the graviton repo and I use
those lots for dev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,4 @@ php:
 script:
   - ./vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
   - php ocular.phar code-coverage:upload --format=php-clover coverage.clover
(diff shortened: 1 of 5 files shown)


narcoticfresh:
so as said; please move this call out of composer.json into travis and dev-check.sh..

hairmare:
I moved everything to ``composer check``. ``dev-check.sh`` is still here for compat reasons but will fail if you don't have a global composer install.

narcoticfresh:
nice!

narcoticfresh:
i'll wait for travis result to be sure and then merge..

PR #66, opened 2015-03-12
bump to php 5.6

hairmare:
I add 5.6 to ``.travis.yml`` and switch our cloudfoundry deploys to 5.6 as well. I'm doing this because we will be going 5.6 anyway when we start using docker due to php:latest being 5.6 on docker (which is a good thing).


hairmare:
oops, restored

hairmare:
I talked to @narcoticfresh and we feel dropping 54 is ok. Merge if you're ok with that too.

hairmare:
weren't we merging this yesterday evening :trollface:

PR #67, opened 2015-03-12
Docker support

hairmare:
This is a quick summary of how I intend our production docker setup to work.

It still needs some work and I will probably end up creating some wrappers to ``php:fpm`` and ``nginx:latest`` that configure those correctly.

* [x] check/add mongodb support in php-fpm (pending gravity-platform/dockerfiles#4)
* [x] set up nginx so if uses fpm from the link
* [x] link a mongodb to ``graviton-fpm`` and ~~consume ``$DOCKER_*`` env vars~~ use mongodb hostname in container
* [x] add ~~tooling~~ docs on how to create and publish vendorized (aka offline) images
* [x] verify that cf still works


hairmare:
I found out that this does not work with docker-compose due to the fact that there seems to be no volume-only container support in it. (running ``docker run graviton/graviton noop`` fails).

I'm thinking about switching to ``FROM busybox`` but that would still only solve the problem in a half assed way. I'm not really happy with having to go down either the ``FROM composer/composer`` or ``FROM php:fpm`` route.

I also noticed that I'm really close to docker-library/php#54 with my solution. It seems that we are not the only ones with this issue. It does seem like I'm the only one that wants to use ``composer/composer`` in the fashion I describe here together with ``php:fpm`` though.

hairmare:
@libgraviton/hackers @libgraviton/owners This is ready for review :D

lapistano:
pls rebase @hairmare

hairmare:
merged and rebuilt composer.lock

PR #68, opened 2015-03-13
[WIP] Feature/grv 1416 doctrine deny all filter

lapistano:
will break the tests, due to missing security layer for test env » implementation will follow


lapistano:
The ability to functional test controllers after introducing the authorization deciders depends on the ability to stub dynamic generated bundles (contract and related). Will do that in a different PR.

lapistano:
way too old, needs rethinking and a whole new implementation anyways

PR #70, opened 2015-03-16
take collection name from parameters.. fixes #69

### LINKED ISSUE #69: DynamicBundleGenerateCommand: Get mongodb name from config
Mongo db name in Dynamic Bundle generator must be get from Parameter, at the moment it's hard coded..

narcoticfresh:
discovered this upon installing it with different config..

### DIFF
--- a/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
@@ -13,21 +13,20 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\Output;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\DependencyInjection\Container;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\Process\Process;
 
 /**
  * Here, we generate all "dynamic" Graviton bundles..
  *
- * @todo create a new Application in-situ
- * @todo see if we can get rid of container dependency..
+ * @todo     create a new Application in-situ
+ * @todo     see if we can get rid of container dependency..
  *
  * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
  * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
  * @link     http://swisscom.ch
  */
-class GenerateDynamicBundleCommand extends Command implements ContainerAwareInterface
+class GenerateDynamicBundleCommand extends Command
 {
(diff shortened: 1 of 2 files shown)


lapistano:
I guess it is no big deal to either add the 'ContainerAwareInterface' as implemented interface or use the most preferred option and define the command as a service and constructor inject the container. 

And yes the setter and getter for the container was not introduced with this PR, but I'd really like to see this fixed.

narcoticfresh:
@lapistano 
IMHO this is already done - the Container gets injected by the DIC(?)

services.xml:

```xml
<service id="graviton_generator.command.generatedynamicbundle"
         class="Graviton\GeneratorBundle\Command\GenerateDynamicBundleCommand">
    <call method="setContainer">
        <argument type="service" id="graviton_generator.command.generatedynamicbundle.container"/>
    </call>
    <call method="setProcess">
        <argument type="service" id="graviton_generator.command.generatedynamicbundle.process"/>
    </call>
    <tag name="console.command" />
</service>
```

That why it only extends from `Command`

lapistano:
@narcoticfresh correct the container was injected. But the container is the DIC here and it is injected by a method called setter/getter injection. 
My point is the fist implementation the approach was not done properly. 
I am not judging and am not saying that you did something wrong. 
I am just pointing out errors and want to get rid of them.

So what was sloppy in the first implementation:
- no checks if the process member is actually set before use
- the setContainer() neither had a type hint, nor a validation the correct object was passed 
- getContainer() did not check, if the property was actually set 

To give an example how to do this approach correct I implemented https://github.com/libgraviton/graviton/commit/2f71431219e71fd9d9fb35d20c6faafdebeb1306
Another drawback of using Setter/getter injection is not only that the service definition gets unnecessarily more complex, but it hides dependencies of the class to the first view. I have to read the complete code to find them.

The 2nd implementation  (https://github.com/libgraviton/graviton/commit/6a4ae74658a216eb97056484294492dea4dec314) uses constructor injection. This makes it more obvious what the dependencies of a class are. The benefit of that is that I do not have to read the complete code to find them, but also have an indicator if there are too many dependencies and a splitting might be a good idea.

narcoticfresh:
OK; acknowledged.. 

Personally, I just think we need to have a 'standard', a concensus how to implement it in Graviton.. Thats the maybe 5th I had to change stuff from either feedback from you or @hairmare .. Whoever reviews it, the different the feedback is.. This makes it hard to get my PRs through in time..

Many ways lead to Rome as we know, so could you two please agree on a way and document the 'standard way to do it' somewhere, so we know how it should be done in the future..

I personally prefer the Constructor injection as it's most clear and leads to less setters and getters, less cluttering in general..

So thanks for the rewrite and input; this can be merged now IMHO.. I will use Constructor injection (if that's the agreed way now) in the future..

PR #71, opened 2015-03-17
Feature/make generator aware of entity interface

lapistano:
to be able to make the security component testable and documents more reliable the ability to define a set of interfaces to be recognized when a document is created.

### DIFF
--- a/dev-cleanstart.sh
+++ b/dev-cleanstart.sh
@@ -1,6 +1,3 @@
-export SYMFONY__GENERATOR__DYNAMICBUNDLES__MONGOCOLLECTION="LoadConfig"
-export SYMFONY__GENERATOR__DYNAMICBUNDLES__MONGOCOLLECTION__CRITERIA='{"loadPackage":"szkb-wcp"}'
-
 rm -Rf app/cache app/logs app/bootstrap.php.cache
 ./composer.phar install
 php app/console doctrine:mongodb:fixtures:load
--- a/dev-cleanstart.sh
+++ b/dev-cleanstart.sh
@@ -1,4 +1,7 @@
+export SYMFONY__GENERATOR__DYNAMICBUNDLES__MONGOCOLLECTION="LoadConfig"
+export SYMFONY__GENERATOR__DYNAMICBUNDLES__MONGOCOLLECTION__CRITERIA='{"loadPackage":"szkb-wcp"}'
+
 rm -Rf app/cache app/logs app/bootstrap.php.cache
 ./composer.phar install
 php app/console doctrine:mongodb:fixtures:load
-php app/console server:run
+#php app/console server:run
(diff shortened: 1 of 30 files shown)


narcoticfresh:
Hmm.. this actually contains a lot more then it says it does..
RQL Bundle changed, Auth stuff, Interface.. hard to see if it breaks nothing..

If that is merged, what actually changes in behavior as opposed to current `develop`?

dev-cleanstart.sh
```diff
@@ -1,4 +1,7 @@
+export SYMFONY__GENERATOR__DYNAMICBUNDLES__MONGOCOLLECTION="LoadConfig"
+export SYMFONY__GENERATOR__DYNAMICBUNDLES__MONGOCOLLECTION__CRITERIA='{"loadPackage":"szkb-wcp"}'
+
```
narcoticfresh: this shouldn't be here (environment variables)

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -73,38 +73,38 @@ public function setRepository(ObjectRepository $countries)
...
         if (!empty($filter)) {
             // set filtering attributes on request
             $request->attributes->set('filtering', true);
 
-            $queryParser = new Query(urldecode($filter));
-            $queriable = new MongoOdm($this->repository, $numberPerPage, $startAt);
-            $queriable = $queryParser->applyToQueriable($queriable);
-            $records = $queriable->getDocuments();
+            // define offset
+            $queryBuilder->skip($startAt);
```
narcoticfresh: It seems, `queryBuilder->skip()` only gets called when a filter is present -> should be called either way..

src/Graviton/CoreBundle/Document/GravitonShowcaseInterface.php
```diff
@@ -0,0 +1,34 @@
...
+    public function getContact();
+    public function setContact(\Graviton\PersonBundle\Document\PersonContact $contact);
+    public function getContacts();
+    public function setContacts($contacts);
+    public function getContactcode();
+    public function setContactcode(\GravitonDyn\ShowcaseBundle\Document\ShowcaseContactCode $contactCode);
+    public function getNestedarray();
+    public function setNestedarray($array);
+    public function getUnstructedobject();
+    public function setUnstructedobject($object);
```
narcoticfresh: this interface shouldn't be here..

narcoticfresh:
As said - IMHO most OK (even if it mixes a lot). We'll leave that PR on hold as for now, so we can merge everything except that concerning the Interface problems of generated Documents. That we'll need to solve separately. As said, Showcase Entities should always be present in the PHPUnit run so they should be testable without Interfaces.

Once Airlock is in a seperate bundle; we can test that stuff in the wrapper, where all necessary bundles are in place and available..

lapistano:
this one is too bis .. will be spit

PR #72, opened 2015-03-17
Release v0.7.0

hairmare:
***Do not merge on github!!!!***

We need to start issuing regular releases again. This one is awfully large and thus not really handy to document.

To merge this you should do a local checkout of the branch and then ``git flow release finish`` it.

Since this is the first release I'm doing it as a PR. This is so we can discuss it and how we would like to do releases going further.

~~I propose the following:~~
* ~~we start the ``releases/<next>`` branch early~~
* ~~we then just add to its changelog during the sprint without merging or rebasing anything on the release branch~~
* ~~at release time the branch gets squashed and finished using ``git flow``~~
* ~~we then use a pr like this one for voting on finalizing a release (using :+1: and :-1:)~~
* ~~this pr is created and closed before our sprint demo (early enough so CD has a chance to run through)~~
* ~~If noone of the contributors to the release respond with a :-1: the release is merged by whoever created the release branch~~
* ~~if anyone responds with :-1: he should explain why and we'll take if from there~~
* ~~every contributor to a sprint is expected to :+1: , if someone doesn't we'll ignore that on a case-to-case base (ie. vacation, ...)~~

So basically, if all the devs that where active in the last 2 weeks :+1: this, I'll just merge it. For the next release, we'll have a bit more to do since there will (hopefully) be some serious content in CHANGELOG.md.

tldr: I propose a release strategy, if you don't give a flying duck about releases answer with ``:+1:`` 

edit: my proposal was crap, we should use github milestones and define a releasor on duty for each release. We should also kill CHANGELOG.md with fire, but I'm not changing that in this release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 0.7.0
+
+This release concludes some months of development without release. It contains heaps
+of backwards incompatible changes. In future we will be doing releases more often.
+
+Trying to summarize this release contains much work in the following areas:
+
+* dynamic-bundle support that we use to generate graviton-based apps
+* a lot of infrastructure work
+* countless fixes and changes
+* much very important stuff
+
 ## 0.6.0
 
 * full i18n support
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "graviton/graviton",
-    "version": "0.7.0-dev",
+    "version": "0.7.0",
     "license": "GPL",
     "type": "project",
     "description": "The base package for graviton",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "54774f986d3b84c34663a834bd9bf2af",
+    "hash": "79c40e7967c118571cf97c0ff8a5f2bd",
     "packages": [
         {
             "name": "behat/transliterator",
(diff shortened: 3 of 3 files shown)


lapistano:
+1 for getting rid of the changelog file.
but I am not sure about the release strategy. I am on the same page when it comes to having a releasor, but I do not get why mixing GH Milestones and git flow releases. Am I missing something? 

Afair GH Milestones come in handy when you are managing your project in GH and using the GH issue tracker to handle Sprints, bug tracking, etc. A GH Milestone is then a container for a defined set of issues forming a sprint. 
In our case an external Issue tracker is mainly used and we somewhen decided to follow the gtiflow-Workflow. This already defines how to handle releases, doesn't it?

hairmare:
Naturally, I would like to keep the gh milestones and internal issue tracker releases in sync somehow. For my point of view gh milestones track every closed pr thats part of a tagged release. tracker releases track the same thing for tracker issues. I think this is needed due to them not always being mappable.

Talking about this kinda leads down the rabbit hole of having the wat do talk about the gh issue tracker in general. I don't think we should deactivate gh issues fully due to them being a major contact point with the community at large. There is the approach of getting a cf-bot to sync gh issues and prs with our internal tracker but I don't feel like that makes sense. 

I think we really need integration for all of this to work in the long run but need to somehow populate (and thus sync) the gh things for proper transparency.

Regarding git-flow (actually github-flow) it is mentionable that milestones are quite integrated with tagging releases the way we do (the names need to match) which is handy.

Should we use https://github.com/libgraviton/graviton/tags instead of the CHANGELOG?

lapistano:
I never thought about disabling the GH Issue tracker, since it is as essential for the communication with the OS community as the GH wiki for documentation. 

And yes let's talk about the sync and stuff vis-a-vis.

PR #73, opened 2015-03-19
added ability to provide a wrapper specific service whitelist

narcoticfresh:
Now, please consider my thoughts when looking into this.. it's a no-brainer actually ;-)

**What we need**
In specific Graviton wrappers we need to be able to exclude certain services from generating a controller.  In ml3k context, we do this by overriding the JSON file (inheritance there) and setting the `service.routerBase` to `null`. This here adds the same ability to Graviton, but obviously in a different fashion as we don't have JSON file inheritance here.

It's crucial we generate all `Documents` from *all* available JSON files - but *don't* generate a `Controller` to it as those `Document` classes are needed in logic controllers to do their work. See elaboration below.

**What solutions I considered**
Now, why is this an environmental setting influencing the Generator? I considered and discarded the following approaches
* Splitting JSON files in Bundles<br />Actually, I always thought we can solve this by re-arranging the JSON files in the bundle. So I would make a "parameter" bundle containing only parameter services (those that should not be generated on the external-facing wrapper) and one "logic" bundle containing services that should be generated on both wrappers.<br />This is actually not feasible as doing this would lead to the external facing wrapper *missing documents* (i.e. the documents having the configuration) and the wrapper would not be able to find it's required `Document` classes to implement his logic (i.e. `MortgageCost` service not being able to use `FinancingUseCase` documents) 
* Using ACL<br />This would be the wrong approach is the `Controller` endpoint would still exist, it would be listed in `MainController` (front listing all services) but then only on GET/POST it would be denied. This *would* work - I know - but it's more secure to not even generate `Controller`s to services we don't want to provide anyway in a certain wrapper.

### DIFF
--- a/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
+++ b/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
@@ -230,7 +230,7 @@ class GenerateDynamicBundleCommand extends Command
             );
 
             // controller?
-            if (!$jsonDef->hasController() || $this->isBlacklistedController($jsonDef->getRouterBase())) {
+            if (!$jsonDef->hasController() || $this->isNotWhitelistedController($jsonDef->getRouterBase())) {
                 $arguments['--no-controller'] = 'true';
             }
 
@@ -516,6 +516,7 @@ class GenerateDynamicBundleCommand extends Command
 
     /**
      * Checks an optional environment setting if this $routerBase is whitelisted there.
+     * If something is 'not whitelisted' (return true) means that the controller should not be generated.
      * This serves as a lowlevel possibility to disable the generation of certain controllers.
      * If we have no whitelist defined, we consider that all services should be generated (default).
      *
@@ -523,14 +524,14 @@ class GenerateDynamicBundleCommand extends Command
      *
      * @return bool true if yes, false if not
      */
-    private function isBlacklistedController($routerBase)
+    private function isNotWhitelistedController($routerBase)
     {
-        // if no whitelist is set, we're generating everything..
+        // if no whitelist is set, everything is whitelisted
         if (!$this->container->hasParameter('generator.dynamicbundles.service.whitelist')) {
             return false;
         }
 
-        // if param is there, default is 'yes' - everything is blacklisted..
+        // if param is there our default is 'yes' - everything is not whitelisted by default.
         $ret = true;
 
         $whitelist = json_decode(
@@ -538,7 +539,7 @@ class GenerateDynamicBundleCommand extends Command
             true
         );
 
-        // if in whitelist, it's not blacklisted ;-)
+        // whitelist it if in list..
         if (is_array($whitelist) && in_array($routerBase, $whitelist)) {
             $ret = false;
         }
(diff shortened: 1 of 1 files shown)


src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -513,4 +513,36 @@ private function determineSearchCriteria()
...
+    /**
+     * Checks an optional environment setting if this $routerBase is whitelisted there.
+     * This serves as a lowlevel possibility to disable the generation of certain controllers.
+     * If we have no whitelist defined, we consider that all services should be generated (default).
+     *
+     * @param string $routerBase router base
+     *
+     * @return bool true if yes, false if not
+     */
+    private function isBlacklistedController($routerBase)
```
lapistano: the notion is correct the naming is a bit off... 
the method name isBlacklistedController() implies there is an actual blacklist which is not available. I'd go for 
isWhitelistedController() or isNotWhitelistedController() to make it more clear that the mechanism is base on white- and not blacklisting.

src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php
```diff
@@ -513,4 +513,36 @@ private function determineSearchCriteria()
...
+
+        // if param is there, default is 'yes' - everything is blacklisted..
+        $ret = true;
+
+        $whitelist = json_decode(
+            $this->container->getParameter('generator.dynamicbundles.service.whitelist', array()),
+            true
+        );
+
+        // if in whitelist, it's not blacklisted ;-)
```
lapistano: this is what I mean .. there is a comment necessary to make it clear what happens ;)

narcoticfresh:
OK; I inverted the naming and changed the comments to reflect it..

Actually, 'not whitelisted' as the same as 'being blacklisted' - but I see the point that it implies the existence of a blacklist (which of course; doesn't exist).. it only exists implicitly..

lapistano:
am waiting 4 the CI jobs to finish. the rest looks good 4 me

hairmare:
This is great. I actually need this and already had it somewhere in the back of my head to do exactly this.

I'm absolutely unhappy about using json_decode and feel we should stop using that anti-pattern for injecting stuff. The config on this should be done with a proper yml file in graviton-standard-edition and not by injecting ENV.

I'm quite sure we already have a story on the yml thing somewhere.

narcoticfresh:
I agree - it gets quite unreadable when passing big structures via ENV this way.. 

A yaml would be nice - but it needs to be possible to have a yaml file containing params that override params *and* are outside of `vendor` and outside the project root..

hairmare:
Basically the wrappers need to contain a ``app/config/`` structure that may be used to override whatever is already in ``vendor/graviton/graviton/app/config``. This should support ``parameters.yml-dist`` and maybe also ``services.yml``. As said, we have a [story](https://issue.swisscom.ch/browse/GRV-1456) for the parameters part of this.

hairmare:
Some docs would have been nice on this, I can't even seem to find an example internally :(

hairmare:
something ala in the GeneratorBundle Readme or so.
```bash
export SYMFONY__GENERATOR__DYNAMICBUNDLES__SERVICE__WHITELIST='["\/bundle\/service"]'
```

PR #74, opened 2015-03-19
move fixed services to json def

narcoticfresh:
We should generate all front-facing services via JSON defs..

So this PR serves as a primer - we move `/app/core` to JSON def and see if travis can test our result correctly (as `AppControllerTest` now tests the generated `Controller`)..

If this succeeds, we can move `Product` as well.. And once established we can write a Test for `Showcase.json` which actually tests all capabilities of our Generator and serves as an excellent integration test for the Generator (IMHO)..


narcoticfresh:
@hairmare What do you think about this, should we pursue this right?
Seems to be working fine with /app/core..

hairmare:
@narcoticfresh we certainly need to pursue this. Using showcase.json to test the generator as unit makes sense to me.

If I recall correctly we could just as well kill ``Product``. I don't think it's being used anywhere and i think everything testable in it is already covered by ``Country``. The ``Product`` thing was originally intend to be a place where we load product info into and that has been solved with ml3k and the loadconfigs anyhow.

narcoticfresh:
@hairmare 
OK, good.. I struggle with porting `Language` - as it's referenced as dependency by a Listener, it cannot be moved to `GravitonDyn` it seems.. But alas, we'll leave it..

--> What about `/entity/country/` - can we ditch that too? Or port it to JSON def?

Thing is - whatever is not a JSON def (as `Country`) cannot be hidden by the whitelist or anything - it will be always visible in every instance we deploy..

hairmare:
I feel like leaving the ``Language`` stuff in Graviton makes sense. We should start using it more. In theory we should be able to keep translations from loads in there as well.

We can get rid of ``Country`` if we need to. The tests written to cover ``Country`` need to be changed to reflect that and Country must be replaced somehow. Keep in mind that Country is mainly used as a test-case for read-only material. It also showcases a way to load data from a third party (the world bank in that case).

I know most of this is done better elsewhere nowadays. This needs work due to missing tests throughout large parts of the new codebase. I need the /core/app and /entity/country services to be reliable since I use them for manual testing and sounding of ideas all the time. Right now I can't use the generator stuff for that since the untested areas of it are still to vast. They also cover some use-cases that are still not possible with generators (upserting for instance).

hairmare:
@narcoticfresh I think this still had some changes with merit. Do plan on opening something similar again?

PR #78, opened 2015-03-24
Feature/grv 1415 roles2json

lapistano:
Add role definition to JSON configuration files.

### DIFF
--- a/src/Graviton/GeneratorBundle/Generator/ResourceGenerator.php
+++ b/src/Graviton/GeneratorBundle/Generator/ResourceGenerator.php
@@ -5,6 +5,7 @@
 
 namespace Graviton\GeneratorBundle\Generator;
 
+use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Inflector\Inflector;
 use Graviton\GeneratorBundle\Definition\DefinitionElementInterface;
 use Graviton\GeneratorBundle\Definition\JsonDefinition;
@@ -44,13 +45,17 @@ class ResourceGenerator extends AbstractGenerator
      * @private
      */
     private $input;
-
     /**
      * our json file definition
      *
      * @var JsonDefinition
      */
     private $json = false;
+    /** @var ArrayCollection */
+    protected $xmlParameters;
+    /** @var \DomDocument */
+    private $serviceDOM;
+
 
     /**
      * Instantiates generator object
@@ -68,6 +73,7 @@ class ResourceGenerator extends AbstractGenerator
         $this->filesystem = $filesystem;
         $this->doctrine = $doctrine;
         $this->kernel = $kernel;
+        $this->xmlParameters = new ArrayCollection();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Graviton/GeneratorBundle/Tests/Generator/ResourceGeneratorTest.php
```diff
@@ -65,14 +65,24 @@ public function tearDown()
      */
     public function testGenerateDocument($base)
     {
-        $servicesMock = $this->getMockBuilder("\DOMDocument")
+        $servicesMock = $this->getMockBuilder("\\DOMDocument")
```
hairmare: Y U no single-quotes? ;)

src/Graviton/GeneratorBundle/Tests/Generator/ResourceGeneratorTest.php
```diff
@@ -65,14 +65,24 @@ public function tearDown()
...
             ->setMethods(array("saveXml"))
             ->getMock();
 
+        $jsonDefinitionMock = $this->getMockBuilder("\\Graviton\\GeneratorBundle\\Definition\\JsonDefinition")
```
hairmare: ``'`` instead of ``"``

src/Graviton/GeneratorBundle/Tests/Generator/ResourceGeneratorTest.php
```diff
@@ -81,7 +91,7 @@ public function testGenerateDocument($base)
         $dir = self::GRAVITON_TMP_DIR;
         $document = "DocumentTest";
 
-        $generator = $this->getMockBuilder("Graviton\GeneratorBundle\Tests\Generator\ResourceGeneratorProxy")
+        $generator = $this->getMockBuilder("Graviton\\GeneratorBundle\\Tests\\Generator\\ResourceGeneratorProxy")
```
hairmare: ``'`` instead of ``"`` would make the need for ``\\`` go away afaik

narcoticfresh:
@lapistano OK; the main integration is done.. how come it doesn't get generated somewhere (twig templates)? Are you planning on using the `JsonDefinition` classes directly to get the roles?

IMHO the `JsonDefinition` classes are used only on generation time. Whatever data we need should be generated somewhere (ie as params or whatever) in order to retrieve it through Symfony, not using the Definition classes..

src/Graviton/GeneratorBundle/Generator/ResourceGenerator.php
```diff
@@ -331,6 +326,59 @@ protected function addParam(\DOMDocument $dom, $key, $value)
...
+     *
+     * @param \DomDocument   $dom            services.xml document to be modified.
+     * @param JsonDefinition $jsonDefinition json configuration, read from resources dir.
+     * @param string         $docName        Name of the document the parameter shall be generated for.
+     *
+     * @return \DomDocument
+     *
+     * @link http://symfony.com/doc/current/book/service_container.html#array-parameters
+     */
+    protected function addRolesParameter(\DomDocument $dom, JsonDefinition $jsonDefinition, $docName)
```
narcoticfresh: I think it's too specific too do the XML right here..

IMHO this one should fill a `$this->xmlParameter` (i.e.) array. Later, this array (in an own function) should be generated into the xml structure.

That way, we can add new params into the xml just by adding a new setter function that also extends the same array and it will land in the xml without further work. It's quite possible we will add more params into the xml in the future..

PR #79, opened 2015-03-25
support outputting referenced documents as JSON-Reference

hairmare:
This isn't quite the fork of schmittjoh/serializer we had expected (yet?). While there are still some kinks to be worked out, It looks to me like a SerializerHandler is the way to go.

This works for pushing out mongodb-references as $ref (see ``/core/module`` for an example). There is still some hardcoded stuff and magic in it that I expect to remove over the next few pushes by refactoring it into sane stuff like container tags and configuration.

I've added some tasks and will also mark up some stuff in the code for further discussion as I proceed.

### Tasks

* [ ] don't depend on the magic field name ``ref`` by making the name configurable through a type somehow
* [ ] make the fixture generator part much smarter (the cheap attempts to do stuff need to go)
* [ ] figure out if the fixtures code has a better way of specifying refs
* [ ] finish writing tests (most are marked as incomplete and only one is partially implemented)
* [ ] support this in swagger as well

### Issues

These are issues that we should tackle at some point, that I'm not standing in between us getting this feature done.

* Lazy-loading of odm objects. Currently the linked object is getting fully loaded from the odm. This does not make sense since we don't use anything from it other than the id that was used to load it. While we should fix this at some point, I don't currently see the clean way to do so. I'm also betting that the penalty we suffer when loading the object isn't too large.


src/Graviton/GeneratorBundle/Resources/skeleton/fixtures/LoadFixtures.php.twig
```diff
@@ -76,6 +76,16 @@ class Load{{ document }}Data extends AbstractFixture implements OrderedFixtureIn
                             $thisRecord->$addFunction($this->getReference($subval->$refName.'-'.$subval->$idName));
                         }
                     }
+
+                    // just as cheap attempt at objects
+                    if (is_object($val) && isset($val->ref) && isset($val->ref->{'$ref'})) {
```
hairmare: I need to get both ``->ref`` and ``->{'$ref`}`` out of here somehow (config). I'm not 100% sure how since ``->ref`` might be an object that is also used for embedding purposes elsewhere.

src/Graviton/GeneratorBundle/Resources/skeleton/fixtures/LoadFixtures.php.twig
```diff
@@ -76,6 +76,16 @@ class Load{{ document }}Data extends AbstractFixture implements OrderedFixtureIn
...
+                        $setterFunction = 'set'.ucfirst($key);
+
+                        $valueRecord = new \GravitonDyn\ModuleBundle\Document\ModuleApp;
```
hairmare: These classname need to be inferred from somewhere during generation, but this whole method  they are in need heavy refactoring anyhow.

src/Graviton/RestBundle/Subscriber/ReferenceHandlerEvent.php
```diff
@@ -0,0 +1,71 @@
...
+     * @param object                   $document document to serialize
+     *
+     * @return object
+     */
+    public function serialize(JsonSerializationVisitor $visitor, $document)
+    {
+        $id = $document->getRef()->getId();
+
+        $link = $this->router->generate(
+            'graviton.core.rest.app.get',
```
hairmare: This needs to be inferred from elsewhere somehow.

src/Graviton/RestBundle/Subscriber/ReferenceHandlerEvent.php
```diff
@@ -0,0 +1,71 @@
...
+            array(
+                'id' => $id,
+            ),
+            true
+        );
+        return array(
+            'id' => $id,
+            '$ref' => $link,
+        );
```
hairmare: I think we should settle on keeping links everywhere in graviton in this format.
```js
"app": {
  "id": "tablet",
  "$ref": "https://..."
}
```

I also think we should support writing these object either as above or as follows.

```js
"app": {
  "$ref": "https://..."
}
```

src/Graviton/RestBundle/Subscriber/ReferenceHandlerEvent.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @return array
+     */
+    public static function getSubscribingMethods()
+    {
+        return array(
+            array (
+                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
+                'type' => 'GravitonDyn\ModuleBundle\Document\ModuleApp',
```
hairmare: I need to check if interfaces work here and add one.

hairmare: It doesn't seem to :( I'll have to do build a list of types for here by tagging services and using a compilerpass... Lets hope those documents already have a service definition.

src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -0,0 +1,473 @@
...
+        $results = $client->getResults();
+
+        $this->assertResponseContentType(self::COLLECTION_TYPE, $response);
+
+        $this->assertEquals(5, count($results));
+
+        $this->assertEquals('tablet', $results[4]->app->id);
+        $this->assertEquals('http://localhost/core/app/tablet', $results[4]->app->{'$ref'});
+
+        $this->markTestIncomplete();
```
hairmare: needs to finished, I won't be adding much more than above and I'll remove most of below for now though.

src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
```diff
@@ -315,4 +315,20 @@ public function setRelType($relType)
     {
         $this->relType = $relType;
     }
+
+    /**
+     * get tags for a class
+     *
+     * @return array
+     */
+    public function getTags()
```
hairmare: I'm pretty sure I'm not using it and should remove it b4 we merge.

hairmare:
This is superseded by #85 and will be closed as soon as we merge that.

hairmare:
I'm closing this for now since i believe in #85. I'm not gonna delete the branch yet tho.

PR #80, opened 2015-03-26
report malformed json

narcoticfresh:
Currently, if a client posts malformed json to a service, he gets (sadly as HTML currently) the following:

```
Argument 1 passed to Graviton\RestBundle\Validator\JsonInput::checkDocument() must be of the type array, null given, called in /home/dn/git/graviton2/src/Graviton/RestBundle/Validator/JsonInput.php on line 79 and defined (500 Internal Server Error)
```

this is unspecific and doesn't help the consumer.. this PR addresses this and throws an explicit Exception containing the json decode error..

### DIFF
--- a/src/Graviton/RestBundle/Listener/ValidationRequestListener.php
+++ b/src/Graviton/RestBundle/Listener/ValidationRequestListener.php
@@ -65,6 +65,7 @@ class ValidationRequestListener
             // specially check for parse error ($input decodes to null) and report accordingly..
             if (is_null($input) && JSON_ERROR_NONE !== json_last_error()) {
                 $e = new MalformedInputException(json_last_error_msg());
+                $e->setErrorType(json_last_error());
                 $e->setResponse($event->getResponse());
                 throw $e;
             }
--- a/src/Graviton/RestBundle/Listener/ValidationRequestListener.php
+++ b/src/Graviton/RestBundle/Listener/ValidationRequestListener.php
@@ -5,6 +5,7 @@
 
 namespace Graviton\RestBundle\Listener;
 
+use Graviton\ExceptionBundle\Exception\MalformedInputException;
 use Symfony\Component\EventDispatcher\Event;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
@@ -57,6 +58,13 @@ class ValidationRequestListener
             // Decode the json from request
             if (!($input = json_decode($content, true)) && JSON_ERROR_NONE === json_last_error()) {
                 $e = new NoInputException();
+                $e->setResponse($event->getResponse());
+                throw $e;
+            }
+
+            // specially check for parse error ($input decodes to null) and report accordingly..
+            if (is_null($input) && JSON_ERROR_NONE !== json_last_error()) {
+                $e = new MalformedInputException(json_last_error_msg());
                 $e->setResponse($event->getResponse());
                 throw $e;
             }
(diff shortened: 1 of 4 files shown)


narcoticfresh:
@lapistano @hairmare 

I'm a bit confused.. There are currently 2 types of `Exceptions` we throw in `ValidationRequestListener`: 1) `NoInputException` 2) `MalformedInputException` (the new one introduced here)..

They are thrown in the same way..

Still - if `NoInputException`is thrown, it currently gets outputted as JSON and HTTP status 400. If  `MalformedInputException` is thrown, the framework renders it as HTML and HTTP Status 500.

Can somebody explain me why (who?) this is this way? Is any Listener involved, anything? I'd like to clean this up in another PR..

narcoticfresh:
Hmm.. And in travis context the error seems to be outputted as HTML again - which is not the same behavior as i have locally using no special settings(?)

https://travis-ci.org/libgraviton/graviton/jobs/55929027

That's strange..

narcoticfresh:
@hairmare I needed to add a check for `json_last_error_msg()` as travis failed on the php54 build (function does only exist in php55 and upwards)..

Are you certain we still need to check travis for php54? My guess is nobody uses that anymore (i wouldn't be aware of that).. Maybe we could replace the php54 check with a php56 check to prepare for the future..

narcoticfresh:
@hairmare Ah, ok, i've just seen #66 .. ok ;-) Question remains if we can ditch php54..

src/Graviton/RestBundle/Listener/ValidationRequestListener.php
```diff
@@ -61,6 +62,19 @@ public function onKernelRequest(Event $event)
                 throw $e;
             }
 
+            // specially check for parse error ($input decodes to null) and report accordingly..
+            if (is_null($input) && JSON_ERROR_NONE !== json_last_error()) {
+                $lastJsonError = '';
+                if (function_exists('json_last_error_msg')) { // for php54
```
hairmare: Let's either decide on php55 or keep doing stuff the php54 way. I don't think us starting to add backwards compat in this fashions makes sense.

narcoticfresh: +1 for php55

src/Graviton/TestBundle/Client.php
```diff
@@ -36,55 +36,97 @@ public function getResults()
...
+        array $server = array(),
+        $jsonEncode = true
+    ) {
+        return $this->request(
+            'POST',
+            $uri,
+            $parameters,
+            $files,
+            $server,
+            $jsonEncode ? json_encode($content) : $content
```
hairmare: I know it's only in test here, but do we really need to start using the ternary operator everywhere?

Anyhow i feel we should just not use the ternary operator if possible. It's very good at hiding code smells and comes with a built-in lack of readability.

src/Graviton/RestBundle/Listener/ValidationRequestListener.php
hairmare: Here in this case we should probably wrap the parsing in some bit of symfony so we can let them add compat.

It seems having a php54/php55 compat case just for an error message should be handled by a full-stack framework somehow better.

narcoticfresh: What does this mean now for this PR? How you want it to be now?

hairmare: I feel that we should either ditch php54 alltogether and communicate that or keep it and just not using features it doesn't offer.

If @lapistano also feels that we can switch to php55 then we should probably update #66 to reflect that and remove the kludge above.

If we want to maintain php54 compat we could use a compat lib from packagist that backport the method instead of doing it ourselves.

Personally I'm all for php55 and I'd also activate hhvw as allowed failure if i touch #66 again.

narcoticfresh:
ok, i removed compat stuff and ternary operator stuff..

so please bring #66 into develop, I'll merge develop back (then travis should be green) and then we can merge this..

would be nice if #66 would be merged asap..

narcoticfresh:
@lapistano I merged develop back; now we have it all green.. ready to merge..

PR #81, opened 2015-03-26
generate validation rules for int & float fields

narcoticfresh:
before you could POST 'true' and 'false' (booleans) and it would save it as int's (0 and 1). this is not a sane behavior.. add tests for module to test this..

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
@@ -1,0 +1,268 @@
+<?php
+/**
+ * functional test for /core/module
+ */
+
+namespace Graviton\CoreBundle\Tests\Controller;
+
+use Graviton\TestBundle\Test\RestTestCase;
+
+/**
+ * Basic functional test for /core/module.
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class ModuleControllerTest extends RestTestCase
+{
+    /**
+     * @const complete content type string expected on a resouce
+     */
+    const CONTENT_TYPE = 'application/json; charset=UTF-8; profile=http://localhost/schema/core/module/item';
+
+    /**
+     * @const corresponding vendorized schema mime type
+     */
+    const COLLECTION_TYPE = 'application/json; charset=UTF-8; profile=http://localhost/schema/core/module/collection';
+
+    /**
+     * setup client and load fixtures
+     *
+     * @return void
+     */
+    public function setUp()
+    {
+        $this->loadFixtures(
+            array(
+                'GravitonDyn\ModuleBundle\DataFixtures\MongoDB\LoadModuleData'
+            ),
+            null,
+            'doctrine_mongodb'
+        );
+    }
+
+    /**
+     * test if we can get list of modules paged
+     *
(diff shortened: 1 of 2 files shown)


src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -0,0 +1,268 @@
...
+        $client->request('GET', '/core/module?q='.urlencode('eq(key,investment)'));
+        $response = $client->getResponse();
+        $results = $client->getResults();
+
+        $this->assertResponseContentType(self::COLLECTION_TYPE, $response);
+        $this->assertEquals('investment', $results[0]->key);
+        $this->assertEquals(1, count($results));
+
+        // get entry by id
+        $moduleId = $results[0]->id;
```
lapistano: I would start a separate Test here. 
So assert that the $results[0]->id is the as it is expected and use it as requirement for a next test.

src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -0,0 +1,268 @@
...
```
lapistano: same as above

src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -0,0 +1,268 @@
...
+     */
+    public function testDeleteModule()
+    {
+        // get id first..
+        $client = static::createRestClient();
+        $client->request('GET', '/core/module?q='.urlencode('eq(key,investment)'));
+        $results = $client->getResults();
+
+        // get entry by id
+        $moduleId = $results[0]->id;
```
lapistano: same as above.
Separating these tests will probably result in reducing code duplication as well.

narcoticfresh: Well, that is my problem - I don't know the record IDs..

`/core/module` is a service where we decided to not fix the ID - the user should lookup by the fixed field `key` (what we called "string id") instead. So I need to do that also in the test.

That's why I need to do this that way. I just wanted to do it once - but each time `setUp()` is called, i get potentially new IDs for the records, so I need to re-query them in every test..

I did the whole test basically only for the `testPostInvalidInteger` test - to test the thing i introduced here.. Also, this is all redundant to `AppControllerTest` anyway..

Maybe we should only test `testPostInvalidInteger` here and then test all the things the generator does (as this service is also generated) in a central `ShowcaseControllerTest` that needs to be done in a separate PR.

Do you know how to make it nicer under these circumstance (not knowing the ID & need to re-fetch every time)?

narcoticfresh:
I can obviously do the "key-to-ID" lookup in a separate function, that would reduce code duplication - but I cannot make an `assert` on the ID itself.. And I think it's that what you imply..

src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
lapistano: ok, did not know that there are no fixed ids. I leave the tests here as well, even if they are verifying stuff already tested elsewhere. 
So under this circumstances I currently do not see another way as well .

hairmare: We should actually ditch stringId if what @narcoticfresh desecribes is it's sole purpose here. There is now reason not to use id and fix upserting and POSTing those.

PR #84, opened 2015-03-30
fix translation

narcoticfresh:
i'm not trying to re-invent the translation wheel or fix all broken things. This PR just wants translation to actually *work*

It corrects some bugs:
* Make sure that both Listeners use the same `domain` for there translation (thus finding them in the first place)
* Touch the trigger file for the loader not only when the file *doesn't* exist..
* Create a new `I18nUtils` service that keeps all this stuff in one place - it is called `intUtils` in the classes as `$i18nUtils` is not valid according to syntax check.

After this merge, translations actually and really work across all controllers. Hooray!

- [x] move default locale to config.yml - shouldn't be a param
- [x] more save checks in event (postpersist)
- [x] look for alternative for `touch()` - a `fopen($fp, 'w+')` also truncates a file automatically
- [x] rename `intUtils` to `utils` in classes..
- [ ] fix weird travis issues with triggering a reload of translatables during a test

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -6,6 +6,7 @@ parameters:
     graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
     graviton.authentication.user_provider.model: ~
     graviton.security.services.whitelist: ~
+    graviton.rest.special_mimetypes: {}
 
     # service info injected by cloudfoundry
     vcap.services: ~
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -3,6 +3,7 @@ parameters:
     mongodb.default.server.db: db
     mongodb.default.server.uri: ~
 
+    graviton.translator.default.locale: en
     graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
     graviton.authentication.user_provider.model: ~
     graviton.security.services.whitelist: ~
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -3,7 +3,6 @@ parameters:
     mongodb.default.server.db: db
     mongodb.default.server.uri: ~
 
-    graviton.translator.default.locale: en
     graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
     graviton.authentication.user_provider.model: ~
     graviton.security.services.whitelist: ~
--- a/src/Graviton/I18nBundle/Listener/I18nDeserializationListener.php
+++ b/src/Graviton/I18nBundle/Listener/I18nDeserializationListener.php
@@ -7,10 +7,7 @@ namespace Graviton\I18nBundle\Listener;
 
 use Graviton\I18NBundle\Service\I18NUtils;
 use JMS\Serializer\EventDispatcher\PreDeserializeEvent;
-use Symfony\Component\HttpFoundation\Request;
 use Graviton\I18nBundle\Document\TranslatableDocumentInterface;
-use Graviton\I18nBundle\Document\Translatable;
-use Graviton\I18nBundle\Model\Translatable as TranslatableModel;
 
 /**
  * translate fields during serialization
@@ -29,18 +26,18 @@ class I18nDeserializationListener
     /**
      * @var \Graviton\I18nBundle\Service\I18nUtils
      */
-    protected $i18nUtils;
+    protected $intUtils;
 
     /**
-     * set utils
+     * set intUtils (i18nutils)
      *
-     * @param \Graviton\I18NBundle\Service\I18NUtils $i18nUtils utils
+     * @param \Graviton\I18NBundle\Service\I18NUtils $intUtils utils
      *
      * @return void
      */
-    public function setI18nUtils(I18NUtils $i18nUtils)
+    public function setIntUtils(I18NUtils $intUtils)
     {
-        $this->i18nUtils = $i18nUtils;
+        $this->intUtils = $intUtils;
     }
 
     /**
@@ -53,7 +50,7 @@ class I18nDeserializationListener
     public function onPreDeserialize(PreDeserializeEvent $event)
     {
         $eventClass = $event->getType()['name'];
-        $defaultLanguage = $this->i18nUtils->getDefaultLanguage();
+        $defaultLanguage = $this->intUtils->getDefaultLanguage();
         $object = new $eventClass;
         if ($object instanceof TranslatableDocumentInterface) {
             $data = $event->getData();
@@ -82,7 +79,7 @@ class I18nDeserializationListener
         \array_walk(
             $this->localizedFields,
             function ($values) {
-                $this->i18nUtils->insertTranslatable($values);
+                $this->intUtils->insertTranslatable($values);
             }
         );
     }
--- a/src/Graviton/I18nBundle/Listener/I18nDeserializationListener.php
+++ b/src/Graviton/I18nBundle/Listener/I18nDeserializationListener.php
@@ -5,6 +5,7 @@
 
 namespace Graviton\I18nBundle\Listener;
 
+use Graviton\ExceptionBundle\Exception\DeserializationException;
 use Graviton\I18NBundle\Service\I18NUtils;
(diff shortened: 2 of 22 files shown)


src/Graviton/I18nBundle/Listener/I18nDeserializationListener.php
```diff
@@ -69,6 +50,7 @@ public function setTranslatables(TranslatableModel $translatables)
     public function onPreDeserialize(PreDeserializeEvent $event)
     {
         $eventClass = $event->getType()['name'];
```
lapistano: if ```$event->getType()['name']``` is empty this will crash at line 54.  Probably a validation and a specific exception would be nice.

lapistano: also a validation that there is a ```$eventClass``` registered, should be in place.

src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
```diff
@@ -42,13 +42,13 @@ public function postPersist(LifecycleEventArgs $event)
...
-                touch($triggerFile);
+            // @todo faster less blocking operation than using touch()
+            $transDir = __DIR__.'/../Resources/translations/';
+            $triggerFile = $domain.'.'.$locale.'.odm';
+
+            if (is_dir($transDir)) {
+                touch($transDir.$triggerFile);
             }
-            $fp = fopen($triggerFile, 'rw');
-            ftruncate($fp, 0);
```
hairmare: It took quite some digging to find this optimisation. If hou have any questions I can answer them tomorrow, but we should not throw this optimisation away without benching it again first. AFAIK this has more to do with libc than with php.

src/Graviton/I18nBundle/Tests/Controller/LanguageControllerTest.php
```diff
@@ -37,13 +37,17 @@ public function testFindAll()
         $this->assertResponseContentType(self::CONTENT_TYPE . 'collection', $response);
 
         // we assume that initially all systems will only know of the english lang
-        $this->assertcount(1, $results);
+        $this->assertcount(3, $results);
```
hairmare: tsk tsk tsk, you should not change tests for this. It's not like the tests where wrong before. This is the exact reason why we **must** to differentiate between test-fixtures and demo-data!

src/Graviton/I18nBundle/Tests/Controller/LanguageControllerTest.php
```diff
@@ -106,14 +110,14 @@ public function testDontReturnUnknownLanguage()
     {
         $client = static::createRestClient();
 
-        $client->request('GET', '/i18n/language/en', array(), array(), array('HTTP_ACCEPT_LANGUAGE' => 'en,de'));
+        $client->request('GET', '/i18n/language/en', array(), array(), array('HTTP_ACCEPT_LANGUAGE' => 'en,es'));
```
hairmare: Don't change tests please. They test a known baseline for reasons and should not be changed based upon your personal preferences.

app/config/parameters.yml.dist
```diff
@@ -3,6 +3,7 @@ parameters:
     mongodb.default.server.db: db
     mongodb.default.server.uri: ~
 
+    graviton.translator.default.locale: en
```
hairmare: I don't like this change at all. It completely contradicts what we defined a long time ago (keeping the default as ``en`` and making everything else translated).

src/Graviton/I18nBundle/Listener/I18nDeserializationListener.php
```diff
@@ -26,37 +24,20 @@ class I18nDeserializationListener
     protected $localizedFields = array();
 
     /**
-     * @var \Symfony\Component\HttpFoundation\Request
+     * @var \Graviton\I18nBundle\Service\I18nUtils
      */
-    protected $request;
+    protected $intUtils;
```
hairmare: Have you considered just storing ``I18nUtils`` in ``$utils``. I find that the context should be enough for folks to realise that we mean ``I18Utils``. Personally I thing ``inUtils`` is even more confusing.

src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
hairmare: To bench it you would have to write heaps of records using faker or something similar. Since this is called after each update in translatable I was saving heaps of time when importing large amount of data. This made quite a difference on travis at the time.

narcoticfresh:
@hairmare *Changing tests*
I don't see how we can test translation stuff with only 1 `Language`? So we need more in order to translate. <br>That's why I changed that `unknown Test` to `es`, because `de` is a `Language` now. Same with the amount check..

*Default locale setting*
I don't expect somebody to change that default. But I don't like hardcoded stuff scattered all over the place. At least now we have an external string everybody needing the "default locale" can refer to.

src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
narcoticfresh: @hairmare 
OK - Problem is, my cache *never* got invalidated using this. It only got invalidated (thus regenerated once using `touch`), then I never got translations again to the client as the subsequent  `PostPersist`s never triggered the cache correctly anymore..

My guess as `touch` already produces an empty file, doing the truncate doesn't actually modify the file if it is already empty..

hairmare:
@narcoticfresh The default value should not be hardcoded anywhere, thats where I'm very happy with this change. It just shouldn't be exposed in parameters.yml-dist, but rather in config.yml or so.

Regarding the tests, I'll have a version with reverted tests and a new one for testing multiple languages soonish and will push it to this PR.

narcoticfresh:
@hairmare ACK. I'll let you push that, then I'll add the stuff I added in the PR description above.

narcoticfresh:
@hairmare 

OK; i think it looks much better now.. Let's wait for Travis then you can merge it I guess..

I think the `fopen()` approach is probably fast enough.. if we find another method, we can always change that part..

hairmare:
@narcoticfresh Do you plan on addressing https://github.com/libgraviton/graviton/pull/84#discussion-diff-27408603R52 before we merge? Seems like a simple check-and-throw-execption might be ok there... OTOH this should in the events contract somehow.

hairmare:
@narcoticfresh I would have merged this had travis not failed. You might need to ``rm -rf app/cache`` to make sure everything acts the same way as on travis locally.

hairmare:
``fclose(fopen($file, 'a'))`` seems to be faster than what's in b64a5f5. With this PHPUnit outputs ``Time: 12.32 seconds, Memory: 145.00Mb``.

It might be to not defining ``$fp`` or more likely not doing the ``fopen($file, 'w')``.

I'd write something sounding highly intelligent about zvals and mutexes but I'm not sure I really grok where these penalties are coming from.

I'm pushing this change in a bit so we can see if this works on travis...

hairmare:
My last change made it a tiny bit faster but doesn't fix the race condition. The most weird thing is that this is only failing with the es->es_ca part locally but always fails at the en->es part on travis as far as i can tell.

narcoticfresh:
@hairmare commit dac1721 also let's fail tests locally.. opening in mode `a` doesn't seem to be enough.. if i change it to `w`, tests don't fail locally anymore in my case..

but i don't see what travis' problem is.. ;-//
i'll change the file mode to `w` to see if it helps..

hairmare:
OK, travis is green again, but I did have to add another ``touch`` for ``es`` to the config.

Since I haven't really been able to reproduce this reliably locally I feel we should just merge the PR at this stage and get on with our lives. This probably will need more fixing at some point, but I'd like to let it autodeploy from develop and then retest it there before we carry on.

@narcoticfresh whatcha think?

Even though this might well still be broken, I don't believe we will be able to fix it without making some noise upstream. Basically, Symfony should have better support for non file based i18n backends.

narcoticfresh:
@hairmare 

Hm, I really don't get why this initial touch fixes this.. But well, I don't get Travis' problem in the first place..

So agreed, let's merge this..

PR #85, opened 2015-04-07
Digest and render $refs by using a doctrine custom type

hairmare:
This is the hacky doctrine custom type solution. It completely bypasses the serializer when it comes to doing its thing and it also has some integration quirks of its own. Nevertheless it seems to work ok and fulfil our basic use-case.

* converting a url from a request to it's storage equivalent is done in the custom type.
* conversion from a mongodb-ref to an url is done in a listener due to the fact I was unable to inject a router (or anything for a matter of fact) into the hydrator that was doing the work.
* Lots of things are done through config and in compiler passes. While some of the generators aren't as clean as they could be, at least they don't affect the runtime.

I've still some small issues open, but this is at a stage where I need a PR to get better scrutinizer info. It's probably already in enough shape for review keeping the following tasks in mind.

* [x] integrate ``extref`` type with generator
* [x] ~~test URL as an id edge case~~ This case seems invalid to me at this point, due to me not seeing how I would even start tackling this. It also doesn't make sense since we use IDs as parts of URLs and an URL should not be parts of an URL.
* [x] link /core/module with /core/app
* [x] add docs in a ``README.md``
* [x] clean up the code some more, scrutinizer seems to help seeing what's wrong so far
* [x] solve generating documents with fields that contain ``$``
* [x] check that ~~second level~~  refs still work in coded case
* [x] check that second level refs work in generated case
* [x] move ``/core/module.app`` to ``/core/modules/app/$ref``
* [x] get rid of ``substr`` in ``getPathFromUrl``
* [x] ~~make reading and writing of ``extref`` types fully recursive by testing nth level things somehow (partially taken care of, might need work in ``ExtRefListener`` and in seemingly also in low-level parts of generator I haven't touched in this PR :unamused:)~~ Since the generator doesn't do recursions here I won't for now either. In any case they should be addable later without to many pain.
* [x] make ``ExtRefFieldsCompilerPass`` load subobjects into fields
* [x] fix ``deserializeRef`` in generated ``Document.php`` files
* [x] get rid of hardcoded ``ref`` in ``ExtRefFieldsCompilerPass``
* [x] support ``embed-many`` case ~~and consider supporting ``reference-*``~~ We'll add ``reference-*`` when we need it.
* [x] get #89 merged so this will work on nova cloud
* [x] figure out why ``app.$ref`` in ``/core/module`` isn't getting rendered as expected


src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -171,7 +174,7 @@ public function testPostInvalidInteger()
     {
         $testModule = new \stdClass;
         $testModule->key = 'test';
-        $testModule->appId = 'testapp';
+        $testModule->app = 'http://localhost/core/app/testapp';
```
narcoticfresh: @hairmare Just a general question.. Shouldn't be `app` property of a Module (and thus any object property having a relation by link) have an object as:

```js
{"module": {"app": { "$ref": "http://localhost/core/app/testapp"}}}
```

I thought we would render out an object?

src/Graviton/DocumentBundle/Types/ExtReference.php
```diff
@@ -0,0 +1,164 @@
...
+     * get path from url
+     *
+     * @param string $url url from request
+     *
+     * @return string
+     */
+    private function getPathFromUrl($url)
+    {
+        $ret = $url;
+        if (substr($url, 0, 4) == 'http') {
```
narcoticfresh: if you pass the url to `parse_url`, it will return `false` on malformed input anyway (and you check that) - why the `substr` check?

src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
```diff
@@ -0,0 +1,109 @@
...
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        if (!$event->isMasterRequest() || empty($event->getResponse()->getContent())) {
+            return;
+        }
+
+        $data = json_decode($event->getResponse()->getContent(), true);
+
+        if (is_array($data) && !empty($data) && !is_string(array_keys($data)[0])) {
+            foreach ($data as $index => $row) {
```
narcoticfresh: Do I see that correctly that this only works on the 1st level? So if we return a an Account with an embedded object that has a ref - this doesn't work anymore?

narcoticfresh:
I can obviously not say if this does what it should, I don't have enough time to really check that currently.. It's really hackish and a bit all over the place, but we agreed on that, so hmm.. ok..

What impact does the merge have?
What does need to be in a service definition json so this will do something? How to specify the extref type? 

If this is not done yet (generator part for whatever it needs to get activated); please do this in this PR..

src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
hairmare: I initially wanted to render an object and this solution turned out easier. In the end I think we should support both at some point, but I'm afraid that this might be broken for generated sub-objects but I need to check.

src/Graviton/DocumentBundle/Types/ExtReference.php
hairmare: Yupp, thats true. I'll refactor this. It happened due to the fact that I had the ``substr`` ages before i implemented the ``parse_url`` call and its check.

src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
hairmare: I'll need to test the non 1st level cases. It might well be that they don't work properly yet.

hairmare: I believe this needs to support dot notation or something to work properly.

narcoticfresh:
Ah ok, ACK to the README.. Didn't see that.. Good ;-)

It's obvious we need to replace this completely with a Serializer approach (which serializer one doesn't matter *g) as soon we have enough time at some point in the far distant future ;-)

I think we can merge this as soon `ExtReferenceListener.php` supports infinite (or at least ~3 if that matters) levels.. That's a must have.. After that, we can see what more it could bring up, but we can merge this first step..

hairmare:
Yeah, I can see that n levels need to be taken care of before this is ready to merge. I'll need to think over it a bit to find a non-crappy solution though and I don't feel doing it recursive makes sense. I'll probably try grabbing it from conf.

About the serializer. I somehow feel we should just write our own highly opinionated serializer and be done with it.

I'm not quite sure as how I would wrap the low level things in it and where to start though. I'm also sure that we would need to do some pretty nasty hacks to write a proper serializer and I'm not really keen on that.

The serializer discussion probably boils down to us keeping this hacky solution for now and hoping someone else really takes care of serialization in a sensible manner. Maybe symfony-3 will bring the kind of changes needed for this, but I'm not betting on that :stuck_out_tongue:

hairmare:
@narcoticfresh I'm winding down on this one.

I think I've got all the things we mentioned covered except for the following points.

If we merge it like this, I'll add trackers for the remaining issues.

#### using ``$ref`` from the nth level

This doesn't seem to work at the generator level. Adding the following to ``Showcase.json`` has no effect whatsoever and isn't even close to what I would need here ( ``my.language.in.deeply.nested.ref``).

```js
// fixture
                "somewhat": {
                    "nested": {
                        "text": "Hello World!"
                    }
                },
// field
            {
                "name": "somewhat.nested.text",
                "type": "string",
                "title": "Nested String",
                "description": "Deeply nested string"
            },
```

It works as intended on the ``app.$ref`` level in ``/core/module`` and as ``$language`` in ``/i18n/translatable`` so I'm ok with it for now. It might work for more cases, but I would certainly need to test against those to be sure.

Do you know of a working case of nth level stuff with the generator?

#### ``embed-many`` and ``reference-*`` refs

For now I only follow ``embed-once`` links when generating mappings for serialization/deserialization. Before I add them I'd like to clarify their use cases so they get added in places we actually need them.

hairmare:
One more thing and then I'm done for the day... I had to open #89 to fix our cloud deploy and then needed to create the "figure out why app.$ref in /core/module isn't getting rendered as expected" task.

This means there is still some work here  :(

narcoticfresh:
I don't quite get the whole picture - but you seem to have a `ExtRef` php class - can't you just register a [Serializer handler](http://jmsyst.com/libs/serializer/master/handlers) for that type and replace the whole Listener?

Before you need to specify that class type in the Serializer config..

hairmare:
@narcoticfresh I did try using a deserializer at some point. I removed it in 42ead34 since it was not working as I intended. I'll have to look into this again at some point, but am not sure if I'll do so in this PR.

hairmare:
All righty then. My work here is done for the moment. I believe I've addressed all pressing concerns as per my tasks and notes in the main issue text.

You can verify that this works on a cloudy instance of mine:
* [referencing a language in translatable](http://graviton-bil-lardiest-capitalization.nova.scapp.io/i18n/translatable)
* [reference app from generated module](http://graviton-bil-lardiest-capitalization.nova.scapp.io/core/module)
* [array of references from generator](http://graviton-bil-lardiest-capitalization.nova.scapp.io/hans/showcase)

While there is still room for improvement, I wont be doing them on this branch due to it's already large size and the fact that we finally have something that works.

We should be able to get [GRV-1571](https://issue.swisscom.ch/browse/GRV-1571) and [GRV-1426](https://issue.swisscom.ch/browse/GRV-1426) reviewed (and hopefully closed) with this.
I'm not sure if we can close the [GRV-1425](https://issue.swisscom.ch/browse/GRV-1425) Epic after merging. There might still be dragons lurking and we'll have to create issues as needed. I suspect that we will have lots of requirements to use this, so we'll be finding them soonish.

@narcoticfresh, @lapistano Please review and merge as soon as travis and scrutinizer did their deeds (I'm monitoring them and they will either be green or I have more work to do).

narcoticfresh:
@hairmare  @lapistano 
I'll look at the links you mentioned.. IMHO it would make sense if @lapistano did the merge. If something pops up on this one in the coming weeks it's important he'll seen it more in-depth..

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param string  $doc      document name
+     * @param boolean $embedded is this an embedded doc, further args are only for embeddeds
+     * @param string  $name     name prefix of document the embedded field belongs to
+     * @param string  $prefix   prefix to add to embedded field name
+     *
+     * @return void
+     */
+    private function loadFields(&$map, $ns, $bundle, $doc, $embedded = false, $name = '', $prefix = '')
+    {
+        if ($ns === 'gravitondyn') {
```
lapistano: if (strtolower($ns) === 'gravitondyn')  » more safe to match

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param ContainerBuilder $container container builder
+     * @param array            $services  services to inspect
+     *
+     * @return void
+     */
+    public function processServices(ContainerBuilder $container, $services)
+    {
+        $map = [];
+        foreach ($services as $id) {
+            list($ns, $bundle,, $doc) = explode('.', $id);
```
lapistano: if $id does not have at least 4 '.' this will fail with no warning.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
hairmare: Fixed, will push as soon as there are no other small things remaining that do not each warrant their own build.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param string  $prefix   prefix to add to embedded field name
+     *
+     * @return void
+     */
+    private function loadFields(&$map, $ns, $bundle, $doc, $embedded = false, $name = '', $prefix = '')
+    {
+        if ($ns === 'gravitondyn') {
+            $ns = 'GravitonDyn';
+        }
+        $file = implode(
```
lapistano: missing input validation. $bundle being NULL will cause a  warning when ucfirst(), being an empty string will cause the bundle name be 'Bundle' ... and so on .. » please add input validation/assertion

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param array   $map      map to add entries to
+     * @param string  $ns       namespace
+     * @param string  $bundle   bundle name
+     * @param string  $doc      document name
+     * @param boolean $embedded is this an embedded doc, further args are only for embeddeds
+     * @param string  $name     name prefix of document the embedded field belongs to
+     * @param string  $prefix   prefix to add to embedded field name
+     *
+     * @return void
+     */
```
lapistano: this is way to long, I'd appreciate if split in multiple methods, separately tested.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
hairmare: I added some guard code for ``$bundle`` and ``$doc``. If ``$ns`` is empty everything whent titsup earliery anyhow.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
hairmare: The guard code in the caller takes care of not passing an empty string now.

I can add ``assert('!empty($ns) && !empty($bundle)');`` if we need more checking, but then again it's not like our code-base has assertions throughout.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
hairmare: Yeah I'd appreciate that too, thats what I was alluding to when I was talking about unclean code in the intro to the PR. At least it doesn't get executed during runtime.

The code looks worse than it actually is, it's complexity is only 9 and a lot of the length is due to formatting.

I'm kinda fed up with this whole ``$ref`` crap so I'd rather we just add this refactoring to the technical dept epic and get on with life.

The method does have some other problems as well (don't you dare changing the name of GravitonDyn or even be reccless enough to change where it's stored, we should just instantiate a finder instead of not using symfony tools, ...).

If I were to really clean it up, the individual concerns should not just get their own methods but even their own objects to encapsulate everything cleanly. When I saw the additional work that would go into that I decided on opening the PR as-is.

lapistano: My former comment was a proposal for a first step. 
OK, go ahead and add a re-factoring story and see it will be recognized for a spint in the near future.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/AbstractExtRefCompilerPass.php
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @param ContainerBuilder $container container builder
+     *
+     * @return void
+     */
+    final public function process(ContainerBuilder $container)
+    {
+        $gravitonServices = array_filter(
+            $container->getServiceIds(),
+            function ($id) {
```
lapistano: I see different stuff here:

1. matching/finding part of strings without taking care of upper/lowercase.
2. I'd like to see the actual filter callback as a strategy. But am fine with 
it, when a story (technical debt) will be added to the backlog.
```php 
[..]
function ($id) use $ServiceFilterStrategy { return $ServiceFilterStrategy->apply($id); }
[..]
```

src/Graviton/DocumentBundle/DependencyInjection/Compiler/AbstractExtRefCompilerPass.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * null implementation of method
+     *
+     * @param ContainerBuilder $container container
+     * @param array            $services  services
+     *
+     * @return void
+     */
+    public function processServices(ContainerBuilder $container, $services)
```
lapistano: this should be abstract. If there is a need for a NoOpCompilerPass it has to have a dedicated implementation. Further an abstract class with no abstract method does not make sense either (abstract vs base class)

src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
lapistano: I guess, too.

src/Graviton/DocumentBundle/Types/ExtReference.php
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * get php value when field is used as identifier
+     *
+     * @param \MongoDBRef $value ref from mongodb
+     *
+     * @return string
+     */
+    public function convertToPHPValue($value)
+    {
+        return $this->router->generate($this->mapping[$value['$ref']], ['id' => $value['$id']]);
```
lapistano: pls check for null values.

src/Graviton/DocumentBundle/Types/ExtReference.php
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * return the mongodb representation from a php value
+     *
+     * @param string $value value of reference as URI
+     *
+     * @return array
+     */
+    public function convertToDatabaseValue($value)
+    {
```
lapistano: pls check for null values.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/AbstractExtRefCompilerPass.php
hairmare: On your first point. See:

1. Run ``app/console container:debug``
2. Count number of services containing uppercase letters
3. Add the exact same number of calls to ``strtolower`` to this method
4. Done

On the second point, I see your point here but there certainly are more pressing matters than rewriting a one-line lambda arg to an ``array_filter`` call that doesn't even do anything notably complex.

src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
hairmare: BTW: This now supports the same notion of recursing as the generator bundle does. While it's not full recursion it should work with the examples in showcase.json. If we ever need more here, this should be easy enough to refactor at that point. But then again, we'd also be refactoring the generator for that to make sense.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/AbstractExtRefCompilerPass.php
hairmare: Fixed

src/Graviton/DocumentBundle/Types/ExtReference.php
hairmare: check added

src/Graviton/DocumentBundle/Types/ExtReference.php
hairmare: Done

src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefFieldsCompilerPass.php
hairmare: Yeah, no worries. As you can see from my post I also see lots of refactoring potential.

src/Graviton/DocumentBundle/DependencyInjection/Compiler/AbstractExtRefCompilerPass.php
lapistano: your first point:
service ids are only lowercase by convention. the unforunate thing is that people or generators do make mistakes and then one is entitled to debug and this error will be a hell of a thing to find .. seriously. 
So please add the strtolower() and be sure stuff does not break because on a erroneous matching.

hairmare: Added...

narcoticfresh:
Duu sö mörtsch??

PR #86, opened 2015-04-08
add setter for id property in generated documents

narcoticfresh:
i need this in a service bundle in order to generate objects from different objects and return objects that have an id property..


src/Graviton/GeneratorBundle/Resources/skeleton/document/Document.php.twig
```diff
@@ -64,6 +64,18 @@ class {{ document }} implements TranslatableDocumentInterface
         return $this->id;
     }
 
+    /**
+    * Set id
```
hairmare: Y U indent weird?

hairmare:
Sorry about nitpicking. I also retriggered scrutinizer and will merge as soon as he's finished.

PR #87, opened 2015-04-08
Small fixes throughout

hairmare:
These are all really small. Basically I just fixed a heap of scrutinizer issues to get some distance to the whole ref thing for a change.

I also ran a ``composer update`` just for the heck of it and pulled in various small changes there as well.

The most important change is probably in f5c3be8 and I'll split that out if we can't get this merged ASAP.


narcoticfresh:
I'm ok with the changes.. But I'm a bit worried about the composer update.. Last time I did this it broke VCAP stuff for the cloud push.. Hmm..

But I guess we cannot be frozen on the current lock; I know..

narcoticfresh:
@hairmare Could we maybe take the composer update out of this PR?

hairmare:
Do you really think we have to? I thought the composer update issues where resolved now that the buildpack can do composer update again. We can also use an older version of the build pack if this still crops up. Don't we have that workaround in place anyway?

hairmare:
@narcoticfresh I just split out the most urgent matter into #88. Could you please also address my Q regarding the workaround in the last comment?

narcoticfresh:
@hairmare Sorry, had no focus on Github this morning 'til you spammed me ;-)

Let's merge this and see what happens with the wrappers.. We need to deal with this anyway..
I'll close #88..

hairmare:
@narcoticfresh Thanks

PR #88, opened 2015-04-09
Stop spamming hipchat

hairmare:
This should only notify fails and when a build goes from fail to success, at least thats what I'm intending with it.

It i a cherry pick of the corresponding commit from #87. We still need to merge #87 and I'm absolutely not happy with #87 being help up over a ``composer update``. As promised, this contains just the hipchat change on its own.


narcoticfresh:
Not needed; merged with #87

PR #89, opened 2015-04-13
Fix cf deploy once again

hairmare:
There was in fact a bug in the last ``composer update`` we merged.

This PR skips symfony/symfony v2.5.6 and thus fixes pushing to the cloud.

I still need to sift through https://github.com/symfony/symfony/compare/v2.6.5...v2.6.6 and figure out what where this issue is coming from (or just create an issue upstream if I can't). It could be https://github.com/symfony/symfony/commit/ce2764d416c7287d2126e3d1fb9338f329c1d80f but I need to check...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     },
     "require": {
         "php": "~5.4",
-        "symfony/symfony": "2.6.*",
+        "symfony/symfony": "=2.6.5 | >=2.6.7",
         "doctrine/orm": "~2.4",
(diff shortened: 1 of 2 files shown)


lapistano:
@hairmare I trust you trust this does the trick ;)

hairmare:
I did trust it and it does solve one small issue uncovering the next one didn't exist when I wrote this. It's also a bit hacky and I need to sort out upstream or adapt our param handling to make this future proof.

PR #90, opened 2015-04-15
fix translation load exception

narcoticfresh:
I had the [InvalidArgumentException thrown by Symfony finder](http://api.symfony.com/2.6/Symfony/Component/Finder/Finder.html#method_in) if the directory it must search for files doesn't exist.. As this is thrown in app/console context, the Exception is unhandled and it leads to the stop of the fixture loading process. 

We can safely catch that exception and continue normal operation there..

### DIFF
--- a/src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
+++ b/src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
@@ -62,6 +62,8 @@ class PostPersistTranslatableListener implements EventSubscriber
                     $fs->remove($file->getRealPath());
                 }
             } catch (\InvalidArgumentException $e) {
+                // InvalidArgumentException gets thrown if the translation cache dir doesn't exist.
+                // we tolerate that is it's normal under some circumstances (no cache warmup yet)
             }
         }
     }
--- a/src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
+++ b/src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
@@ -62,9 +62,6 @@ class PostPersistTranslatableListener implements EventSubscriber
                     $fs->remove($file->getRealPath());
                 }
             } catch (\InvalidArgumentException $e) {
-                // InvalidArgumentException gets thrown if the translation cache dir doesn't exist.
-                // we tolerate that is it's normal under some circumstances (no cache warmup yet)
-                return true;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


narcoticfresh:
@hairmare any idea why Scrutinizer gets canceled so often?! Am I doing something wrong? ;-)

hairmare:
@narcoticfresh travis is taking ages again and scrutinizer started tracking it closer. For some reason it is canceling builds after some non configurable timeout. In scrutinizers logs it usually says something about travis failing the build. I tried adding a timeout but scrutinizer seems to ignore it.

hairmare:
Also, I retriggered the build on scrutnizer and will merge as soon as it's through.

src/Graviton/I18nBundle/Listener/PostPersistTranslatableListener.php
```diff
@@ -51,14 +51,20 @@ public function postPersist(LifecycleEventArgs $event)
...
 
-            foreach ($finder as $file) {
-                $fs->remove($file->getRealPath());
+                foreach ($finder as $file) {
+                    $fs->remove($file->getRealPath());
+                }
+            } catch (\InvalidArgumentException $e) {
+                // InvalidArgumentException gets thrown if the translation cache dir doesn't exist.
+                // we tolerate that is it's normal under some circumstances (no cache warmup yet)
+                return true;
```
hairmare: @narcoticfresh I feel like the ``return`` line could be remove and will do so on merging if that's ok with you

lapistano: I think it would be nice to log the exception in debug/dev mode.
The return shall be removed, since a listener callback shall not return anything.

hairmare: @lapistano nope, this shouldn't get logged. What this fixes is running any code after deleting ``app/cache/`` and the InvalidArg is not really an error condition since the missing dir (``app/cache/*/translations``) gets recreated ASAP anyway.

If anything we could check if ``$cacheDirMask`` exists before iterating.

narcoticfresh:
@hairmare @lapistano 
`return` is removed now.. I second it shouldn't be logged - it's just not noteworthy enough - we don't need to delete what doesn't exist ;-)

@hairmare I tried to check for existence of `$cacheDirMask`. But because it's a *mask* (having the wildcard what is very nice and comfy), i couldn't check it with `is_dir()` or such.. Letting it handle `Finder` seemed reasonable..

hairmare:
@narcoticfresh Yupp, I also think its preferable to handle it this way and will merge it when CI is through.

narcoticfresh:
Damn you scrutinizer ;-/

PR #91, opened 2015-04-16
ensure we don't create a new record when there is no ID on PUT request..

narcoticfresh:
This bug was encountered during a issue testing from Kyrill..

For some reason, he made a PUT request with a payload without an "id" field (as said, in the payload) - but obviously specifying one in the URI he requested. 
Graviton in this case inserted a new record and returned this one (because of the way we insert it in `DocumentModel`).  It's wrong to insert a record on PUT.

This fixes that, ensuring that we take the ID from the URI request if it's not on the payload already. I also added tests that ensure that behavior. If the entity on question doesn't have a `setId()` function and there is no ID in the payload, we quit and tell the user.

Normally, a read/write entity would always have a `setId()` function, I just want to make it more reliable there as this is not *always* the case..

### DIFF
--- a/src/Graviton/RestBundle/Controller/RestController.php
+++ b/src/Graviton/RestBundle/Controller/RestController.php
@@ -6,6 +6,7 @@
 namespace Graviton\RestBundle\Controller;
 
 use Graviton\ExceptionBundle\Exception\DeserializationException;
+use Graviton\ExceptionBundle\Exception\MalformedInputException;
 use Graviton\ExceptionBundle\Exception\NotFoundException;
 use Graviton\ExceptionBundle\Exception\SerializationException;
 use Graviton\ExceptionBundle\Exception\ValidationException;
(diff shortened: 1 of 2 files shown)


hairmare:
It is not ''wrong'' to insert on PUT. That is a the classical upsert case.

hairmare:
:+1:  on merging

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -329,6 +332,17 @@ public function putAction($id, Request $request)
             $this->getModel()->getEntityClass()
         );
 
+        // handle missing 'id' field in input to a PUT operation
+        // if it is settable on the document, let's set it and move on.. if not, inform the user..
+        if ($record->getId() != $id) {
+            // try to set it..
+            if (is_callable(array($record, 'setId'))) {
```
lapistano: I basically do not like this way to determine a class provides a certain method. I'd rather like to see this reflected  in an interface and the contract verified. I know this is not done in a minor refactoring, but I think we should think about the way documents are managed.

lapistano: nevertheless will merge.

hairmare: **truth**

PR #92, opened 2015-04-16
add hook-in possibilities for controllers that inherit RestController..

narcoticfresh:
Please, read all this first ;-)

**Why a change**
Currently, RestController offers no possibility from a Controller that inherits to easily influence what gets sent to the user / persisted to the database. But this is very often necessary.
The need for this came out from GRV-1490. I then created GRV-1581 to describe what we need and why.. 

One has to do *horrible* things currently to wrap RestController logic (to not duplicate it) because RestController doesn't offer any hook-in possibilities. Please see this [Controller in Stash](https://git.swisscom.ch/projects/GRV/repos/graviton-service-bundle-util/browse/src/Graviton/UtilBundle/Controller/LockController.php) implementing the locking service to see what I mean.

**Please please please**
I'm well aware this is not the nicest approach. But I have *very very* limited time. We can *always* make this more sophisticated at some point. But at the moment, this approach does everything we need and it's easy to use.

**Testing**
Please note that the fact that all normal tests run through at least demonstrate that this change has no impact on current functionality. That is the most important aspect.

I would like to have done some tests, but I'm not aware how to mock this efficiently (inject a different controller inheriting from RestController). I'm sorry, I don't have the time to dig in to this to see how it would be done.

### DIFF
--- a/src/Graviton/RestBundle/Controller/RestController.php
+++ b/src/Graviton/RestBundle/Controller/RestController.php
@@ -74,8 +74,7 @@ class RestController implements ContainerAwareInterface
             ->setStatusCode(Response::HTTP_OK);
 
         $record = $this->findRecord($id);
-
-        $response->setContent($this->serialize($record));
+        $response->setContent($this->serialize($this->onOutputOne($id, $record)));
 
         return $this->render(
             'GravitonRestBundle:Main:index.json.twig',
@@ -196,10 +195,12 @@ class RestController implements ContainerAwareInterface
             $model->setPaginator($paginator);
         }
 
+        $records = $this->onOutputAll($model->findAll($request));
+
         $response = $this->getResponse()
             ->setStatusCode(Response::HTTP_OK)
             ->setContent(
-                $this->serialize($model->findAll($request))
+                $this->serialize($records)
             );
 
         return $this->render(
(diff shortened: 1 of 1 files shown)


hairmare:
I'm absolutely not happy about adding what looks like document lifecycle callbacks to the already bloated REST controller. If anything these things should be listeners or even better doctrine callbacks.

I also thing the controller you linked should only validate expiry and deletion should get handled in the background.

Right now im :-1: on this since it looks way to much like legacy code.

hairmare:
Also on @narcoticfresh having limited time. I'm fully aware of that but we can always help out finishing the feature if you tell us where to help. Using it as an argument for crappy code is unfair.

narcoticfresh:
@hairmare 
Yes, it's not nice.. But *no*, this shouldn't be on Doctrine level exclusively as one also needs to be just alter what gets sent to the user, not only persisted..

And yes, I wrote already that we can make this always more sophisticated.. As said; this is needed in a critical issue and we really cannot (and don't need to) do a Rolls Royce when we can perfectly live with  VW at the moment..

hairmare:
It looks and feels like legacy and doesn't even add an ``interface``. I'm not budging on this without new arguments and real changes.

You are already abusing this *feature* to delete records on idempotent requests and this really open the floodgates to more legacy behaviours.

I'm really sorry but this just crosses the line in too many ways.

narcoticfresh:
@hairmare 
So please write down an *acceptable* approach here so we know what is expected..

We'll put this on hold then and mark that in the critical issue or split the missing part out of there..

narcoticfresh:
So we agreed to do it as follows.. We want to add to the Generator the abilities to
a) specify a parent class for the generated `Document` class
b) specify which fields should never be persisted to MongoDB

hairmare:
I found this on not persisting but haven't looked up its xml equivalent yet.

http://doctrine-mongodb-odm.readthedocs.org/en/latest/reference/annotations-reference.html#notsaved

PR #96, opened 2015-04-29
Use global composer install

hairmare:
I just stared at a ``Composer\DependencyResolver\SolverProblemsException`` due to an issue that was long gone. Seriously, I'm adding that you should install composer globally to our guidelines ASAP.


lapistano:
to install composer globally see https://getcomposer.org/download/ for further information.

hairmare:
Or use https://github.com/gravity-platform/vagrant-centos-7-php if you're on a non-unix platform :smile_cat:

PR #101, opened 2015-05-07
Added new service definition file

lapistano:


src/Graviton/CoreBundle/Resources/definition/Config.json
```diff
@@ -0,0 +1,61 @@
...
+                "description": "Name of this module",
+                "required": true,
+                "translatable": true
+            },
+            {
+                "name": "actualValue",
+                "type": "varchar",
+                "title": "Current value",
+                "description": "The current value of the parameter.",
+                "required": true,
```
hairmare: Should this really be required?

lapistano: there is no acceptance criteria, nor a description of any field to be mandatory or not. 
We could discuss the 'default Value'. Every other field has to be mandatory, if not it would not make any sense to store a parameter with no value, would it?

lapistano: it further is not defined, if a field shall be translatable. On the second thought, I will remove this definition.

hairmare: From my understanding ``actualValue`` is to be used as a way to override ``defaultValue`` hence me thinking it need not be required.

I also don't think this stuff should be translatable since it should be used under the hood anyway.

src/Graviton/CoreBundle/Resources/definition/Config.json
```diff
@@ -0,0 +1,61 @@
...
+    "id": "Config",
+    "description": "A service containing to admin general configuration parameters.",
+    "service": {
+        "readOnly": false,
+        "routerBase": "/core/config/",
+        "fixtures": [
+            {
+                "id": "123j12j3j123j12j3",
+                "key": "random key",
+                "app_id": "$ref tablet",
```
hairmare: Should be ``http://localhost/core/app/tablet``.

src/Graviton/CoreBundle/Resources/definition/Config.json
```diff
@@ -0,0 +1,61 @@
...
+                "type": "varchar",
+                "title": "ID",
+                "description": "Unique identifier for an item.",
+                "required": true
+            },
+            {
+                "name": "key",
+                "type": "varchar",
+                "title": "Key",
+                "description": "A fixed string identifying this module.",
```
hairmare: needs wording to describe key in this context

src/Graviton/CoreBundle/Resources/definition/Config.json
```diff
@@ -0,0 +1,61 @@
...
+                "required": true
+            },
+            {
+                "name": "app_id",
+                "type": "extref",
+                "title": "Application-Link",
+                "description": "Link to a application.",
```
hairmare: Link to a**n** application.

src/Graviton/CoreBundle/Resources/definition/Config.json
```diff
@@ -0,0 +1,61 @@
...
+            },
+            {
+                "name": "key",
+                "type": "varchar",
+                "title": "Key",
+                "description": "A fixed string identifying this module.",
+                "required": true
+            },
+            {
+                "name": "app_id",
```
hairmare: We've mostly been using ``app.ref`` here, which should lead to this:

```
{
  "app": {
    "$ref": "<url>"
  }
}
```

src/Graviton/CoreBundle/Resources/definition/Config.json
```diff
@@ -0,0 +1,61 @@
...
+                "title": "Application-Link",
+                "description": "Link to a application.",
+                "exposeAs": "$ref",
+                "required": true
+            },
+            {
+                "name": "defaultValue",
+                "type": "varchar",
+                "title": "Default value",
+                "description": "Name of this module",
```
hairmare: describe defaultValue here, I think some string that may be used as input are in the GRV

hairmare:
And then I noticed that this is all WIP *facepalm*

lapistano:
replaced by #107

PR #107, opened 2015-05-10
finish /core/config service

hairmare:
Finishes the work @lapistano did on #101. It also adds some meaningful fixtures that look like the kind of data we would like in this service.

### DIFF
--- a/src/Graviton/CoreBundle/Resources/definition/Config.json
+++ b/src/Graviton/CoreBundle/Resources/definition/Config.json
@@ -1,16 +1,26 @@
 {
     "id": "Config",
-    "description": "A service containing to admin general configuration parameters.",
+    "description": "A key-value store for keeping global configuration state.",
     "service": {
         "readOnly": false,
         "routerBase": "/core/config/",
         "fixtures": [
             {
-                "id": "123j12j3j123j12j3",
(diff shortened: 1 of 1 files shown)


lapistano:
see branch feature/GRV-1263, the current PR #107 does not reflect the acceptance criteria in complete .. 
btw. the verification is breaking my neck here ..

hairmare:
Yeah, we won't be able to do "only existing app can be used" as of now. We need a story to create that as a real feature and it won't be easy.

Looking at the validation criteria in the story it seems we mis-estimated this badly. Maybe we should merge something like in this PR and crea backlog stories on the validation parts.

I looks as though there is still some low-level work to do there.

lapistano:
yeah, I thought of doing this with this story, but you are probably right to extract this into another story.

lapistano:
so we merge this am flush feature/GRV-1263?

hairmare:
@lapistano If it's ok with you.

I've already started on the reference validation bit since that seems to be implementable using a customer validator. My work is on the [feature/app-ref-validation](https://github.com/libgraviton/graviton/tree/feature/app-ref-validation) branch.

lapistano:
ok ... fine with me .. merging ...

PR #114, opened 2015-05-19
Create CustomerDiff Collection in MongoDB

hairmare:
This is for [GRV-1777](https://issue.swisscom.ch/browse/GRV-1777) and adds a ``CustomerDiff`` document and repository.

It also injects the repo into ``AbstractCustomerController`` from #113.

Basically with this we are ready to start implementing some actions in ``AbstractCustomerController``.

I expect this new collection to gain better test coverage and thus become more resilient when we start doing real stuff with the controller.


src/Graviton/PersonBundle/Document/CustomerDiff.php
```diff
@@ -0,0 +1,140 @@
...
+        return $this->date;
+    }
+
+    /**
+     * Set type
+     *
+     * @param string $type type of diff (ie. prospect, ...)
+     * @return self
+     */
+    public function setType($type)
```
lapistano: what about type being a value object with a definition of available types?

src/Graviton/PersonBundle/Document/CustomerDiff.php
```diff
@@ -0,0 +1,140 @@
...
+     */
+    public function getType()
+    {
+        return $this->type;
+    }
+
+    /**
+     * Set original
+     *
+     * @param object $original original version of object
```
lapistano: what object?

src/Graviton/PersonBundle/Document/CustomerDiff.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+
+namespace Graviton\PersonBundle\Document;
+
+/**
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class CustomerDiff
```
lapistano: mo should every setter have a type hint (if no scalar is expected) of what parameter type is expected or at least a validation, if many types are allowed; Value objects with validation would also be a way to go

src/Graviton/PersonBundle/Document/CustomerDiff.php
hairmare: The idea is that this contains a complete copy of the customer. I'm not adding a type since we would need to do a lot of refactoring to get the ``GravitonDyn``-``Customer`` exposed properly. If it wasn't in the ``ml3k-loadconfig`` scope I would reference the real object.

Since we aren't allowing clients to interact directly with the diff-collection at the moment I feel that this is sane.

src/Graviton/PersonBundle/Document/CustomerDiff.php
hairmare: See my above comment about depending on unavailable ``GravitonDyn`` stuff.

src/Graviton/PersonBundle/Document/CustomerDiff.php
hairmare: I wasn't sure if we should only implement the ``prospect`` type we were talking about or if there will also be other types.

My aim with this PR is to get us ready to work on different actions individually. The type field was added since we'll be needing it anyhow but it won't help with implementing any ``*Action`` methods. I'll create a subtask on the main issues for type validation after we merge this.

lapistano: kk

lapistano:
already mered ...

PR #112, opened 2015-05-17
bump dependencies (3rd party and rql-parser-bundle)

hairmare:
le_bump

> OK GitHub... Please bump all composer deps.

* contains rql-parser-bundle 0.2.0
* stabilizes some deps where possible

PR #124, opened 2015-05-26
Make RestUtils testable again by adding an interface

hairmare:
I made RestUtils final in #123 thus introducing breakage in #112. This is now being fixed by adding the ``RestUtilsInterface`` and coding against it.

### DIFF
--- a/src/Graviton/RestBundle/Service/RestUtilsInterface.php
+++ b/src/Graviton/RestBundle/Service/RestUtilsInterface.php
@@ -20,19 +20,6 @@ use Graviton\RestBundle\Controller\RestController;
 interface RestUtilsInterface
 {
     /**
-     * @param ContainerInterface   $container         container
-     * @param Router               $router            router
-     * @param Serializer           $serializer        serializer
-     * @param SerializationContext $serializerContext context for serializer
-     */
-    public function __construct(
-        ContainerInterface $container,
-        Router $router,
-        Serializer $serializer,
-        SerializationContext $serializerContext = null
-    );
-
-    /**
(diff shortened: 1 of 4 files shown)


src/Graviton/RestBundle/Service/RestUtilsInterface.php
```diff
@@ -0,0 +1,119 @@
...
+ */
+interface RestUtilsInterface
+{
+    /**
+     * @param ContainerInterface   $container         container
+     * @param Router               $router            router
+     * @param Serializer           $serializer        serializer
+     * @param SerializationContext $serializerContext context for serializer
+     */
+    public function __construct(
```
samsamann: I think the constructor should be an implementation detail.

hairmare: It's in here because it specifies a contract on what dependencies the implementing class will be need to have. I'm rather undecided on this now. Could you elaborate on the reasoning why you feel it should be an implementation detail?

samsamann: Because it is a fix contract with concrete classes. The class which implements this interface can not extend the constructor. I think a interface should be independent and only define points of intersection.

A factory is concerned about creation of an object and not an interface.

hairmare: Just because Java and C# somehow collectively decided that adding constructor signatures are a bad thing doesn't mean their right. I'm sure as hell not going to change this into a factory... 

I'm not convinced at all, but I'll remove it anyway since I really don't think it matters that much.

PR #126, opened 2015-05-28
Fix camel case services in extref

hairmare:
With these changes using $ref references that point to a service that has a camel-cased collection name in MongoDB should work.

* [x] get rid of ``ucfirst()`` in ``DocumentBundle/Types/ExtReference.php``
* [ ] get rid of ``ucfirst()`` in ``DocumentBundle/DependencyInjection/Compiler/ExtRefMappingCompilerPass.php``

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "libgraviton/codesniffer": "~1.3",
         "lapistano/proxy-object": "~2.1",
         "johnkary/phpunit-speedtrap": "~1.0",
-        "graviton/test-services-bundle": "^0.2"
+        "graviton/test-services-bundle": "^0.3"
     },
     "scripts": {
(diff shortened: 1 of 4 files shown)


hairmare:
@rat82 This won't work, It was tagged as WIP since it still had an open task... I'm not sure if this breaks something though but I hope it shouldn't. Without also fixing ``ExtRefMappingCompilerPass`` the camelcased service ids won't work.

Btw, the code that needs fixing is here https://github.com/libgraviton/graviton/blob/develop/src/Graviton/DocumentBundle/DependencyInjection/Compiler/ExtRefMappingCompilerPass.php#L34. It looks like it'll still be a piece of work to get this done since the compiler-pass currently does not consult any json-defs.

PR #127, opened 2015-05-28
Removed container and created a constructor

RamonB:


src/Graviton/RestBundle/Subscriber/RestEventSubscriber.php
```diff
@@ -104,28 +122,4 @@ public function onKernelResponse(FilterResponseEvent $event, $name, EventDispatc
     {
         $dispatcher->dispatch("graviton.rest.response", $event);
     }
-
-    /**
-     * Get the di container
-     *
-     * @return Container $container DI container
-     */
-    public function getContainer()
```
PanteraLX: Probably we should keep this getter here.

hairmare: You could also try marking the ``RestEventSubscriber`` class as ``final`` and remove it. We needed to keep it in the ``RestController`` due to the fact that other classes where using it.

Since no one should be extending ``RestEventSubscriber`` it should be safe to remove it here.

PR #131, opened 2015-06-03
fix two classes that expect things that are not always there ;-)

narcoticfresh:
making the way towards oauth; those two classes made some problems as they expect things that are not always there as they think.. this PR tries to fix that.

### DIFF
--- a/src/Graviton/RestBundle/Subscriber/RestEventSubscriber.php
+++ b/src/Graviton/RestBundle/Subscriber/RestEventSubscriber.php
@@ -62,8 +62,9 @@ class RestEventSubscriber implements EventSubscriberInterface
     public function onKernelRequest(GetResponseEvent $event, $name, EventDispatcherInterface $dispatcher)
     {
         $restEvent = $this->getEventObject($event);
-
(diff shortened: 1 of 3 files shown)


hairmare:
Please also touch [AuthenticationLoggerTest](https://github.com/libgraviton/graviton/blob/develop/src/Graviton/SecurityBundle/Tests/Listener/AuthenticationLoggerTest.php).

I think we should probably just get rid of the contract thing in the auth-listener. An ''auth'' listener should not need to be concerned about stuff that is clearly in the scope of an acl-listener.

narcoticfresh:
@hairmare I added a test method that tests it with a standard token (not our contract thing which is wrongly named).. is that what you meant?

narcoticfresh:
@hairmare and +1 to the fact that this logger should not be concerned about the contracts.. all over, the "contract" (which is a banking system specific term already) is intermixed with other stuff (and with A*irlock named classes).. yes, we need to separate concerns more, especially to support other stuff easier..

hairmare:
yupp, the '?' is now tested. You could have also just yanked out the whole contract stuff and I'd also been  happy. The contract confusion was the reason that acls failed to get implemented...

PR #132, opened 2015-06-03
removed most container usage

RamonB:


RamonB:
didn't remove some containers which contains implode

src/Graviton/RestBundle/Resources/config/services.xml
```diff
@@ -14,16 +14,16 @@
         <parameter key="graviton.rest.request.class">Symfony\Component\HttpFoundation\Request</parameter>
         <parameter key="graviton.rest.validation.jsoninput.class">Graviton\RestBundle\Validator\JsonInput</parameter>
         <parameter key="graviton.rest.listener.xversionresponselistener.class">Graviton\RestBundle\Listener\XVersionResponseListener</parameter>
-        <parameter key="graviton.rest.services" type="collection"/>
+        <parameter key="graviton.rest.services" type="collection"></parameter>
```
hairmare: These changes are rather nonsense. In most cases the short form is actually preferable since it does not imply an xml node containing an empty string '' but rather a tag without contents. Could you revert 1d7081d56b9a5d7b42b3b0c2d2c152e99ca6bc17 again?

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -31,34 +35,73 @@
...
     private $container;
-
+    
     /**
-     * Get the container object
-     *
-     * @return \Symfony\Component\DependencyInjection\ContainerInterface
+     * @var \Symfony\Component\HttpFoundation\Response
      */
-    public function getContainer()
```
hairmare: We should keep ``getContainer`` for now. There are still some cases of code (ie. in service bundles that still use it).

I would recommend that we annotate it with ``@obsolete`` now and remove it as soon as everything has been de-containerized.

src/Graviton/FileBundle/Controller/FileController.php
```diff
@@ -6,8 +6,15 @@
 namespace Graviton\FileBundle\Controller;
 
 use Graviton\RestBundle\Controller\RestController;
+use Graviton\RestBundle\Service\RestUtilsInterface;
+use Graviton\I18nBundle\Repository\LanguageRepository;
+use Rs\Json\Patch;
```
hairmare: It looks like this line isn't needed here...

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -31,34 +35,73 @@
...
+     */
+    private $router;
+    
+    /**
+     * @var \Graviton\I18nBundle\Repository\LanguageRepository
+     */
+    private $language;
+    
+    /**
+     * @var Symfony\Component\Validator\Validator\ValidatorInterface
```
hairmare: You can use the short form on these just like below in the constructor. If you want to use the full qualified classname in docblocks you need to add the leading ``\`` that is missing here and also in ``$templating``.

I'd recommend that you just use this:
```php
/**
 * @var ValidatorInterface
 */
```

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -526,7 +582,7 @@ protected function validateRecord($record)
      */
```
hairmare: The ``@return`` on this should now be ``@return ValidatorInterface``...

src/Graviton/RestBundle/Tests/Controller/RestControllerTest.php
```diff
@@ -95,19 +97,31 @@ public function testValidateRecord()
...
-            ->expects($this->once())
-            ->method('get')
-            ->with($this->equalTo('graviton.rest.validator'))
-            ->will($this->returnValue($validatorMock));
-
         $record = $this->getMock('\Graviton\CoreBundle\Document\App');
 
-        $controller = new RestControllerProxy();
-        $controller->setContainer($containerMock);
+        $controller = new RestControllerProxy(
```
lapistano: die folgenden 20 Zeilen ist ein Duplikat von Zeile 46ff. Es wäre gut, wenn diese Zeilen in einer Methode consolidiert würden.

PR #133, opened 2015-06-03
Feature/refactor swagger generate command

janfriedli:
indenting fixed

### DIFF
--- a/src/Graviton/SwaggerBundle/Command/SwaggerGenerateCommand.php
+++ b/src/Graviton/SwaggerBundle/Command/SwaggerGenerateCommand.php
@@ -8,6 +8,7 @@ namespace Graviton\SwaggerBundle\Command;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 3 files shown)


gfellerleonard:
When we tried to inject the service as a param we must change a symfony class.
So we stop at this point. A complet refactoring of the swagger is necessary. It should not depend on a request.

hairmare:
Yupp, what Leo said is the sad truth. We need to fully refactor the generator so it works without being dependant on a request. There is some extremely silly code in ``SwaggerGenerateCommand::execute`` that works around ``getSwaggerSpec`` being dependant on the request by doing some rather unholy things with the container.

In fact there is no real reason why generating a swagger spec should need a request at all. If we get rid of that dependency we can re-open this and finish it where we are now closing it.

For reference, [this](https://github.com/libgraviton/graviton/commit/3fbaf4149cc7c3443404c0b25fe0defe5a925e7c#diff-cc97a440812802d8e85bb303c312b2b5R132) is the code I'm talking about and [this](https://github.com/libgraviton/graviton/blob/70c2bcdd1bf06f1bd61782f947f947433f283146/src/Graviton/SwaggerBundle/Service/Swagger.php) is what seems to need refactoring.

PR #134, opened 2015-06-04
oAuth Client integration

narcoticfresh:
This PR integrates the `hwi/oauth-bundle` bundle that uses oAuth for authentication..

I need to discuss how we integrate the multiple Authentication scenarios that Graviton has to support.. Currently, we have 3:
- no auth (local dev)
- oAuth (as in cloud and possibly more)
- A*rlock (some clients)

In this PR; i create a custom environment `oauth_dev` (`oauth_prod` will follow) that combines dev with oauth features.. Please provide some feedback if this approach is the correct one.

In reality, it works really fine.. I can run my local dev Graviton in `oauth_dev` as follows:
```
SYMFONY_ENV=oauth_dev php app/console --router=src/Graviton/CoreBundle/Resources/config/router_env.php server:run
```

If agreed, I will also split out all A*rlock stuff out and create custom configs for that (and an environment), so we have a clean line.

Obviously, we then also need to adress how we can make (IMHO best) a single Authorization layer that works with any Authentication thing we put in front of it - but that's a different discussion.. ;-)


app/config/oauth/config.yml
```diff
@@ -0,0 +1,13 @@
+imports:
+    - { resource: security.yml }
+
+hwi_oauth:
+    firewall_name: secured_area
+    resource_owners:
+        github:
+            type:                github
+            client_id:           "4495acf8e4aa650814a8"
```
lapistano: client_id and client secret should be a parameter handled by incenteev.

narcoticfresh:
@hairmare @lapistano so you guys agree that we should pursuit the environment-based approach? or any other ideas?

.gitignore
```diff
@@ -15,3 +15,4 @@ app/config/parameters.yml
 app/config/parameters_local.xml
 bin
 app/files/
+.idea
```
hairmare: Please remove this line after running the following on your dev box.

```bash
git config --global --add core.excludesfile ~/.gitignore_global
echo '.idea' >> ~/.gitignore_global
```

app/config/parameters.yml.dist
```diff
@@ -5,8 +5,10 @@ parameters:
 
     graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
     graviton.authentication.user_provider.model: ~
-    graviton.security.services.whitelist: ~
+    graviton.security.services.whitelist: {}
     graviton.rest.special_mimetypes: {}
+    graviton.security.oauth.github.client_id: "4495acf8e4aa650814a8"
+    graviton.security.oauth.github.client_secret: "5754f5b8c4e9e4a10dd55d7d0c89383fa2a1704f"
```
hairmare: Please make this injectable through env in composer.json and reset the client secret in github ASAP. These two lines should never have been committed and I'll do the reset myself if it's still open tomorrow.

hairmare:
@narcoticfresh having multiple envs is ok as handling more of this in config would also be. Maybe @lapistano has some additional insights on this.

What I'm not happy with are things like ``src/Graviton/CoreBundle/Resources/config/router_oauth_dev.php`` and ``web/app_oauth_dev.php``. The former needs way more explaining (where is it from, why do we need it, ...) and the latter should just be handled by the existing ``app.php`` and ``app_dev.php`` (We'll have to make those support ``SYMFONY_ENV`` again).

narcoticfresh:
@hairmare 

**Secrets**
Well, you're right about the secrets.. Those are just test apps, so after the Jira Issue we'll remove those apps.. In Github, every app is bound to 1 callback url - this one is locked to the test cloud app that i will remove after the sprint review..

Which leads me to the question: Where do we put it after? We also cannot put them into composer.json or anywhere in Git.. so how and where to put(?)..

Btw, all Symfony params are by default settable by ENV, no need to enable that.. [See this link](http://symfony.com/doc/current/cookbook/configuration/external_parameters.html)

**Routers and web/* files**
I thought about making app_dev.php and app_prod.php ENV aware, but the Symfony docs tell you to do otherwise, so I'm not sure if there is anything I'm missing. But I agree with you..

The `router_oauth_dev.php` seems to be needed when you use `server:run` in `app/console` - the FrameworkBundle looks for a file named `router_{kernel.environment}.php` in the `Resources/config` folder - if there is no such file, an error is displayed. Then you need to specify your own router.. But also there, we could do an ENV aware file..

hairmare:
**envs**

regarding the symfony envs... I know about the way to override vars you are linking, that's ok to use for some quick hacks on dev or really exotic use cases. For a default use case the configuration should work with easy to read env variables through Incenteev/ParameterHandler [as describe in the docs](https://github.com/Incenteev/ParameterHandler#using-environment-variables-to-set-the-parameters).

This enables some use cases.
* developers get prompted for the variable during ``composer install``, not specifying a value should deactivate oauth.
* customers can use their configuration management system to deploy the app including a parameters.yml and will get notified about (potentially unconfigured) new parameters during ``composer install``
* cloudy deploys can use ``cf set-env`` or ``docker run -e`` to override variables allowing us to customize everything

In the end this allows us to control configuration in a much more managed fashion than using SYMFONY__ style variables (which in turn are not compatible with parameters.yml and lead to issues during the last couple symfony version bumps).

**router / web***

I don't think the router does any harm and I could even see us adding it to dev all the time. It works around a legit issue and everything. We need to ensure that it is only active wen using the built in server and it also need some more docs pointing to where it came from exactly (i'm assuming there is also something on SO and not only the php bug that is already linked).

Maybe we should read the env in web files, or we should I'm not sure on that one. If we don't read the env their it looks like we're going to have heaps of the files and creating something more abstract might make sense.

narcoticfresh:
> not specifying a value should deactivate oauth

@hairmare 
i'm afraid this is not possible (or not that easy to be precise). to bundle doesn't have an "on/off switch" - loading it activates it. if you load it, it assumes some security params which in turn activates it. the only way to don't use it while being loaded is obviously to not use it in security.yml in the firewall - but i don't know that this can be changed dynamically. maybe compiler pass.

anyway, i think it's clearer to explicitly use execution env's. if you want oauth - you enable an oauth_* env and there you go. then it's also up to you to set the corresponding param values..

i will add more flexibility to the app_* files in this PR..

narcoticfresh:
@hairmare 
OK; I flexibilized the whole `app_*.php` thing. now we can add more envs without custom routers and entry points in `web/`. I also adapted the cloud push to use the new `app_env.php` point and set the symfony env with a env var..

the oauth_dev can now used as follows on the command line:
```
SYMFONY_ENV=oauth_dev php app/console --router=src/Graviton/CoreBundle/Resources/config/router_env.php server:run
```

**Where?** I'd like to add some Doc and list the existing envs somewhere. Is readme.md in the root the correct place or where?

After the doc; I think we can safely merge this. It doesn't affect anything unless the `oauth_dev` env is used.. Sadly, in order to create the `oauth_prod` env; I first need to pull out all the A*rlock stuff out of the current `prod` env.. I'd like to clean that up in a following PR..

app/AppKernel.php
```diff
@@ -72,10 +72,10 @@ public function registerBundles()
             new \Exercise\HTMLPurifierBundle\ExerciseHTMLPurifierBundle(),
             new \Graviton\RqlParserBundle\GravitonRqlParserBundle(),
             new \Knp\Bundle\GaufretteBundle\KnpGaufretteBundle(),
-            new \Eo\AirbrakeBundle\EoAirbrakeBundle(),
+            new \Eo\AirbrakeBundle\EoAirbrakeBundle()
```
hairmare: please revert this change...

app/config/config_oauth_dev.yml
```diff
@@ -0,0 +1,8 @@
+imports:
+    - { resource: config_dev.yml }
+    - { resource: oauth/config.yml }
+
+framework:
+    router:
+        resource: "%kernel.root_dir%/config/routing_oauth_dev.yml"
+        strict_requirements: true
```
hairmare: please add a newline to end of this file...

hairmare:
Doc are usually somewhere in ``app/Resources/doc``. It might make sense to start writing an AUTHENTICATION.md or something there. ``src/Graviton/SecurityBundle/README.md`` would be another great place to add some docs...

hairmare:
Also we need to find a way to not make ``SYMFONY_ENV=oauth_dev php app/console --router=src/Graviton/CoreBundle/Resources/config/router_env.php server:run`` be such a horribly looking command :(

What do you think about always loading the router env and making it be env sensitive itself?

app/config/security_dev.yml
```diff
@@ -1,22 +1,22 @@
...
 #    # http://symfony.com/doc/current/book/security.html#where-do-users-come-from-user-providers
-#    providers:
-#
-#        graviton_airlock_authentication_key_user_provider:
-#            id: graviton_airlock_authentication_key_user_provider
-#
+    providers:
+        default:
+            id: graviton_airlock_authentication_key_user_provider
+
```
hairmare: Won't this make development without a security context harder?

composer.json
```diff
@@ -48,7 +48,8 @@
         "graviton/rql-parser-bundle": "^0.2@alpha",
         "knplabs/knp-gaufrette-bundle": "^0.2@dev",
         "aws/aws-sdk-php": "~2.7",
-        "eo/airbrake-bundle": "~1.0"
+        "eo/airbrake-bundle": "~1.0",
+        "hwi/oauth-bundle": "0.4.*@dev"
```
hairmare: Please change this to use a a released version of the bundle or explain what kind of features in 0.4 are essential to our integration.

src/Graviton/CoreBundle/Resources/config/router_env.php
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * This file implements rewrite rules for PHP built-in web server.
+ *
+ * See: http://www.php.net/manual/en/features.commandline.webserver.php
+ *
+ * If you have custom directory layout, then you have to write your own router
+ * and pass it as a value to 'router' option of server:run command.
+ *
+ * @author: Michał Pipa <michal.pipa.xsolve@gmail.com>
+ * @author: Albert Jessurum <ajessu@gmail.com>
```
hairmare: Could you add a link to where you got this file from? I feel we should keep track of that due to licensing purposes.

hairmare:
One last remark... I think you should move ``web/app_env.php`` to ``web/app.php`` and make it default to prod. That way we don't need the putenv hack that currently is in app.php and app_env is fully obsolete.

.bp-config/php/php-fpm.conf
```diff
@@ -490,6 +490,9 @@ ping.response = pong
 ; build pack rewrites this to the list of environment variables at compile time
 ;  produces a list like "env[<var>] = $<var>"
 #{PHP_ENV}
+
+env[SYMFONY_ENV] = dev
```
hairmare: We might also consider using ``clear_env = no`` instead of whitelisting individual vars. It might be worth checking if the buildpack exposes some way to populate the ``#{PHP_ENV}`` placeholder here.

composer.json
narcoticfresh: The author states first in the doc:

> All the installation instructions are located in the documentation, check it for specific version:
> 0.4 with support for Symfony >=2.3
> 0.3 with support for Symfony >=2.1,<2.7

Currently, there are no release for the 0.4.* versions.. So we're stuck with `@dev`, but we can switch to release as soon as available..

hairmare: We are on symfony 2.6.7, so 0.3 should work...

narcoticfresh: Oh please don't do this, the config is quite different and i'll need to re-check everything ;-( - 0.4.* is the branch where future work will be done and it works fine with us; why not stay on that and jump on a release when it's there?

.bp-config/php/php-fpm.conf
narcoticfresh: @hairmare 
Hm, strange.. I looked up the buildpack docs; but can't find anything in regard to `#{PHP_ENV}`..

Also, when I check the current file the buildpack generates for use of php_fpm, it seems to be untouched:
```
 cf f graviton-develop-env app/php/etc/php-fpm.conf
[...]
; build pack rewrites this to the list of environment variables at compile time
;  produces a list like "env[<var>] = $<var>"
#{PHP_ENV}
```
Seems to be an abandoned feature(?)

composer.json
hairmare: We should always prefer using stable releases of things. If not the need should be clearly documented and explained. The problem with using dev-master now is that it is very much in flux and hasn't had a tagged version at all (not even a pre-release).

Overall this also depends on how soon the bundle gets a real release, until then we are going to introduce possible breakage every time we update composer.lock. That might become a drag rather quickly...

.bp-config/php/php-fpm.conf
hairmare: This was indeed removed by upstream and replaced with the ``clean_env = no`` feature.

Upstreams file is [here](https://github.com/cloudfoundry/php-buildpack/blob/master/defaults/config/php/5.6.x/php-fpm.conf). You could replace ours with that if you want to (or even check if it ain't needed anymore).

narcoticfresh:
@hairmare 
Hm, it seems not that easy to change the default router used in `server:run`.. please [see this code](https://github.com/symfony/FrameworkBundle/blob/2.7/Command/ServerRunCommand.php#L152).. any ideas how to inject something?

narcoticfresh:
@hairmare 
OK; I set `SYMFONY_ENV` through `manifest.yml` now (defaults to `dev` for cloud push) and I updated the `php-fpm.conf`. I couldn't remove our injection for the `CF` env vars, this would have broken the connection and I don't want to mess with this now ;-)

narcoticfresh:
@hairmare OK; fixing the oauth-bundle to the 0.3 releases seems to have no functional change; so I'm fixing this on the release.. that's preferable of course..

README.md
```diff
@@ -22,6 +22,18 @@ and
 app/console
 ```
 
+## Custom Environments
+
+Some bundles (as `SecurityBundle`) support custom environments (please refer to their documentation for specifics).
+
+To support other environments then the standard ones, Graviton provides an `ENV` aware router you can use like this:
```
hairmare: app/console and the env aware router docs should be in app/Resource/doc/DEVELOPMENT.md... IMHO this is a bit much for the main README.md which should mainly contain pointers to more docs.

src/Graviton/SecurityBundle/README.md
```diff
@@ -106,12 +109,12 @@ Example:
...
+
 - \Graviton\CoreBundle\Repository\AppRepository
 - \Graviton\RestBundle\Model\ModelInterface ( this should be resolved asap)
 - \Symfony\Component\Security\Core\User\UserInterface
 
-Future things
--------------
+# Future things
+
 - add command to find out what strategies are available.
```
hairmare: Thanks for cleaning this up into the nicer markdown syntax :+1:

hairmare:
@narcoticfresh Please move the docs you're adding out of the main README as discussed and merge develop, then I'll merge this...

narcoticfresh:
@hairmare OK; done..

PR #137, opened 2015-06-09
Use proper HttpException as base

hairmare:
This makes our home-baked exceptions use symfonys HttpException as base.

In itself this does not change much, it's part of the larger picture of us refactoring parts to use more framework and less home-made stuff.

### DIFF
--- a/src/Graviton/ExceptionBundle/Exception/NotFoundException.php
+++ b/src/Graviton/ExceptionBundle/Exception/NotFoundException.php
@@ -5,6 +5,7 @@
 
 namespace Graviton\ExceptionBundle\Exception;
 
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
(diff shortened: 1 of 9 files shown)


src/Graviton/ExceptionBundle/Exception/RestException.php
```diff
@@ -14,36 +14,7 @@
...
-    /**
-     * Get the response object
-     *
-     * @return Response $response Response object
-     */
-    public function getResponse()
-    {
-        return $this->response;
-    }
+    use RestExceptionTrait;
```
lapistano: is the trait used elsewhere? If not, I think this is not necessary to abstract.

hairmare: It's also in ``NotFoundException`` which now directly extends ``Symfony\Component\HttpKernel\Exception\NotFoundHttpException``, hence the trait.

src/Graviton/ExceptionBundle/Exception/NotFoundException.php
```diff
@@ -14,8 +15,10 @@
  * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
  * @link     http://swisscom.ch
  */
-final class NotFoundException extends RestException
+final class NotFoundException extends NotFoundHttpException implements RestExceptionInterface
 {
+    use RestExceptionTrait;
```
hairmare: ie. here...

PR #147, opened 2015-06-19
Work on php 7 and hhvm on travis

hairmare:
Both are allowed to fail (and even do so for various reasons) and both get a bit further along than before.

This should help us keep track of when mongodb/doctrine stops blocking both platforms.


lapistano:
cannot be merged. Probably rebase to develop necessary?

hairmare:
rebased

PR #149, opened 2015-06-22
Bump rql-parser-bundle and rql-parser

hairmare:
* [x] basic integration
* [ ] nice exception handling (maybe in bundle...)


hairmare:
I haz internal issue [GRV-1801](https://issue.swisscom.ch/browse/GRV-1801) which includes finishing the rql stuff and will be pushing a new bap tag with that as soon as we get graviton-next out.

That might take some PRs since we want to finish automating all bap deploys in that window as well.

PR #151, opened 2015-06-23
use deploy scripts

hairmare:
This switches graviton to use the new deploy scripts... I'll need to change it in jenkins after we merge this as well... Jenkins will be less complex after this change :)


app/config/deploy.yml.dist
```diff
@@ -0,0 +1,14 @@
+# this file contains the main configuration for the library
+deploy-scripts:
+  cf_bin: cf
+  cf_process_timeout: 3600
```
lapistano: the timeout is set to an hour (60x60sec), I guess this is by intention?

hairmare: hmmm... 1 hr might be a bit much... I'll make it a bit smaller...

PR #155, opened 2015-06-24
Bump to graviton/deploy-scripts-0.2.1

hairmare:


composer.json
```diff
@@ -49,7 +49,7 @@
         "aws/aws-sdk-php": "~2.7",
         "eo/airbrake-bundle": "~1.0",
         "hwi/oauth-bundle": "~0.3",
-        "graviton/deploy-scripts": ">=0.2.0"
+        "graviton/deploy-scripts": ">=0.5.0"
```
narcoticfresh: hmm.. it seems composer installed `dev-master` now..(?).. (see lock file). 0.5.0 tag doesn't seem to be present in `deploy-tools` - forgot to push your tags?

hairmare: It should be ``>=0.2.1``, update incoming...

PR #157, opened 2015-06-29
Feature/grv 1820

lapistano:
splits the monster function `GenerateDynamicBundleCommand::execute()` on [github](https://github.com/libgraviton/graviton/blob/develop/src/Graviton/GeneratorBundle/Command/GenerateDynamicBundleCommand.php#L104) in to testable and probable reusable pieces. 
Do so it introduces:
* [x] `CommandRunner` » using `Symfony\Component\Process\Process` to execute console commands
* [x] `XmlManipulator` » providing the possibility to alter XML Documents


src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
```diff
@@ -51,6 +51,7 @@ public function __construct($filename)
     {
         $this->filename = $filename;
 
+        // TODO [lapistano]: THIS SHALL NOT BE HERE!!
```
hairmare: We might be able to rewrite this to lazy loading... Otherwise I see no fast way to get the code out of the constructor...

src/Graviton/GeneratorBundle/Tests/Resources/Definition/testDefinition.json
```diff
@@ -0,0 +1,40 @@
+{
```
hairmare: Are you sure that this definition won't be found during ``app/console graviton:generate:dynamicbundles``? Or should it?

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
lapistano: I tried to estimate the damage it would do if the $filename would be changed to a $jsonText.
This is why I just added the todo ;) 

But I agree, that a JsonDefinition::extractJson($filename) would be a first step.

src/Graviton/GeneratorBundle/Tests/Resources/Definition/testDefinition.json
lapistano: I run a lot of these commands and never found it getting any attention, nor a dyn bundle appearing

hairmare: Works for me... Doesn't really speak for the sanity of the loader, but meh...

hairmare:
@lapistano is this ok to merge? It looks good but still has the WIP tag...

lapistano:
Ich dachte ich hätte das entfernt. Ja ist ready to merge

PR #166, opened 2015-07-06
Generate date and datetime fields properly

hairmare:
Make date and datetime fields validate properly...

### DIFF
--- a/src/Graviton/DocumentBundle/Form/Type/DocumentType.php
+++ b/src/Graviton/DocumentBundle/Form/Type/DocumentType.php
@@ -70,10 +70,8 @@ class DocumentType extends AbstractType
                 $type = clone $this;
                 $type->initialize($options['data_class']);
             } elseif ($type === 'date' || $type == 'datetime') {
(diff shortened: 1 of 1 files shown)


src/Graviton/DocumentBundle/Form/Type/DocumentType.php
```diff
@@ -69,6 +69,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             if ($type == 'form') {
                 $type = clone $this;
                 $type->initialize($options['data_class']);
+            } elseif ($type === 'date' || $type == 'datetime') {
+                echo 'got form type and making string of it';
+                $options['widget'] = 'single_text';
+                $options['input'] = 'string';
+                $options['invalid_message'] = 'lerrooooooor';
```
lapistano: what about a message like: "The provided string is not a date nor a datetime."?

hairmare: I went back to using the default message... Would that be ok?

PR #169, opened 2015-07-07
Don't return body in OPTIONS requests

hairmare:
tahdah! This should fix [GRV-1915](https://issue.swisscom.ch/browse/GRV-1915).


src/Graviton/I18nBundle/Tests/Controller/TranslatableControllerTest.php
```diff
@@ -27,9 +27,6 @@ public function testOptionsHasCors()
 
         $client->request('OPTIONS', '/i18n/translatable/i18n-de-German');
         $this->assertCorsHeaders('GET, POST, PUT, DELETE, OPTIONS', $client->getResponse());
-
-        $results = $client->getResults();
-        $this->assertContains('id', $results->required);
```
narcoticfresh: please make an assertion that body is empty..

PR #171, opened 2015-07-08
added OPTIONS to canonical routes

janfriedli:


hairmare:
I just looked into merging this and realized that it is also being fixed by my pr #169.

Since #169 would overwrite the changes from here as well, I'm closing this in favor of the latter.

PR #174, opened 2015-07-09
coded against interface instead of concret implementation

RamonB:
addViolation is found in ExecutionContextInterface and buildViolation in ConstraintValidator. It's better to code against an interface instead of a concret class.


hairmare:
sweet :+1:

PR #175, opened 2015-07-09
The Dollarsign ($ref) in the required field was missing. 

PanteraLX:


src/Graviton/SchemaBundle/SchemaUtils.php
```diff
@@ -128,7 +128,16 @@ function ($language) {
             $schema->addProperty($field, $property);
         }
 
-        $schema->setRequired($model->getRequiredFields());
+        $requiredFields = $model->getRequiredFields();
+        $countFields = count($requiredFields);
+        for ($index = 0; $index < $countFields; $index++) {
```
hairmare: Please us ``foreach`` instead of ``for``.

```
foreach ($requiredFields as $index => $field) {
}
```

PR #180, opened 2015-07-10
Feature/grv-1955-post-with-id

gfellerleonard:
The problem was that it did an UPDATE request when the user did a POST request with an ID in the payload, what is actually not right. This should solve the problem and returns an error message by validating the request.

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
@@ -245,6 +245,25 @@ class AppControllerTest extends RestTestCase
     }
 
     /**
+     *
+     */
+    public function testPostWithId()
+    {
+        $helloApp = new \stdClass();
+        $helloApp->id = 101;
+        $helloApp->name = "tubel";
+
+        $client = static::createRestClient();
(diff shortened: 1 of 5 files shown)


narcoticfresh:
(i want to hold this back until we have a stable release for bap)..

hairmare:
I'm :+1: on this change and don't feel we should be holding it back for any reason whatsoever.

This PR still needs to add some tests to show that it really is working...

src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
```diff
@@ -245,6 +245,30 @@ public function testPostMalformedApp()
...
+        $helloApp->name = "tubel";
+
+        $client = static::createRestClient();
+        $client->post('/person/customer', $helloApp);
+        $response = $client->getResponse();
+        $results = $client->getResults();
+
+        $this->assertEquals(
+            'Can not be given on a POST request. Do a PUT request instead to update an existing record.',
+            json_decode($response->getContent())[0]->message
```
hairmare: You can use ``$results[0]->message`` that already does the decoding.

RamonB:
new https://github.com/libgraviton/graviton/pull/188

PR #181, opened 2015-07-10
make PUT to behave like POST

narcoticfresh:
as was noted by our customers, PUT now behaves different from POST (returning content instead of Location/Link header)..

This PR tries to make PUT behavior the same as POST. 

This has some urgency, as (in a slightly unrelated matter) some services (as `Consultation`) fail to get there content back on the current PUT implementation is the Serializer struggles to deserialize what comes from Symfony form (ie some dates now seem to be String instead of DateTime)..

@hairmare There seems mystery is happening in `FileControllerTest` ("make array of things writable"). Do you have any insight?  Somehow, when the `links` array is altered and PUT'ted, then GET'ted again, the `links` array has two items - which lead the test to fail.

Obviously, that test needs to pass - but I struggle to understand why the changes in `RestController` could render "make array of things writable" broken - any ideas?

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
@@ -263,17 +263,23 @@ class AppControllerTest extends RestTestCase
         $response = $client->getResponse();
         $results = $client->getResults();
 
-        $this->assertResponseContentType(self::CONTENT_TYPE, $response);
+        // we sent a location header so we don't want a body
+        $this->assertNull($results);
+        $this->assertContains('/core/app', $response->headers->get('Location'));
 
-        $this->assertEquals('tablet', $results->id);
+        $client = static::createRestClient();
+        $client->request('GET', $response->headers->get('Location'));
+        $response = $client->getResponse();
+        $results = $client->getResults();
+
+        $this->assertResponseContentType(self::CONTENT_TYPE, $response);
         $this->assertEquals('Tablet', $results->title->en);
         $this->assertFalse($results->showInMenu);
-
         $this->assertContains(
-            '<http://localhost/core/app/tablet>; rel="self"',
+            '<http://localhost/core/app/'.$results->id.'>; rel="self"',
             explode(',', $response->headers->get('Link'))
         );
-        $this->assertEquals('*', $response->headers->get('Access-Control-Allow-Origin'));
+
     }
 
     /**
@@ -321,7 +327,13 @@ class AppControllerTest extends RestTestCase
         $response = $client->getResponse();
         $results = $client->getResults();
 
-        $this->assertResponseContentType(self::CONTENT_TYPE, $response);
+        // we sent a location header so we don't want a body
+        $this->assertNull($results);
+        $this->assertContains('/core/app', $response->headers->get('Location'));
+
+        $client = static::createRestClient();
+        $client->request('GET', $response->headers->get('Location'));
+        $results = $client->getResults();
 
         $this->assertEquals('tablet', $results->id);
         $this->assertEquals('New tablet', $results->title->en);
--- a/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
@@ -276,7 +276,7 @@ class AppControllerTest extends RestTestCase
         $this->assertEquals('Tablet', $results->title->en);
         $this->assertFalse($results->showInMenu);
         $this->assertContains(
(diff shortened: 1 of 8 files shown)


src/Graviton/FileBundle/Controller/FileController.php
```diff
@@ -183,8 +183,26 @@ public function putAction($id, Request $request)
...
+        $request->attributes->set('id', $record->getId());
+
+        $response = $this->getResponse();
+        $response->setStatusCode(Response::HTTP_OK);
+
+        $routeName = $request->get('_route');
+        $routeParts = explode('.', $routeName);
+        $routeType = end($routeParts);
+
+        if ($routeType == 'put') {
```
hairmare: I would be easier to just check if the last chars from ``$request->get('_route');`` are ``".put"``.

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -412,13 +412,27 @@ public function putAction($id, Request $request)
...
 
-        return $this->render(
-            'GravitonRestBundle:Main:index.json.twig',
-            ['response' => $this->serialize($record)],
-            $response
+        $routeName = $request->get('_route');
+        $routeParts = explode('.', $routeName);
+        $routeType = end($routeParts);
+
+        if ($routeType == 'put') {
```
hairmare: Use ``substr($request->get('_route'), -4) == '.put')`` or something similar?

src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
```diff
@@ -263,17 +263,23 @@ public function testPutApp()
...
+        $response = $client->getResponse();
+        $results = $client->getResults();
+
+        $this->assertResponseContentType(self::CONTENT_TYPE, $response);
         $this->assertEquals('Tablet', $results->title->en);
         $this->assertFalse($results->showInMenu);
-
         $this->assertContains(
-            '<http://localhost/core/app/tablet>; rel="self"',
+            '<http://localhost/core/app/'.$results->id.'>; rel="self"',
```
hairmare: Please leave this hardcoded as it was. It really isn't the server that gets to decide what the id should be.

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -412,13 +412,24 @@ public function putAction($id, Request $request)
         } else {
             $this->getModel()->updateRecord($id, $record);
         }
+
+        // store id of new record so we dont need to reparse body later when needed
+        $request->attributes->set('id', $record->getId());
```
hairmare: Do we need this?

hairmare:
@narcoticfresh 97468f7 is going to fail. Some issues where fixed but I still need to figure out the "remove-the-last-record-in-a-collection" edge case.

narcoticfresh:
@hairmare I tried to dig in.. it seems that even though we have an empty `ArrayCollection` present in the `links` attribute we're trying to persist with OM; it doesn't get sent that way to MongoDB..

So we have this after `checkForm()` in `RestController`:

```
object(GravitonDyn\FileBundle\Document\File)[553]
  protected 'id' => string '55a215884046a4792d8b4569' (length=24)
  protected 'deletedDate' => null
  protected 'links' => 
    object(Doctrine\Common\Collections\ArrayCollection)[542]
      private 'elements' => 
        array (size=0)
          empty
```
And this is the query sent to Mongo (dev.log):
```
[2015-07-12 09:29:01] doctrine.DEBUG: MongoDB query: {"update":true,"query":{"_id":"55a215884046a4792d8b4569"},"newObj":{"$set":{"metadata":{"_id":"55a2173d4046a4792d8b4573","size":"4","mime":"text\/plain","filename":"hans"}}},"options":{"upsert":true},"db":"db","collection":"File"} [] []
```

There are 2 things that bug me:
* ODM sends `upsert: true` to mongo - but `$upsert` in `RestController` is `false`
* The `links` attribute is not set on query - and `upsert: true` leaves it there I guess..

I really don't understand why Doctrine behaves that way..

hairmare:
@narcoticfresh This has been bugging me since I realized it as well.

As you found out the form is really doing what it should and record contains an empty collection of links :point_right: I expect it to be empty during saving.

I actually hacked the same kind of test directly into doctrine/mongo-odms testsuite. It worked as intended there.

I hadn't noticed the ``upsert`` so far and will look into that next.

hairmare:
@narcoticfresh Here we go :) It looks like 9669459 finally fixes the issue.

hairmare:
@narcoticfresh look like we can merge this :dancers:

PR #182, opened 2015-07-10
fix sort: apply _id sort after rql sort..

narcoticfresh:
see title and comment..

```
* apply our default sort *after* rql so it's not the dominant one.
* doing it first will *disable* all sorting after as they are stacked.
* this one only disables to sort id,desc - but it's the best we can do right now.
```

### DIFF
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -94,9 +94,7 @@ class DocumentModel extends SchemaModel implements ModelInterface
 
         /** @var \Doctrine\ODM\MongoDB\Query\Builder $queryBuilder */
         $queryBuilder = $this->repository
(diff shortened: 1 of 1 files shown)


src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -115,6 +113,15 @@ public function findAll(Request $request)
         $queryBuilder->skip($startAt);
         $queryBuilder->limit($numberPerPage);
 
+        /**
+         * apply our default sort *after* rql so it's not the dominant one.
+         * doing it first will *disable* all sorting after as they are stacked.
+         * this one only disables to sort id,desc - but it's the best we can do right now.
```
hairmare: See #184 for how we can do better ;)

narcoticfresh:
closed in favor of #184

PR #185, opened 2015-07-13
make sure we don't override rql limit()

narcoticfresh:
currently, the applying of the default value of `perPage` *always* overrides whatever is passed by RQL.. Thus, limiting via RQL is not possible at all currently..

IMHO rql `limit()` should always be the master and should not be overriden by `GET[perPage]`, this PR implements that..

### DIFF
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -116,7 +116,7 @@ class DocumentModel extends SchemaModel implements ModelInterface
         if (!array_key_exists('limit', $queryBuilder->getQuery()->getQuery())) {
             $queryBuilder->limit($numberPerPage);
         } else {
-            $numberPerPage = (int) $queryBuilder->getQuery()->getQuery();
+            $numberPerPage = (int) $queryBuilder->getQuery()->getQuery()['limit'];
         }
 
         /**
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -111,7 +111,9 @@ class DocumentModel extends SchemaModel implements ModelInterface
         }
 
         // define offset and limit
-        $queryBuilder->skip($startAt);
+        if (!array_key_exists('skip', $queryBuilder->getQuery()->getQuery())) {
+            $queryBuilder->skip($startAt);
+        }
 
         if (!array_key_exists('limit', $queryBuilder->getQuery()->getQuery())) {
             $queryBuilder->limit($numberPerPage);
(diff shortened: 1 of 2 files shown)


src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -109,9 +109,15 @@ public function findAll(Request $request)
...
         }
+
         // define offset and limit
         $queryBuilder->skip($startAt);
-        $queryBuilder->limit($numberPerPage);
+
+        if (!array_key_exists('limit', $queryBuilder->getQuery()->getQuery())) {
+            $queryBuilder->limit($numberPerPage);
+        } else {
+            $numberPerPage = (int) $queryBuilder->getQuery()->getQuery();
```
hairmare: huh?

narcoticfresh: wat?

narcoticfresh: ou.. damn... ok..

hairmare: You are casting an array to int here...

narcoticfresh:
@hairmare that was a nice one :+1:

narcoticfresh:
hmm.. looking at this, rql `limit()` as actually not the same as the purpose of `GET[perPage]`.. 

`perPage` concerns only paging whereas `limit()` should limit the length of the entire result of paging, meaning the 'total number of records', obviously also impacting paging.. but i somehow see no way how to distinguish and implement this..

the current state that rql `limit()` has no effect is not ideal at all and confusing users..

@hairmare what is your opinion on that?

hairmare:
@narcoticfresh I see the following on [persvr/rql](https://github.com/persvr/rql).

> limit(count,start,maxCount) - Returns the given range of objects from the result set

Why would this not be like perPage?

ie.

limit(10) :point_right: page 1, docs 1-10
limit(10,10) :point_right: page 2, docs 11-20

So it's basically like ``perPage`` and ``page`` with slightly different syntax. We might need to change the code that's applying it but only tests will tell ;)

hairmare:
Also 'limit' in ``->getQuery()->getQuery()`` might be something totally different...

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -109,9 +109,15 @@ public function findAll(Request $request)
             /** @var \Doctrine\ODM\MongoDB\Query\Builder $qb */
             $queryBuilder->find($this->repository->getDocumentName());
         }
+
         // define offset and limit
         $queryBuilder->skip($startAt);
```
hairmare: Looks like you also need to take care of this if you want it to work.

narcoticfresh:
@hairmare ok, agreed - didn't know about the 2nd param of `limit`.. I guess it's the same now.. I added more assertions..

PR #195, opened 2015-07-15
ShowCase Unit Test

narcoticfresh:
this is a PR that should cover the Generator results by testing our `ShowCase.json` service.

Results found, stuff we need to fix/decide
- [x] **Current fail**: If you don't POST a translatable at all; Graviton returns it empty (`{text: {en: ''}}`) - this is IMHO wrong as it's not possible to actually POST that to Graviton. Or is the test assertion wrong?


narcoticfresh:
@hairmare I added the 'complete' test now.. so the wanted test in the sprint - the one for "free form object" does actually exist now ;-) 

IMHO we can merge this as soon CI is through..

PR #201, opened 2015-07-20
Make module translatable

hairmare:
Also fixes some bits and pieces that where in the way of having multiple translatable fields in a single object and some generator stuff.

We'll need to do some serious refactoring on the translatable stuff before it becomes real useable and not as smelly as it is right now, but for the time being this has to make do.


narcoticfresh:
@hairmare Please create some 'tech debts' user stories if you know what is still wrong with translations.. At least we need to know if it actually works (i think it does?) or if it's just smelly as in the code s*cks? ;-)

PR #203, opened 2015-07-27
add file mtime

narcoticfresh:
add and test for file mtime

### DIFF
--- a/src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
+++ b/src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
@@ -191,19 +191,19 @@ class FileControllerTest extends RestTestCase
 
         $expectedErrors = [];
         $expectedError = new \stdClass();
-        $expectedError->property_path = "data.metadata.size";
+        $expectedError->propertyPath = "data.metadata.size";
         $expectedError->message = "The value \"data.metadata.size\" is read only.";
         $expectedErrors[] = $expectedError;
         $expectedError = new \stdClass();
-        $expectedError->property_path = "data.metadata.mime";
+        $expectedError->propertyPath = "data.metadata.mime";
         $expectedError->message = "The value \"data.metadata.mime\" is read only.";
         $expectedErrors[] = $expectedError;
         $expectedError = new \stdClass();
-        $expectedError->property_path = "data.metadata.createDate";
+        $expectedError->propertyPath = "data.metadata.createDate";
         $expectedError->message = "The value \"data.metadata.createDate\" is read only.";
         $expectedErrors[] = $expectedError;
         $expectedError = new \stdClass();
-        $expectedError->property_path = "data.metadata.modificationDate";
+        $expectedError->propertyPath = "data.metadata.modificationDate";
         $expectedError->message = "The value \"data.metadata.modificationDate\" is read only.";
         $expectedErrors[] = $expectedError;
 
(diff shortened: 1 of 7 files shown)


src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
```diff
@@ -83,10 +83,33 @@ public function testPostAndUpdateFile()
         $response = $client->getResponse();
         $this->assertEquals(201, $response->getStatusCode());
 
+        // ehm - how to force that mtime will differ?
+        sleep(1);
```
hairmare: Whaaat :( Maybe just don't use ``assertGreaterThan`` on line 108 instead of wasting this second. It's not like our test suite is blazingly fast right now and we have that second to spare.

It might be interesting to assert that the fields contain something formatted as an actual date instead.

narcoticfresh: @hairmare I *knew* you wouldn't like that :grin: 
Well format is fine - but how to check if it actually *changes* on PUT?

src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
```diff
@@ -179,6 +204,9 @@ public function testPostAndUpdateFile()
         $expectedErrors[2] = new \stdClass();
         $expectedErrors[2]->property_path = "data.metadata.createDate";
         $expectedErrors[2]->message = "The value \"data.metadata.createDate\" is read only.";
+        $expectedErrors[3] = new \stdClass();
+        $expectedErrors[3]->property_path = "data.metadata.modificationDate";
+        $expectedErrors[3]->message = "The value \"data.metadata.modificationDate\" is read only.";
```
hairmare: Also these are a wreck and their getting worse with every added example...

Y U no write as follows or similar?

```php
$error = new \stdClass;
$error->....
$expected[] = $error;
```

These long chains of similar stuff with only a manually increased counter are extremely irritating...

src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
hairmare: Seeing we only have the precision of a second it might make more sense to just wait for the next second to start rather than to always wait for a full second...

This might also be a case of "we-should-let-e2e-catch-it"...

src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
narcoticfresh: @hairmare Well; I tried to do that..

problem is when i do it as:

```php
$expectedErrors = [];
$expectedError = new \stdClass();
$expectedError->property_path = "data.metadata.size";
$expectedError->message = "The value \"data.metadata.size\" is read only.";
$expectedErrors[] = $expectedError;
```

Then i get a CS complaint:
```
FILE: .../src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php
----------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
----------------------------------------------------------------------
 194 | ERROR | Variable "property_path" is not in valid camel caps
     |       | format
 198 | ERROR | Variable "property_path" is not in valid camel caps
     |       | format
 202 | ERROR | Variable "property_path" is not in valid camel caps
     |       | format
 206 | ERROR | Variable "property_path" is not in valid camel caps
     |       | format
```

That's why I changed this to array access; bringing in the need to sequential numbering.. I think we can live with that?

hairmare: I somehow don't get that ``$expectedError->property_path`` should differ from ``$expectedError->[0]property_path``. Seems to be an edge case in our sniffer.

Also wouldn't ``propertyPath`` be the right way anyhow?

Care to change the naming on this PR or do you want me to raise an issue?

narcoticfresh:
@hairmare OK; i changed the naming, fixed it everywhere, removed the `sleep()` and so on..

hairmare:
@narcoticfresh Thanks, will merge pending CI

PR #204, opened 2015-07-27
ensure RQL encoding is correct in Link header

narcoticfresh:
It was suggested that the encoding of RQL parser specialties (ie the encoding of the dash ["-"] character in an encoded string) is wrongly encoded in the generated Link header.

I added a test to ensure it's correct. As it turns out, the encoding is fine. 

I need to check with the Graviphoton people for an example that gets encoded wrong..
@hairmare You know any wrongly encoded expression?

### DIFF
--- a/src/Graviton/CoreBundle/Resources/definition/Config.json
+++ b/src/Graviton/CoreBundle/Resources/definition/Config.json
@@ -21,6 +21,14 @@
                     "$ref": "http://localhost/core/app/admin"
                 },
                 "default": "/"
+            },
+            {
+                "id": "admin-additional+setting",
+                "key": "additional-setting",
+                "app": {
+                    "$ref": "http://localhost/core/app/admin"
+                },
+                "default": "stronghold"
             }
         ]
     },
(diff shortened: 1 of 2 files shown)


hairmare:
@narcoticfresh Anything with a ``+`` in it might be a good example. Also, anything with a ``like()`` operator containing a wildcard.

narcoticfresh:
@hairmare But do you mean `+` as part of an encoded string or as part of `sort(+field)`? also the wildcard glob shouldn't be encoded anyway, i think that is covered already(?)

hairmare:
I meant ``+`` as part of an encoded string... What was even the reason this was raised?

From the looks of it it seems to work.

hairmare:
Also, if it's only about ``$`` as per the original issue, my work in 8a97be96731575f56267669b3d0f9b8ecb41de87 might help (or I'm not grasping what was the issue in the issue).

narcoticfresh:
@hairmare I added `+` and a glob..
I'm struggling to find bad behavior from the Link header generation.. This all comes from GRV-2031 and has been seen at the sprint demo..

As I see it; the issue is that the Xiag parser example of 'how to escape' uses `rawurlencode()` while are Link header is somewhere (still don't know where exactly) done with `urlencode()` - so we've seen different encoding and were confused it seems.

But AFAIK this shouldn't make a difference; my test should make that sure that everything looks as expected..

PR #205, opened 2015-07-28
add deep fields in showcase test

narcoticfresh:
this adds deeply nested fields to the ShowCase test

@hairmare This one actually highlights a week spot of the Graviton behavior at the moment. As you can see, I needed to add empty fields to `showcase-minimal.json`

It seems that Graviton
* Always renders out empty arrays, even if they are subproperties (n levels) of actually empty properties that nobody set in the first place (see `deepNestedArray`) 
* Somehow renders out the deep field `deepNestedProperty.thisis.one.goes.very.deep.down.thats` as array (it should be an object). Also it shouldn't render it at all cause nobody sets it.

Both fields mentioned in the bullet points above get rendered by Graviton in after the POST creation *regardless if they have been passed in by the client* - they just get rendered no matter what. 

This goes contrary to the concept of "not rendering null values" somehow as we're rendering out stuff that isn't set. Should we address this or leave it as-is?


hairmare:
This was probably also the reason I added ``contactCode.text`` recently. Maybe we should take care of the root issue here.

@narcoticfresh I saw some nested recursion errors in form generation when I first switched to the new test-bundle. Did you experience anything similar and change anything to get it to work?

hairmare:
@narcoticfresh I just rechecked without any of my local changes and this is really leading to nested recursion exceptions.

The problem is in ``ShowcaseControllerTest::testMissingFields``. See [this pastebin](http://pastebin.com/8p6VJYw3) for the full output.

I'm pretty sure that this stems from me being on OSX and thus not having a case sensitive file system (ie. it's probably a case of ``ShowCase`` vs. ``Showcase``).

I recently renamed the ``Showcase.json`` to ``ShowCase.json`` to catch exactly this kind of error...

Could you check to see if you can fix this?

narcoticfresh:
@hairmare so to be clear.. your errors disappear when the file is named `Showcase.json`?

hairmare:
@narcoticfresh Probably. I can't check right now without backing out of a largish change...

hairmare:
@narcoticfresh This ended up having nothing to do with upper vs. lowercase filenames :point_right:  68cdded

hairmare:
however, something else is now broken :(

narcoticfresh:
@hairmare it seems the newly rql `select()` introduces a nasty side effect: each time the `ShowCase.json` structure/fixtures change (as it does with this PR), the test will fail as it duplicates the fixtures at https://github.com/libgraviton/graviton/blob/feature/add-more-showcase-testing/src/Graviton/CoreBundle/Tests/resources/showcase-rql-select-initial.json

so the test failure reflects that fact.

narcoticfresh:
@hairmare slooow travis.. but it's getting greener..

hairmare:
I'll merge when CI is through, I don't expect much from scrutinizer...

PR #206, opened 2015-07-28
Update parser and parser bundle

hairmare:
The updates to ``graviton/php-rql-parser`` and ``graviton/rql-parser-bundle`` remove quite some cruft and make integrating with the parser much easier.

This PR changes the bare minimum to get everything up and running with the latest parser version. I'm expecting to do some more cleaning over the course of implementing the changes I'm actually after.

Feature wise this does the following things.

* [x] use ``RequestListener`` from parser bundle to do parsing
* [x] apply query to visitor in ``DocumentModel``
* [x] update various deps that made sense to update anyhow
* [x] fix ``./app/console debug:event-dispatcher`` my removing ``scope="request"`` from ``graviton.rest.response``
* [x] make rql in ``Link`` headers less encoded

It also fixes the issue where ``$`` was already being decoded php all of PHP's magic methods. I originally had that fix in graviton, but It became part of the silly ``$_SERVER`` parsing in [the listener](https://github.com/libgraviton/GravitonRqlParserBundle/blob/develop/src/Listener/RequestListener.php#L60).

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/ConfigControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/ConfigControllerTest.php
@@ -46,9 +46,10 @@ class ConfigControllerTest extends RestTestCase
      */
     public function testLinkHeaderEncodingDash($expression, $resultCount)
     {
-        $expression = urlencode($expression);
         $client = static::createRestClient();
+        $_SERVER['QUERY_STRING'] = 'q=' . $expression;
         $client->request('GET', '/core/config?q='.$expression);
+        unset($_SERVER['QUERY_STRING']);
         $response = $client->getResponse();
 
         $this->assertContains($expression, $response->headers->get('Link'));
--- a/src/Graviton/CoreBundle/Tests/Controller/ConfigControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/ConfigControllerTest.php
@@ -65,15 +65,15 @@ class ConfigControllerTest extends RestTestCase
     {
         return array(
             array(
-                'eq(id,'.$this->encodeString('tablet-hello-message').')',
+                'eq(id'.$this->encodeString(',tablet-hello-message').')',
                 1
             ),
             array(
-                'eq(id,'.$this->encodeString('admin-additional+setting').')',
+                'eq(id'.$this->encodeString(',admin-additional+setting').')',
                 1
             ),
             array(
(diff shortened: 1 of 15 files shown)


src/Graviton/CoreBundle/Tests/Controller/ConfigControllerTest.php
```diff
@@ -90,7 +90,7 @@ private function encodeString($value)
         return str_replace(
             array('-', '_', '.', '~'),
             array('%2D', '%5F', '%2E', '%7E'),
-            $value
+            rawurlencode($value)
```
narcoticfresh: @hairmare if you `rawurlencode()` here - don't you need to remove the already present `urlencode()` in the test function? but then it won't match anymore I assume.. or what is the reason to `rawurlencode()` here?

hairmare: I need to check, there sure is something smelly here...

hairmare: @narcoticfresh It seems to be needed as a workaround due to the fact that our test runner does not override ``$_SERVER['QUERY_STRING']`` on requests. I'll check what I can do.

hairmare:
@narcoticfresh I'm through here, let me know if you see anything else.

src/Graviton/CoreBundle/Tests/Controller/ConfigControllerTest.php
mrix: I try to clarify encoding rules in RQL
As written in [parser docs](https://github.com/xiag-ag/rql-parser#encoding-rules) all non `[a-z0-9]` characters must be encoded with percent encoding.
In basic RQL string values must be encoded according to RFC 3986. RFC 3986 says "-" (minus) character should be not encoded.
I added new restriction because "-" character is keyword used in "select" statement.
So condition `abc = "a-b c"` looks in RQL as `eq(abc,a%2Db%20c)`.
Then we pass RQL expression to query string parameter:
* `'http://localhost?'.http_build_query(['q' => 'eq(abc,a%2Db%20c)'])`
* `'http://localhost?q='.rawurlencode('eq(abc,a%2Db%20c)')`
* `'http://localhost?q='.urlencode('eq(abc,a%2Db%20c)')`

When you give RQL from request `$requert->query->get('q')` you receive correct RQL `eq(abc,a%2Db%20c)`.

PR #208, opened 2015-07-30
GRV-2059 implemented PATCH request

smoskalenko:
Implemented PATCH request

* [x] fix composer.lock
* [x] write ``Showcase`` tests
* [x] use form component instead of directly serializing


narcoticfresh:
@smoskalenko can you first resolve the conflicts github complains about?

Then, please add extensive testing in the `Showcase` Controller test (https://github.com/libgraviton/graviton/blob/develop/src/Graviton/CoreBundle/Tests/Controller/ShowcaseControllerTest.php)

I'd like to see the tests demonstrate
* PATCHing $refs
* PATCHing Translatabes -> do they make it into the `Translatable` collection?
* PATCHIng "free form objects" (objects that are not really defined in the JSON format/doctrine)
* PATCHing subsubsub levels (deep nested)
* and so on..

Basically, you find all that we support in that Showcase Test (you'll find the json to it in `vendor/graviton/test-services-bundle/Resources/definition/ShowCase.json`) assuming you have `--dev` installed with Composer

composer.json
```diff
@@ -49,7 +49,8 @@
         "aws/aws-sdk-php": "~2.7",
         "eo/airbrake-bundle": "~1.0",
         "hwi/oauth-bundle": "~0.3",
-        "graviton/deploy-scripts": ">=0.2.2"
+        "graviton/deploy-scripts": ">=0.2.2",
+        "mikemccabe/json-patch-php": "dev-master"
```
hairmare: Please switch to a versioned release here. We prefer using tagged releases rather that just using master.

If there are feature you need that are only in master that's fine, but we need to be informed.

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -423,6 +425,45 @@ public function putAction($id, Request $request)
...
+            $patchedDocument = JsonPatch::patch(
+                json_decode($serialized, 1),
+                json_decode($request->getContent(), 1)
+            );
+        } catch (JsonPatchException $e) {
+            throw new BadRequestHttpException('Invalid PATCH request.');
+        }
+
+        // Update object
+        $updatedRecord = $this->deserialize(json_encode($patchedDocument), $this->getModel()->getEntityClass());
```
hairmare: This should work through ``getForm``/``checkForm`` so it can do all the validation that is triggered by using forms.

composer.lock
```diff
@@ -5119,6 +5146,7 @@
         "php-jsonpointer/php-jsonpointer": 5,
         "graviton/php-rql-parser": 15,
         "knplabs/knp-gaufrette-bundle": 20,
+        "mikemccabe/json-patch-php": 20,
```
hairmare: The ``composer.json`` change should also be added to the lockfile.

smoskalenko: composer.lock file generates when I run "composer update" command, and because of not all packages in composer.json file have exact versions, I get a lot of changes in lock file and also we will get always merge conflicts. Should I regenerate lock file always when change composer.json file? Or probably you will do it before new releases, which way suits you?

narcoticfresh:
@smoskalenko 
composer.lock updates must be done in the pull request we're they needed. if another pull request updated composer.lock before you (that will lead to conflict), you resolve that by merge `develop` again (take theirs) and re-run composer update.

also in a normal pull request like this, you should *not* just run `composer update`! you should add your new require and then run `composer update mikemccabe/json-patch-php`, only updating this single dependency! not all of them!

as I can see, the current changes in the `composer.lock` look OK; it only adds your dependency..

hairmare:
@narcoticfresh The changes in compose.lock are not OK as they currently stand.

@smoskalenko After you changed to an exact version of ``json-patch-php`` in 2a75fbfbc54f6172d1543c121eccaff09e70432e you need to rerun ``composer require mikemccabe/json-patch-php`` or something similar.

Right now the composer.lock is still referencing the dev-master version of the library even though it should reference the same version as in the lock file.

Right now if I clone your branch and run ``composer install`` I get the following message:

```
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.
```
This is not ok and needs fixing in any case.

smoskalenko:
@hairmare Yep I know that lock file should be regenerated, I planed to commit changes to lockfile with one commit together with more test cases and other fixes. Or do you prefer to have separate commit for this change?

hairmare:
@smoskalenko Having the lockfile changes in one commit is perfectly ok. It's just that we usually make sure that those changes are in the same commit as the composer.json changes are.

As long as the changes end up in you final pull request that is ok.

smoskalenko:
Hi Lucas and Dario,

I’m working on patching $ref attributes and we have a lot of discussion with Alexander about possible implementation. We see the following implementation here:

   1.    to patch $ref attribute we will use standard JSON patch command like this (just simple example): [ ‘op’ => ‘replace’, ‘path’ => ‘/nestedApps/0/$ref’, ‘value’ => ‘http://localhost/core/app/admin' ]
   2.    most libraries for JSON patching working with php arrays, so we need to convert our entities to php arrays to apply patch
   3.    in RestController we have serialize($result) method, but this method return us $ref attributes in the following doctrine format: [ '$ref' => '{"$ref":"App","$id":"admin"}’ ], but to apply patch we need URL format: [ ‘$ref’ => ‘http://localhost/core/app/admin' ]
   4.    the final conversion of $ref attributes to URL format occurs in ExtReferenceListener::onKernelResponseListener() where response contents changes

So in current logic we should somehow change content to URL format after serialization of entity. But we think that it will be good idea to have this conversion logic from $ref to URL format in one place, for example in serialization process. Please let me know what do you think?

hairmare:
Hi @smoskalenko and @mrix 

Having the conversion logic of refs to URLs in one place makes lots of sense. In fact so much sense that I was already planning on doing it later today. 
Currently I'm integrating some of the new rql visiting changes and to get my current story done that would be next on my agenda. 
Taking that into account, feel free to take stab at it. I'll push early so you can see if I get started on it, but it will most likely be in the CEST afternoon.

q-src:
@smoskalenko
We have found some things which are not working correctly:
* add is behaving like replace if the field does already exist
* missing op or path parameter ends in a PHP undefined index notice
* Uses only or mostly generic "Invalid PATCH request." error message, which is not very meaningful
* replace operation with empty value removes entire field
* some fields (translatables) get removed after an invalid replace operation for another field
* successful requests should return a json response body according to jsonpatch.org
* it is possible to remove required fields
* trying to remove id removes other (translatables) field

smoskalenko:
@q-src 

I've implemeneted new changes, and here is my comments regarding each point:

> * add is behaving like replace if the field does already exist

This is correct behavior and this case is described in [RFC6902 Section 4.1] (http://tools.ietf.org/html/rfc6902#section-4.1)

> * missing op or path parameter ends in a PHP undefined index notice

I've switched code to use [php-jsonpatch/php-jsonpatch] (https://github.com/raphaelstolt/php-jsonpatch) library for applying JSON patches instead of [json-patch-php] (https://github.com/mikemccabe/json-patch-php). New library validate such cases correctly 

> * Uses only or mostly generic "Invalid PATCH request." error message, which is not very meaningful

Handling errors added with more information in response error message

> * replace operation with empty value removes entire field

As described in specification replace function works identically to remove + add function - [RFC6902 Section 4.3] (http://tools.ietf.org/html/rfc6902#section-4.3) So and when you first remove some property and then add empty value it's just ignore and do not add to result

> * some fields (translatables) get removed after an invalid replace operation for another field

Can’t reproduce, probably it was the same issue as with removing ID

> * successful requests should return a json response body according to jsonpatch.com

[jsonpatch.com] (http://jsonpatch.com) don’t tell us what should be returned in HTTP Response. But in official documentation there is some recommendations and examples:
http://tools.ietf.org/html/rfc5789#section-2.1 So as a one of solution we can return Content-Location header, any success status code can be used, in our case we return 200 OK status and Content-Location header

> * it is possible to remove required fields

If you try to remove required field this request will not pass form validation and object will not be changed. And we have test for this case.

> * trying to remove id removes other (translatables) field

New validation rule has been added - any operations with ID field not allowed.

RamonB:
@smoskalenko
To reproduce translatable field bug send
[{"op": "add", "value": "abc", "path": "/order"}]
to /core/app/admin
Answer: [{"propertyPath": "children[order]", "message": "This value is not valid."}]
But the "title" field gets removed.

q-src:
@smoskalenko 
The required field removal issue and the empty value issue seem to affect services only which are built with older definition files (e.g. app). Services built with the current definition structure are not affected. So there are currently no changes required from your side.

However, there is one issue concerning arrays:
Existing array items don't get shifted to the right when an item is being added to a given position within the item range (See [RFC6902 Section 4.1](http://tools.ietf.org/html/rfc6902#section-4.1)). In fact, the new item does not even get inserted.

smoskalenko:
Hi @q-src

Just checked adding elements to array to specific position, it works - for example if you need to add new element to array of objects to "/hans/showcase/500" you should send the following patch request 

Add object to array of objects:

PATCH to URL: /hans/showcase/500
```javascript
[
  {
    "op": "add", 
    "path": "/nestedArray/1", 
    "value": [{"name": "123"}]
  }
]
```

I've also added this case to tests.

If you need to add element to simple list, you can try this PATCH request:
```javascript
[
  {
    "op": "add", 
    "path": "/unstructuredObject/listField/1", 
    "value": "test"
  }
]
```

Let me know if it works for you

hairmare:
@q-src The *app* service was built without using the generator. We should still be to PATCH it so this needs looking into.

q-src:
Hi @smoskalenko 
Sending the request as you described works.
My previous comment was based on the following PATCH request to /hans/showcase/500 (object not within array):
```json
[
  {
    "op": "add",
    "path": "/nestedApps/1",
    "value": {"$ref": "http://localhost:8000/core/app/tablet"}
  }    
]
```
This request ended with a HTTP 200 response, but did nothing. However, now it ends with a HTTP 400 status and a "This value is not valid." message (maybe there was a bug with $ref validation which has been solved by the last commit).

But during this test, I've encountered that the behaviour / validation is not always as expected. Moreover, it seems that it is currently not always the same value format (within array or not) required:

path | "value": [{"$ref": "http://localhost:8000/core/app/not-existent-app"}] | "value": {"$ref": "http://localhost:8000/core/app/not-existent-app"} | expected behavior
---|---|---|---
/nestedApps/1| 200 OK, ref gets inserted to pos 1, other items get shifted to the right | error "This value is not valid." | 200 OK, ref gets inserted to pos 1, other items get shifted to the right
/nestedApps/- | error "not a valid ref" + error "extra fields are not allowed" | 200 OK, ref gets appended | 200 OK, ref gets appended
/nestedApps/2 | error "not a valid ref" + error "extra fields are not allowed" | 200 OK, ref gets appended | 200 OK, ref gets appended (as the array contains already 2 items)
/nestedApps/7 | 200 OK, but nothing inserted | 200 OK, but nothing inserted | 400 Bad Request, index out of range
At least one of these input formats (within array or not) should consistently produce the expected behaviour.
Please note, that inserting references to not existent resources must be possible. This enables us to import data from relational databases without caring about the resource / table order.

smoskalenko:
Hi @q-src 

Thanks for you test cases, it seems that it's a bug in library which we use for patching JSON, I've posted issue to the author - [JSON Patch Issue #6] (https://github.com/raphaelstolt/php-jsonpatch/issues/6), but it's not fixed yet. To accelerate fix I can create fork of this library, fix the issue and create pull request to original library.

hairmare:
@smoskalenko Try to fix the issue if you can. I don't think there is another library out there that is better than the one we are using and the dev seems reactive enough to warrant the effort.

smoskalenko:
Hi @q-src and @hairmare 

Issues in php-jsonpatch library have been fixed and merged to master repository, I've bumped version to 1.2.2. Also I want you to note that correct value for patching arrays with array elements should not be wrapped to array in json patch request, so the correct patch is:

```javascript
[
  {
    "op": "add",
    "path": "/nestedApps/1",
    "value": {"$ref": "http://localhost:8000/core/app/tablet"}
  }    
]
```

To manage the last test case and return 400 Bad Request:

path | "value": {"$ref": "http://localhost:8000/core/app/not-existent-app"} | Expected result
--------------------|----------------------------------------|--------------------------
/nestedApps/7 | 200 OK, but nothing inserted	 | 400 Bad Request, index out of range

We need somehow catch such situations. I've suggested to author of php-jsonpatch throw an exception in such cases (so we can catch this and return 400 Response), but he suggested to use "test" operation before patching, where FailedTestException throws. 

I think it's normal to response with 200 OK and do nothing, but if we really need to have 400 Bad request response to master patch request, I can switch library to my fork with necessary changes (throwing Exception), but I think it's not good idea to point in **composer.json** to some development forked library.

Please let me know how should we implement the last case.

hairmare:
[RFC6902](https://tools.ietf.org/html/rfc6902#section-4.1) is a bit unclear on this edge case.

> An element to add to an existing array - whereupon the supplied
      value is added to the array at the indicated location.  Any
      elements at or above the specified index are shifted one position
      to the right.  The specified index MUST NOT be greater than the
      number of elements in the array.  If the "-" character is used to
      index the end of the array (see [RFC6901]), this has the effect of
      appending the value to the array.

What irks me is the "MUST NOT" part. Looking further [RFC6901](https://tools.ietf.org/html/rfc6901#section-4) tells us.

> Any error condition for which a specific action is not defined by the
   JSON Pointer application results in termination of evaluation.

From my understanding the user should get notified about that termination at some point.

The "test" solution doesn't really seem to apply since the ordering of operations is not guaranteed by JSON-Patch.

My main problem with returning a 200 in this edge case is the fact that it can silently drop data sent by a client. This will be very hard to debug if it ever hits anyone.

I agree that pointing to a fork of the lib is not the way to go. Maybe we should check for the condition in graviton beforehand and only let the library do it's thing if everything is ok.

It might also be worth checking what JSON Pointer implementation the lib is using. It might be possible to make it catchable there.

smoskalenko:
Hi @hairmare 

> It might also be worth checking what JSON Pointer implementation the lib is using. It might be possible to make it catchable there.

php-jsonpatch library uses the same JSON Pointer implementation as graviton - https://github.com/raphaelstolt/php-jsonpointer, and there are several exceptions which can be thrown. 

JSON Pointer exceptions catches inside jsonpatch library in each operation in perform method, for-example here in Add operation: [Add:perform(), Line 44](https://github.com/raphaelstolt/php-jsonpatch/blob/master/src/Rs/Json/Patch/Operations/Add.php#L44)

> The "test" solution <del>doesn't really seem to apply since the ordering of operations is not guaranteed by JSON-Patch.</del> might be the way to go, but it should deny the full patch as far as i can tell.

In RestController after patching we:
1. first check record with form validation 
2. and if validation is passed, then we save object in database

So we update record in database only if patch && validation have been passed successfullty without Exceptions, you can see it here in [RestController::patchAction()](https://github.com/libgraviton/graviton/pull/208/files#diff-0847df0821df0f4a59cfce3f433b54e6R494)

hairmare:
@smoskalenko 

I'm not sure I fully understand. Shouldn't the issue from https://github.com/libgraviton/graviton/pull/208#issuecomment-133364897 get thrown as a ``NonexistentValueReferencedException`` from json-pointer?

smoskalenko:
@hairmare  to implement 400 Bad Request for undefined index in "path" test case, we need library php-jsonpatch throws an exception, currently NonexistentValueReferencedException exception catches in perform() method of each operation, for example here in [Add::perform()](https://github.com/raphaelstolt/php-jsonpatch/blob/master/src/Rs/Json/Patch/Operations/Add.php#L44).

So to implement this I need to fork php-jsonpatch library and change the implementation (as the author of library don't want to have an exception for such cases and suggest to use "test" operation), but as we discussed it's not good idea. So how should I fix this?

smoskalenko:
Hi @hairmare 

As an alternative solution we can implement custom validation service for JSON Patch requests (before applying patch), where we can validate each operation with JSON Pointer, so we can catch NonexistentValueReferencedException exception in patchAction and send 400 Bad Request response to client.

What do you think about this solution?

hairmare:
@smoskalenko 

We absolutely need to make sure no sent data gets ignored. Implementing the check in a custom validator isn't very lightweight, but seems the way to go.

smoskalenko:
Hi @hairmare, @q-src 

I've added custom validation befor applying patches, so for such cases:

path | "value": {"$ref": "http://localhost:8000/core/app/not-existent-app"} | Expected result
--------------------|----------------------------------------|--------------------------
/nestedApps/7 | 200 OK, but nothing inserted	 | 400 Bad Request, index out of range

We now return 400 Bad Request response.

narcoticfresh:
@smoskalenko please merge develop again and make sure scrutinizer runs through (you need to 'retry' it online maybe)..

@q-src what is the state on this? any more work needed?

hairmare:
@narcoticfresh I think retrying might only work for folks that have permissions on scrutinizer. I'll retrigger it as soon as some coverage results have been sent by travis.

smoskalenko:
@narcoticfresh I've merged develop branch. How should I correctly "retry" scrutinizer?

q-src:
@smoskalenko There is still one issue (probably off by one):
According to [RFC6902](http://tools.ietf.org/html/rfc6902#section-4.1), the specified index for adding an element to an array "MUST NOT be greater than the number of elements in the array". So the index can be equal to the number of elements. As an example, it should be possible to execute the following request when nestedApps has already two elements:
```json
[
  {
    "op": "add",
    "path": "/nestedApps/2",
    "value": {"$ref": "http://localhost:8000/core/app/tablet"}
  }    
]
```
However, this is currently not the case and the request fails with `Invalid JSON patch request: Json Pointer '/nestedApps/2' references a nonexistent value`.

hairmare:
@q-src Indexing starts a 0 so a ``/nestedApps`` with two records will have a record at position 0 and at position 1. To append to the array you would want to use ``/nestedApps/-`` instead of ``/nestedApps/2``

q-src:
@hairmare As far as I understand the RFC, adding an item to an array by specifying the next free index (in this case `2`), should be possible.
> The specified index MUST NOT be greater than the number of elements in the array.

To my mind, this means that the specified index can be equal to the number of items within the array. Or am I wrong? Of course this is just a detail and it would be the same behaviour as specifiying the index by `-`.

hairmare:
@q-src now I see it, it's "MUST NOT be **greater**" and not "MUST NOT be greater **or equal**". So yes, ``/nestedApps/2`` should add the same element that ``/nestedApps/-`` targets.

smoskalenko:
@hairmare and @q-src ok I will double check this case with library we use for patching JSON

hairmare:
So. We should probably just get this merged and live with the remaining edge case by adding some docs on it...

Or do we have any more issues here?

smoskalenko:
Hi @hairmare 

Here is the results of my investigations:
1) JSON Path library allows us to use INDEX which is equal to number of elements in array, and operation with path "/nestedApps/2" is equal to adding element to the end of array "/nestedApps/-" 
2) We use JSON Pointer library for validating JSON Patches, but it throws `NonexistentValueReferencedException` in case when INDEX is equal to number of elements as it uses zero-based index. 

I've implemented handling of such cases in our custom validator (which we use for validating JSON Patches before applying) in my last commit, new test case also has been added. 

So please check this

PR #210, opened 2015-08-03
RQL parser updates

hairmare:
Bump the parser to the next version and clean up encoding of ``(`` and ``)`` where it was part of the query.

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
@@ -103,7 +103,7 @@ class ModuleControllerTest extends RestTestCase
     public function testGetModuleWithKeyAndUseId()
     {
         $client = static::createRestClient();
-        $client->request('GET', '/core/module?q='.urlencode('eq(key,investment)'));
+        $client->request('GET', '/core/module?q=eq(key,investment)');
         $response = $client->getResponse();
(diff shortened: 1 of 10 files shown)


narcoticfresh:
@hairmare hmmm.. so where is the dispatcher set now?! can't find that...?

src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -285,7 +285,7 @@ public function testExtRefTransformation()
     {
         $client = static::createRestClient();
 
-        $client->request('GET', '/core/module?q='.urlencode('eq(key,investment)'));
```
narcoticfresh: this seems to be all new - it shouldn't be necessary to `urlencode` the whole `q` expr.. What happens to all clients that don't `urlencode` their q - the way it worked until now(?) This seems like a breaking change if suddenly the entire 'q' must be encoded..

hairmare: The chars ``,``, ``(``, ``)`` should only be encoded when part of a string and not when part of the queries semantics AFAIK.

hairmare:
@narcoticfresh The missing ``setDispatcher`` call is in https://github.com/libgraviton/GravitonRqlParserBundle/pull/23. I'll bump this when that is merged and released.

PR #211, opened 2015-08-03
searchable translations

narcoticfresh:
this PR introduces searchable translations..

It basically tries to catch all RQL exprs that land on translatables and rewrites them to or(eq(),eq()) rql nodes. For the matching values, we search our `Translatables` (this is done using `I18nUtils`). For complexity, I also need to search all untranslated strings too. As untranslated strings are inserted as sugar (like in our /core/app fixtures), the user will expect this to work too. 

I added tests to illustrate all use cases.

todo 
- [x] add test also for translated and non-translated match (`*a*`)
- [x] update composer stuff to lock at tagged stuff again (had to lock branch)
- [x] support all depths (using a `CompilerPass`)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -150,3 +150,6 @@ knp_gaufrette:
 eo_airbrake:
         host: %graviton.errbit.host%
         api_key: %graviton.errbit.api_key%
+        ignored_exceptions: 
+            - "Symfony\Component\HttpKernel\Exception\HttpException"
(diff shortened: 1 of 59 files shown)


narcoticfresh:
for my perspective it works, but in order to have it deeply-nested working i need my `CompilerPass` in #214

narcoticfresh:
@hairmare so this is ready now ;-)

PR #214, opened 2015-08-04
make extref compiler passes reusable generally

narcoticfresh:
in this PR I try to move around stuff in the extref related `CompilerPasses` in order to create a more general `AbstractDocumentCompilerPass`. Based on that, I created a `AbstractDocumentFieldCompilerPass` that has the focus of scanning fields in all defined tagged services.

In order to create a new `CompilerPass` then related to document fields, one can easily implement an abstract method in order to filter out *interesting* fields concerned to that one pass. This is shown in `TranslatableFieldsCompilerPass`


narcoticfresh:
@hairmare I'm having a hard time writing a Test for this.. maybe you could help me?

Thing is, my Test reads out these doctrine.xml files: https://github.com/libgraviton/graviton/tree/feature/generalize-compiler-passes/src/Graviton/DocumentBundle/Tests/DependencyInjection/CompilerPass/Resources/doctrine

Then, my `CompilerPass` needs to be able to find the classes defined there.. I'm having a problem mocking this, as I need named classes (that are loaded) *and* implement the `TranslatableDocumentInterface` and return some fields in the `getTranslatedFields()` method.

The easiest is really when I just implement them as stubs in the `Tests/resources` folder - but I'm not sure how I should load them.. `require()` them in test would be possible but not nice - how to autoload them in a different Namespace so they get searched in the `Tests/` folder?

Any ideas how I can mock that in that way?

narcoticfresh:
@hairmare i think that's ready now.. nice new small compilerpass with a good base and test covered..

hairmare:
@narcoticfresh I just triggered another build after some mucking and merging about and will merge it if everything pans out.

PR #217, opened 2015-08-05
added fix id for testing purposes

janfriedli:
There's another pull request on git.swisscom.com which relates to this one.

### DIFF
--- a/src/Graviton/CoreBundle/Resources/definition/Module.json
+++ b/src/Graviton/CoreBundle/Resources/definition/Module.json
@@ -6,6 +6,7 @@
     "routerBase": "/core/module/",
     "fixtures": [
       {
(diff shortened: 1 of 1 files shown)


src/Graviton/CoreBundle/Resources/definition/Module.json
```diff
@@ -193,4 +195,4 @@
       }
     ]
   }
-}
+}
```
RamonB: new line is missing ;)

hairmare: One could argue that a newline is not really needed in JSON files. There seem to be a couple of opinions out there.

All in all it looks like tools can handle both ways and due to that I've decided not to let this hold us up.

We can always standardize this if a need to arises.

PR #220, opened 2015-08-07
2091 hash type

mrix:
* Added special Doctrine type for type `hash`
* Added JMS serialize handler for type `hash`
* `ExtReferenceListener` supports `hash` encoding

But it works fine only for fields declared as `hash`, not for fields of hashes. Subhashes are not managed by Doctrine. And native PHP driver [does not support](https://jira.mongodb.org/browse/PHP-604) fetching objects as `stdClass`


narcoticfresh:
@mrix so to be clear, this only affects `hash` fields? The `hash` type is only set on what we call *free form objects*, so objects that cannot be defined in all properties in the JSON definitions.

Those `hash` fields are very rare - at least they should, as mostly all properties should be modeled out of all entities.

Do you see any possibility to extend what you did to all `Document` classes? What if we would generate a common subclass to all `Document` classes?

mrix:
@narcoticfresh yes, this only affects `hash` fields. Documents always are serialized properly.
For example, definition:
```json
{
  "id": "TmpDef",
  "service": {
    "routerBase": "/tmp/def/",
    "fixtures": [
      {},
      {"subdoc": {}}
    ]
  },
  "target": {
    "fields": [
      {
        "name": "id",
        "type": "string"
      },
      {
        "name": "subdoc.name",
        "type": "string"
      }
    ]
  }
}
```

outputs:
```json
// 20150818123421
// http://graviton.dev:8000/tmp/def

[
  {
    "id": "55d2d1d97f8b9ae1178b456c"
  },
  {
    "id": "55d2d1d97f8b9ae1178b456d",
    "subdoc": {
      
    }
  }
]
```

And with RQL `select`:
```
// 20150818123925
// http://graviton.dev:8000/tmp/def?select(subdoc)

[
  {
    
  },
  {
    "subdoc": {
      
    }
  }
]
```

PR #224, opened 2015-08-12
2037 embed is optional by default

mrix:
This make optional all embedded fields by default.
I understand this can break compatibility, but it can be helpful in some cases.

I have a little trouble. Symfony processes objects with empty fields as empty object (see `contact` field in test `testEmptyAllFields`).
Couldn't you help me to solve this problem?


narcoticfresh:
@mrix.. 
As I see it, this change indeed can have some impact regarding compatibility.. do you really need this, why exactly? I prefer not to change that if not really necessary..

We should be in a state where we should not change too much anymore if not really needed ;-)

mrix:
@narcoticfresh Yes, this is necessary for BAP. You can get more info from @wuschelchopf.
I have an idea how I can make it without breaking backwards compatibility. I'm working on it.

narcoticfresh:
@mrix once you're done with your coding, could you elaborate on what changed; if it really has no backward compat issues and so on..?

mrix:
@narcoticfresh I've finished this pull request. Please check this again.
The main changes are outlined below:
* we add `required: true` by default for all hash field
* we scan XML constraint mapping files to figure out which fields are required or not
* we pass `required` property value to [options of form field type](http://symfony.com/doc/current/reference/forms/types/form.html#required)
* we build `DocumentType` form in `PRE_SUBMIT` event handler because we need to configure fields depending on submitted data:
 * if hash field data was submitted then we set `required` option to the hash field because required field cannot be a child of optional form (https://github.com/symfony/symfony/blob/v2.7.3/src/Symfony/Component/Form/Form.php#L242)
 * if hash field data was not submitted then we dont add child fields to skip their validation

Thus for definition
```json
[{
    "name": "required",
    "type": "string",
    "required": true
},{
    "name": "optional",
    "type": "hash",
    "required": false
},{
    "name": "optional.abc",
    "type": "string",
    "required": true
},{
    "name": "optional.def",
    "type": "string",
    "required": true
}]
```
the both requests are valid:
```json
{
    "required": "a",
    "optional": {
        "abc": "abc",
        "abc": "def"
    }
}
{
    "required": "a"
}
```

And the request is not valid:
```json
{
    "required": "a",
    "optional": {
        "abc": "abc"
    }
}
```

> if it really has no backward compat issues and so on..?

I think it should be tested :)
At least I can split this into two pull requests - first PR is about `DocumentType` form. Then only the current syntax can be used:
```json
{
"relations": [
    {
        "localProperty": "optionalHash",
        "type": "embed"
    },
    {
        "localProperty": "requiredHash",
        "type": "embed"
    }
],
"fields": [
    {
        "name": "optionalHash",
        "type": "class:SubHash",
        "required": false
    },
    {
        "name": "requiredHash",
        "type": "class:SubHash",
        "required": true
    }
]}
```

And second PR is about extending hash definition like:
```json
[{
    "name": "optional",
    "type": "hash",
    "required": false
},{
    "name": "optional.abc",
    "type": "string",
    "required": true
},{
    "name": "optional.def",
    "type": "string",
    "required": true
}]
```

mrix:
I added `RequiredHashControllerTest` that contains functional tests for this changes.
Service definition for new test is added in https://github.com/libgraviton/GravitonTestServicesBundle/pull/12

narcoticfresh:
@mrix i tagged libgraviton/GravitonTestServicesBundle with `v0.8.0` - can u update that via composer as soon as it's ready to be pulled

> I think it should be tested :)

Yes indeed ;-) You're approach looks fine to me, I see it's necessary to manipulate the validations based on input.. 

In the end it comes down to 2 questions:
* Do we need to change all existing JSON definitions?
* Does anything change for the clients in the behavior how Graviton handles saving on hash structures?

At the moment, it seems both questions can be answered with `no`? Is that assumption correct?

src/Graviton/CoreBundle/Tests/Controller/ShowcaseControllerTest.php
```diff
@@ -87,35 +87,128 @@ public function testGetEmptyObject()
...
+            $client->getResults()
+        );
+    }
+
+    /**
+     * see how our empty fields are explained to us
+     *
+     * @return void
+     */
+    public function testEmptyAllFields()
```
narcoticfresh: @mrix please, refrain of 'refactoring'/'changing' existing unit tests in the same PR you changing the feature that it should test..

this makes it very difficult, i first need to re-evaluate the changes of the unit tests and then see if all is good.

if you want to refactor stuff, do it separately please. I should be able to know that the `ShowcaseTest` as *as it was* - so I know exactly that it tests what was tested before..

This is a basic thing, please refrain from that in the future!

mrix: @narcoticfresh My fault. I reverted my minor changes in `ShowcaseControllerTest`.
But I had to change `testMissingFields`. Because `contact` field is [not required](https://github.com/libgraviton/GravitonTestServicesBundle/blob/v0.8.0/Resources/definition/ShowCase.json#L243) and `contact.*` fields should not be validated. I think it's right.

narcoticfresh:
@mrix 
please re-merge develop again..

> In the end it comes down to 2 questions:
> * Do we need to change all existing JSON definitions?
> * Does anything change for the clients in the behavior how Graviton handles saving on hash structures?

> At the moment, it seems both questions can be answered with no? Is that assumption correct?

can u confirm my above question? is the assumption correct?

mrix:
> Do we need to change all existing JSON definitions?

No. But if field has class type `class:...` and has `required: false` flag then its validation rules are changed.
Now `null` is correct value for such field, even if embedded class has `required: true` fields.
Hash field declared as `hash.field` is required by default.


> Does anything change for the clients in the behavior how Graviton handles saving on hash structures?

No. But I want to deploy this branch to our instance and check this again.

narcoticfresh:
> No. But if field has class type class:... and has required: false flag then its validation rules are changed.
> Now null is correct value for such field, even if embedded class has required: true fields.
well, that's what the change is about isn't it ;-) that's a better behavior then the current one..

mrix:
@narcoticfresh I finished my changes. Could you check this again, please.

narcoticfresh:
@mrix could you please login to Scrutinizer and retry your inspection? It seems I can not trigger it to re-run.. ;-/

narcoticfresh:
@mrix when i testmerge this branch, i get `Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.` from composer.. can u pls run `composer update --lock` to update the hash?

mrix:
> can u pls run composer update --lock to update the hash?

I've merged `develop` into my branch and run `composer update --lock`

> @mrix could you please login to Scrutinizer and retry your inspection? It seems I can not trigger it to re-run.. ;-/

I have no any control buttons on Scrutinizer

PR #226, opened 2015-08-13
Reflect changes in swagger.json

hairmare:
It looks like swagger has no way to express that a query string should be used as the fully query string.

The proper way would be to document all the possible fields as param using FIQL based syntax. I don't really feel like going down that rabbit hole right now since it would make things rather more complex.

~~For now I'm also preparing a version of swagger-ui that has some code that handles rql params as the special little cookie they are.~~ Turns out that the changes I wanted to implement in swagger-ui where both non-trivial and rather smelly. For now I ended up just removing the param from swagger... This means users will have to test their RQL stuff without swagger-ui.

### DIFF
--- a/src/Graviton/SwaggerBundle/Service/Swagger.php
+++ b/src/Graviton/SwaggerBundle/Service/Swagger.php
@@ -237,28 +237,11 @@ class Swagger
             );
 
             $thisPath['parameters'][] = array(
-                'name' => 'q',
+                'name' => '',
                 'in' => 'query',
                 'description' => 'Optional RQL filter',
                 'required' => false,
                 'type' => 'string'
-            );
-            // paging params
-            $thisPath['parameters'][] = array(
-                'name' => 'page',
-                'in' => 'query',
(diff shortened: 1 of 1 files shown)


narcoticfresh:
@hairmare did you try to specify a parameter in query with `name: ""` (empty)? i guess that could work..?

narcoticfresh:
@hairmare or maybe through [path templating](http://swagger.io/specification/#pathTemplating)? If we define the path of GET as `/core/app/?{query}`, then define a parameter with that name, swagger-ui should insert that there and send it, also unencoded..

hairmare:
@narcoticfresh ``name: ""`` lead to ``?=<string>``...  I'll look into path templating.

hairmare:
@narcoticfresh I just finished looking into path templating.

Currently path params are always required. Unless https://github.com/swagger-api/swagger-spec/issues/93 gets fixed, we won't be able to address this properly.

The only way I see is to generate FIQL style params for all the fields as I described in the issue. I'd prefer to wait on the linked issue though.

PR #233, opened 2015-08-19
Cleanup schema info in swagger

hairmare:
* [x] Remove all ``OPTIONS`` paths from swagger.
* [x] Add schema information paths
* [x] Make schema valid by replacing ``extref`` with ``string``

### DIFF
--- a/src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
+++ b/src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
@@ -43,7 +43,7 @@
       "title": "Collection",
       "type": "array",
       "items": {
-        "type": "exref"
+        "type": "string"
       },
       "description": "Array of required field names"
     }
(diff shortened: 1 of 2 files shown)


src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
```diff
@@ -43,7 +43,7 @@
       "title": "Collection",
       "type": "array",
       "items": {
-        "type": "exref"
```
narcoticfresh: @hairmare hm, the `Schema` stuff always manages to confuse me. Didn't we already change all `extref` cases? Somewhere we check for `type = extref` and then set the format accordingly..(?) 

Isn't the `SchemaModel.json` the base of the `SchemaDocument` class? As said, I'm confused..

Anyway, if this was extref, shouldn't it be `format: "extref"` there?

hairmare: This describes the contents of our new ``x-collection`` hint. It isn't extref since it points to a collection endpoints and extrefs need an id. It also isn't ``format: "uri"`` since ``"*"`` is a legal value.

narcoticfresh: Ah, that's the `x-collection`.. ic..

PR #234, opened 2015-08-19
implement new service definition field

samsamann:
We want more control over modifiactions of records. Not all records should be modifiable for clients.

So this PR serves a new service definition field named recordOriginModifiable. With this field, its possible to control the read/write process of records. So we can predefine some values for the recordOrigin field, which are not allowed to save, update or delete.


samsamann:
First of all this PR needs a new release of the GravitonTestServicesBundle with this changes:
https://github.com/libgraviton/GravitonTestServicesBundle/pull/10

hairmare:
Your wish has been granted... bundle tagged as v0.6.0...

app/config/parameters.yml.dist
```diff
@@ -8,6 +8,8 @@ parameters:
     graviton.security.services.whitelist: {}
     graviton.rest.special_mimetypes: {}
 
+    graviton.not_modifiable.origin.records: ['core']
```
hairmare: Could you move this to ``src/Graviton/RestBundle/Resources/config/services.xml``.

While it should stay a param I don't think it should really be exposed to the user during composer install.

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -252,4 +259,28 @@ protected function doRqlQuery($queryBuilder, Query $query)
...
+     * Checks the recordOrigin attribute of a record and will throw an exception if value is not allowed
+     *
+     * @param Object $record record
+     *
+     * @return void
+     */
+    protected function checkIfOriginRecord($record)
+    {
+        if ($record instanceof RecordOriginInterface
+            && $this->container->hasParameter('graviton.not_modifiable.origin.records')
```
hairmare: You should inject this through the constructor instead of accessing the container directly.

src/Graviton/GeneratorBundle/Resources/skeleton/model/schema.json.twig
```diff
@@ -36,6 +36,12 @@
     "id": {
       "title": "ID",
       "description": "Unique identifier"
+{% if recordOriginModifiable is not null %}
+    },
+    "recordOrigin": {
+      "title": "record origin",
+      "description": "A small string like 'core' to determine the record origin"
```
hairmare: Could you describe that ``core`` is reserved for readonly purposes and that all the others are writable.

src/Graviton/SchemaBundle/Model/SchemaModel.php
```diff
@@ -25,7 +25,7 @@ class SchemaModel implements ContainerAwareInterface
     /**
      * @var ContainerInterface
      */
-    private $container;
+    protected $container;
```
hairmare: Please revert this. ``$container`` was private for a reason ;)

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
```diff
@@ -364,4 +364,18 @@ public function getRoles()
...
+     *
+     * @return bool
+     */
+    public function isRecordOriginModifiable()
+    {
+        if ($this->def->getService() !== null && is_object($this->def->getService())) {
+            return $this->def->getService()->getRecordOriginModifiable();
+        }
+
+        return null;
```
narcoticfresh: a function named `isRecordOriginModifiable` should always return `false` or `true`, but not `null` ;-)

src/Graviton/GeneratorBundle/Resources/skeleton/document/Document.mongodb.xml.twig
```diff
@@ -42,5 +42,10 @@
                 {% endif %}
             {% endif %}
         {% endfor %}
+
+        {% if recordOriginModifiable is not null %}
```
narcoticfresh: so this should be a nice `true`/`false` check here..

src/Graviton/GeneratorBundle/Resources/skeleton/document/Document.php.twig
```diff
@@ -84,6 +91,48 @@ class {{ document }} implements TranslatableDocumentInterface
     return $this->deletedDate;
     }
 
+{% if recordOriginModifiable is not null %}
```
narcoticfresh: so this should be a nice true/false check here..

src/Graviton/GeneratorBundle/Resources/skeleton/serializer/Document.xml.twig
```diff
@@ -17,6 +17,11 @@
     <property name="deletedDate" serialized-name="deletedDate" exclude="true" accessor-getter="getDeleteddate">
         <type><![CDATA[DateTime]]></type>
     </property>
+{% if recordOriginModifiable is not null %}
```
narcoticfresh: so this should be a nice true/false check here..

app/config/parameters.yml.dist
samsamann: I think parameters should be in a central place. So you have a good overview. If you place all parameters in the service.xml you will search and search...

Why we could not move this to ``app/config/config.yml`` in the parameter section?

hairmare: Seeing that nobody really uses ``app/config/config.yml`` it would be rather lonely there.

I'd much rather have it in the same config file where it is actually being injected.

Changes other than adding to the array introduce bc breaks anyhow.

src/Graviton/GeneratorBundle/Resources/skeleton/serializer/Document.xml.twig
samsamann: Why? In this case, the check is not important whether the value of the param is true or false, because you can also explicit allow modifications on 'core' records.

narcoticfresh: @samsamann it's just to be clean - if the originating function returns (as it should) a proper boolean in all cases you can do a *normal* `=== true` check, that is way cleaner

hairmare:
Whoops, the fail in 9f7d3e1 was introduced elsewhere... Maybe @narcoticfresh can advise on this...

src/Graviton/GeneratorBundle/Definition/JsonDefinition.php
narcoticfresh: this means, return false by default, only if `$this->def->getService()->getRecordOriginModifiable()` equals to `true`, you return `true`

narcoticfresh:
@hairmare yea, this are still the deep fields from #205.. i'm not sure if that could work now, i'll try..

hairmare:
@narcoticfresh part of it is solved in #224. I haven't been able to grok the remainder.

narcoticfresh:
@samsamann @hairmare i think #205 should be green soon..

app/config/parameters.yml.dist
narcoticfresh: @hairmare IMHO its fine in parameters.yml.dist - as we set a sane default value, user will only see it once and that's it.. but the parameters *are* the proper place I would say..

hairmare: @narcoticfresh no it isn't If it's in parameters.yml-dist everyone that does an install will get asked if he wants to change it. We don't want to ask everyone, since it's not really expected that this is different on each deploy.

The dist thing is way to bloated already... It's meant for stuff we actually need to change on a per deploy base.

narcoticfresh:
@samsamann so now, merge develop again, re-apply your composer changes and your tests should run through..

src/Graviton/GeneratorBundle/Resources/skeleton/model/schema.json.twig
samsamann: @hairmare that's not quite true. you can explicit allow modification with ``recordOriginModifiable: true``

src/Graviton/GeneratorBundle/Resources/skeleton/document/RecordOriginInterfaceDefinition.php.twig
```diff
@@ -0,0 +1,3 @@
+{% if isrecordOriginFlagSet %}
+, \Graviton\RestBundle\Model\RecordOriginInterface
+{% endif %}
```
hairmare: It might be a bit overkill to have this in its own file. It doesn't look very reusable :P

hairmare:
@samsamann Could address https://github.com/libgraviton/graviton/pull/234#discussion_r37710114? There also seem to be some scrutinizer issues that look legit.

PR #237, opened 2015-08-20
1913 proper exposeAs

mrix:
This makes possible to use `exposeAs` not only for `$ref` field.
Also compiler passes in `DocumentBundle` are refactored.

PR #245, opened 2015-08-25
1913 update module schema

mrix:
This pull request depends on #237 because `exposeAs` is partly supported for non-`extref` fields. #237 fixes `exposeAs` usage.
Please check #237 first.


src/Graviton/CoreBundle/Resources/definition/Module.json
```diff
@@ -24,11 +24,8 @@
             "description" : {
               "en": "..."
             },
-            "gui": {
-              "$ref": "/core/product/1"
-            },
             "service": {
-              "$ref": "/core/product/2"
+              "$ref": "/core/product/1"
```
hairmare: From my understanding this should be ``/core/product/`` or any other path to a collection (ie. ``/file/`` or ``/core/app/``) and not contain an id part anymore.

hairmare:
Thanks for the change and bump, will merge when ci is through...

PR #247, opened 2015-08-25
Feature/rabbitmq

lapistano:
created on behalf of @q-src


hairmare:
This is very far away from being a reviewable PR, we need to find out why unit-tests are segfaulting first.

@lapistano Could we stop this silly practice of spamming everyone with heaps of messages in HipChat just because of not wanting to use https://github.com/libgraviton/graviton/branches please.

lapistano:
@hairmare last time I checked, we had the agreement, that if we want something to be reviewed there has to be a PR .. as I was asked to review this story, I made a PR.. 
And just to be clear, I was never asked, nor informed that this is not the practice anymore.

hairmare:
@lapistano I've no idea why this would be considered ready to review. phpunit hasn't run through once without segfaulting so far. It isn't even up to date enough to be mergeable without changes...

PR #250, opened 2015-08-28
Fix segfault in HHVM.

mrix:
Skipping test that is broken in HHVM.

### DIFF
--- a/src/Graviton/GeneratorBundle/Tests/Manipulator/File/XmlManipulatorTest.php
+++ b/src/Graviton/GeneratorBundle/Tests/Manipulator/File/XmlManipulatorTest.php
@@ -52,6 +52,10 @@ class XmlManipulatorTest extends \PHPUnit_Framework_TestCase
      */
     public function testSaveDocument()
     {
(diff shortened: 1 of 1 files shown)


hairmare:
Why would we want to skip this? All skipping it accomplishes is that we ignore that our code fails on hhvm. If hhvm ever fixes such segfaults we won't notice that due to the tests still failing. 

If anything the segfault should get reported to hhvm directly. OTOH we don't really support hhvm and it was added to the build matrix out of pure curiosity.

mrix:
@hairmare it is bug in HHVM. Other developers are doing [the same](https://github.com/symfony/symfony/search?q=defined+HHVM_VERSION).
I've already created an issue in HHVM tracker

mrix:
See https://github.com/facebook/hhvm/issues/6168

narcoticfresh:
@mrix 
I agree with @hairmare. HHVM is not critical for us, we just let it fail until - maybe some day - it will work as expected. In the issue you reference there is already a proposed fixing PR..

PR #251, opened 2015-08-31
Add Queue-like functionality through RabbitMQ for async workers

q-src:
**current state**
* we have 2 services (`/event/worker` and `/event/status`)
* basic logic is complete. If there are workers subscribed on a certain event, an `/event/status` entry is persisted. the url to that resource is included in the Queue output.
* the structure of the queue object generated is fixed through the `QueueEvent` document (also see below)
* all necessary flow logic is implemented
* many parameters have been removed to simplify. now, every change leads to a broadcast to the queue.
* now, JSON is written to the queue. We write a serialized instance of `QueueEvent` to the queue. this looks like this: 
```
{"className":"Graviton\\CoreBundle\\Document\\App","recordId":"admin","event":"update","publicUrl":"http:\/\/localhost:8000\/core\/app\/admin","statusUrl":"http:\/\/localhost:8000\/event\/status\/8aa71f31044f3beebf2ea3fb123f4409","routingKey":"document.core.app.update"}
```
if there is an update on an object without a `Controller`, the field `publicUrl` will be `null`. If there is no worker subscribed on the event, no `EventStatus` will be created and thus `statusUrl` in the `QueueEvent` will be `null`.
* `routingKey` is now always a 4 component string, i.e. `document.core.app.create`. so it's `document` (fixed prefix for doctrine stuff), then `[bundle].[document].[event]`
* workers can subscribe with wildcards, but always 4 component subscription. thus, to receive all, one must subscribe on `document.*.*.*` or `*.*.*.*` - all to core app would be `document.core.app.*`.

*todo*
- [x] write unit tests for the publisher and the basic flow
- [x] add tons of documentation on this feature (both in a README.md in the bundle and also on https://gravity-platform-docs.nova.scapp.io/api/)

*old todo*
- [x] The Bundle is renamed to GravitonRabbitMqBundle, as it is inevitably RabbitMQ specific
- [x] All config parameters "graviton.message.\*" are renamed to "graviton.rabbitmq.\*"
- [x] DocumentEventPublisher is final and all it's protected members are changed to private
- [x] DocumentEventPublisher uses constructor arguments for setting $additionalProperties (optional, default value array()) and $documents (which must be set to private)
- [x] MessageProducer uses constructor arguments for setting $registeredRoutingKeys (which must be set to protected)
- [x] MessageProducer's $replyTo is removed, as it is not necessary
- [x] The JobStatus document is renamed to WorkerStatus


narcoticfresh:
@q-src i'm a little confused looking at the changes.. are there no services (facing the consumer, usable from the client) defined already? AFAIK we need some services to this?

q-src:
@narcoticfresh At the moment, there is only the dump consumer, which can be started by `app/console rabbitmq:consumer dump`. It does nothing but print the messages it receives.
Other, more specific services / consumers are not included within this pull request.

narcoticfresh:
@q-src i think we should start creating 2 services: 
* `/event/worker` - where the worker can register themselves..
* `/event/queue` - where we keep track of the worker status (the "ticket") - or this could also be called `/event/ticket/`.. what u think?

i still have to see how everything works together in the end..

src/Graviton/RabbitMqBundle/Service/MessageProducer.php
```diff
@@ -0,0 +1,66 @@
...
+     * @param string $msgBody              The message to be sent to the queueing server.
+     * @param string $routingKey           The worker channel identifier to be used (e.g. core.app.create)
+     * @param array  $additionalProperties All additional properties
+     * @throws UnknownRoutingKeyException When the given routing key is not registered
+     *
+     * @return void
+     */
+    public function publish($msgBody, $routingKey = '', $additionalProperties = array())
+    {
+        $additionalProperties['reply_to'] = isset($additionalProperties['reply_to']) ?
```
narcoticfresh: wieso `reply_to`? use `camelCase` pls..

q-src: This is a restriction of php-amqplib. See https://github.com/videlalvaro/php-amqplib/blob/master/doc/AMQPMessage.md

narcoticfresh:
@lapistano 
i need some ideas about testing this thing now.. if you look at `DocumentEventPublisher` - how to test that thing?

I was originally thinking on testing it like a `CompilerPass`, mocking everything.. but if you see how much constructor dependencies this thing has, this is quite complicated it seems to me. Mocking the `Router` to give valid routes, the `RequestStack` and so on..

Any way to write a test that has the complete `Application` initialized again with the `DoctrineListener` in place so we could just use the `Client`?

Can you help me with a good approach to testing here? Or maybe author an initial set to see where we can go?

narcoticfresh:
@lapistano bump.. any ideas?

src/Graviton/RabbitMqBundle/GravitonRabbitMqBundle.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Graviton\RabbitMqBundle;
+
+use Graviton\BundleBundle\GravitonBundleInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+/**
+ * GravitonCoreBundle
+ *
+ * WARNING: Don't change me without changing Graviton\GeneratorBundle\Manipulator\BundleBundleManipulator
```
hairmare: puhleeze... copypasta much

narcoticfresh:
@lapistano 
ok, the rewrite to a `RequestListener` is done - can u please help now to test `EventStatusLinkResponseListener`?

the rewrite results in a much smaller codebase then before. the listener was already there anyway to inject the statusUrl into the `Link` header

Why is it better in a `RequestListener` one may ask. Well, I had to find out the hard way
* A typical Graviton request potentially produces other changes - it inserts `Translatables`, inserts in referenced subcollections and so on. The doctrine listener was flooded with stuff - and I couldn't reliably determine which one is the "main change" - as `RequestListener`, this is waaay easier to know
* If more then 1 change happens, then potentially we needed to create multiple `EventStatus` objects if workers are subscribed on the underlying documents. For example when you insert an `/core/app/hans` resource, it creates Translatables. What if somebody has subscribed on `graviton.i18n.*.*`? I also picked it up, generated another `EventStatus` - and this for all n ` Translatables`. In one instance, a total of 30 `EventStatus` resources were created. I thought about "blacklisting" some events, but then I had problems with the `EventStatus` changes, and well.. that just sucks.
* The `EventStatus` (`/event/status`) resource I created also backfired into my listener - opening a vector for an infinite loop, especially if a `worker` subscribed to `graviton.eventstatus.*.*` - very nasty
* My cloud instances crashed instantly after a single PUT - don't know why, but i think it's because of the aforementioned problems.

What advantages does it have as a `RequestListener`
* I easily know what the user requested, I have the "publicUrl" already present thanks to the selfLinkHeaderListener - saves a much code fiddeling around with the mapping from Doctrine document to URL.
* It is sure my listener only get's executed once, not impacting so much on performance as a Doctrine listener can/does.

That's just my reasoning, I felt to went of some steam..

So, @lapistano, let's test this somehow..

narcoticfresh:
@hairmare after your feedback on gravity-platform/doc#7 i've implemented the inclusion of the event names in the schema (and of course swagger.json). 

changes made: 
* new property `x-events`, an array of all events (service level) that this service fires that can be subscribed to
* to make it more uniform and reusable, event names are now generated by a `CompilerPass` (test covered already)
* `SchemaUtils` (and the `Listener`) uses that map generated by this `CompilerPass` - I added a test assertion for the new property in the `FileBundle` schema test.

I'll update the doc to reflect this change..

src/Graviton/RabbitMqBundle/Document/QueueEvent.php
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * @var string $event
+     */
+    public $event;
+
+    /**
+     * @var string $publicUrl
+     */
+    public $publicUrl;
```
hairmare: This should be exposed as ``public.$ref`` and not ``publicUrl`` thus following the conventions described in [JSON-Reference](http://json-spec.readthedocs.org/en/latest/reference.html) as originally noted in https://github.com/gravity-platform/doc/pull/7#discussion_r39739603

lapistano:
about the documentation topic:
- no README.md in Bundle
  - README should link to https://gravity-platform-docs.nova.scapp.io/api
- do we wrote down some information about the metadata.action field in file service and how to use it?
- keep in mind with https://issue.swisscom.ch/browse/EVO-766 the structure of */event/status* will change; https://github.com/gravity-platform/doc/blob/feature/event-doc/source/_pages/api/event.md#anatomy-of-an-eventstatus has to be updated accordingly

hairmare:
regarding the docs... we should do both, a README.md in the bundle as well as updated info in the docs repo.

narcoticfresh:
@lapistano i'll add a README.md that links to the doc once that branch is ready and merged.. until then, the URL is not final

> do we wrote down some information about the metadata.action field in file service and how to use it?

this would be wrong. the `metadata.action` field is a trigger for *some* workers that deal with `/file`. **any** field on any service *can* potentially trigger a worker. so imho it makes no sense to document `metadata.action` on `/file`.

how to use `metadata.action`? as any other field.. there's absolutely no difference.

hairmare:
Yupp, what @narcoticfresh says regarding the docs of the field.

Workers must be strongly advised to document what they are doing and what fields they need to work.

narcoticfresh:
@hairmare @lapistano @samsamann 
sooo.. imho this is ready for merging quite some time now - i've been dragging this around by develop merges since weeks now.. who wants to do the merge honors?

hairmare:
@narcoticfresh Merging this does seem to make sense at this stage. There are still some conflicts that need a manual merge though.

dev-cleanstart.sh
```diff
@@ -1,4 +1,4 @@
 rm -Rf app/cache app/logs app/bootstrap.php.cache
 composer install
-php app/console doctrine:mongodb:fixtures:load
+#php app/console doctrine:mongodb:fixtures:load
```
hairmare: Are we sure about this change?

src/Graviton/RabbitMqBundle/Service/DumpConsumer.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Callback executed when a message is received. Dumps the message body, delivery_info and properties.
+     *
+     * @param AMQPMessage $msg The received message.
+     *
+     * @return void
+     */
+    public function execute(AMQPMessage $msg)
+    {
+        echo str_repeat('-', 60).PHP_EOL;
```
hairmare: Can't we use a proper logger here?

narcoticfresh: This is just a debug consumer that only a dev starts when he wants to see what is sent to the queue and has no other means to do that... imho it makes no sense to rewrite this as the end output (content-wise) would be the same ;-)

narcoticfresh:
@hairmare @lapistano i think now we're ready to merge.. please get this in before 0.32.0 tagging

lapistano:
+1 for merging

PR #254, opened 2015-09-03
generated code changes for cs sniffer

narcoticfresh:
yesterday i needed to port a generated class into Graviton\ namespace and was fed up by all changes I need to make for cs sniffer.. this should make it a bit better ;-)


src/Graviton/GeneratorBundle/Resources/skeleton/document/Document.php.twig
```diff
@@ -13,6 +16,7 @@ use JMS\Serializer\DeserializationContext;
 /**
  * {{ base }}Document\{{ document }}
  *
```
samsamann: (at)package is not set

narcoticfresh:
@samsamann 
yes, i actually thought about removing '@package' everywhere - it isn't there in the code we all write and our cs sniffer doesn't require it  - so why is it in the generated in the first place? 
so i think this should not be added further..

PR #255, opened 2015-09-03
bump deps

narcoticfresh:
bump our deps.. i need this to see if something in #251 refactor could be affected so i have less bumps there..

additionally, i needed to fix something in the `Product` service (that is basically obsolete anyway IMHO); but we id would render as integer now in schema (that is actually correct), but the unit test assumed a string.. so also to conform to 'id's are always' strings and to satisfy the unit test, i changed the id to string..

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "aws/aws-sdk-php",
-            "version": "2.8.17",
+            "version": "2.8.19",
             "source": {
                 "type": "git",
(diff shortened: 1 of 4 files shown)


hairmare:
#255 hätte ja eigentlich ein :beer: verdient gehabt ;) sollten wir dann bei #1024 nachholen :P

lapistano:
nun, da gibt es noch viele Gelegenheiten : 
- 402
- 404
- 418
- ...

PR #256, opened 2015-09-03
rabbitmq-branch refactoring

narcoticfresh:
this is the refactor of #251 - thus, the pull request is against that branch..

current state
* we have 2 services (`/event/worker` and `/event/status`)
* basic logic is complete. If there are workers subscribed on a certain event, an `/event/status` entry is persisted. the url to that resource is included in the Queue output.
* the structure of the queue object generated is fixed through the `QueueEvent` document (also see below)

what this changes
* address all points mentioned in #251 (the todo list)
* all necessary flow logic is implemented
* many parameters have been removed to simplify. now, every change leads to a broadcast to the queue.
* now, JSON is written to the queue. We write a serialized instance of `QueueEvent` to the queue. this looks like this: 
```
{"className":"Graviton\\CoreBundle\\Document\\App","recordId":"admin","event":"update","publicUrl":"http:\/\/localhost:8000\/core\/app\/admin","statusUrl":"http:\/\/localhost:8000\/event\/status\/8aa71f31044f3beebf2ea3fb123f4409","routingKey":"document.core.app.update"}
```
if there is an update on an object without a `Controller`, the field `publicUrl` will be `null`. If there is no worker subscribed on the event, no `EventStatus` will be created and thus `statusUrl` in the `QueueEvent` will be `null`.
* `routingKey` is now always a 4 component string, i.e. `document.core.app.create`. so it's `document` (fixed prefix for doctrine stuff), then `[bundle].[document].[event]`
* workers can subscribe with wildcards, but always 4 component subscription. thus, to receive all, one must subscribe on `document.*.*.*` or `*.*.*.*` - all to core app would be `document.core.app.*`.

what we need in order to merge this into #251
* [x] unit tests segfault again - disable the doctrine listener in unit test context. it makes no sense that it sends all to the queue while tests are running. it also makes it veeery slower on travis..
* [x] feedback from @q-src if this PR is an improvement over #251 

we'll continue work on this once we only have 1 branch (this been merged into #251) - so everything that is missing also in #251 should not be done here but in #251 again..

after todo
this is a todo list i want to carry over to #251
* [ ] write unit tests
* [x] make sure `EventStatus` url is in `Link` header
* [ ] no worker can subscribe to `documents.eventstatus.*` as that would lead to an infinite loop!
* [ ] can we find a way to merge this to `develop` without having all devs to install `rabbitmq-server`? or is that doable that everybody installs `rabbitmq-server` that works on graviton?


narcoticfresh:
@lapistano so i'm waiting here for your idea (or maybe I'm finding something) on how to disable a specific doctrine subscriber in unit test context..

narcoticfresh:
@q-src now it's the pairing part.. can u please review the code and see if u understand (and agree with) all the changes. contact me if you have any questions. once you agree and the segfaulting is gone, we can merge this into #251

app/config/config.yml
```diff
@@ -174,17 +174,17 @@ old_sound_rabbit_mq:
             heartbeat: 0
     producers:
         job:
```
q-src: I think we should choose a different name for the producer, because the job naming isn't used anymore. How about event or document_event?

src/Graviton/RabbitMqBundle/Document/QueueEvent.php
```diff
@@ -0,0 +1,213 @@
...
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class QueueEvent implements TranslatableDocumentInterface
+{
+
+    /**
+     * @var string $className
+     */
+    public $className;
```
q-src: To my mind, public visibility is not necessary, since there are public setters and getters (same for the other properties).

narcoticfresh: @q-src yes, sadly in this case it's necessary because i want to be able to simply `json_encode()` it - i don't see any value to go through the serializer for this. normally, the serializer would call the `get*()` methods and thus the properties could be `protected`/`private`..

app/config/config.yml
narcoticfresh: @q-src actually, this is the stock `oldsound` producer now.. the default one. i didn't see the need to have an own Producer, as we don't need to do anything special.. so hence the name is also the default one.. I don't see why we should introduce a class that inherits from the default one (without additional logic) just so we can use a different name ;-)

q-src: @narcoticfresh I agree that it does not make sense to introduce an own producer. 
However, I think my previous comment is a bit unclear: I am talking about the producer's id, not about it's class name. So I am proposing to change the id `job`, which is the producer id I have chosen for https://github.com/libgraviton/graviton/pull/251 and does not fit anymore  ;-)

src/Graviton/RabbitMqBundle/Document/QueueEvent.php
```diff
@@ -0,0 +1,213 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Get routingKey
+     *
+     * @return string $routingKey
+     */
+    public function getRoutingKey()
```
q-src: In my opinion, the method name should be `getRoutingkey` (and `setRoutingkey`) so it follows the same naming as the other getters / setters.

src/Graviton/RabbitMqBundle/EventSubscriber/DocumentEventPublisher.php
```diff
@@ -0,0 +1,339 @@
...
+use Graviton\RabbitMqBundle\Document\QueueEvent;
+use Monolog\Logger;
+use OldSound\RabbitMqBundle\RabbitMq\ProducerInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * Publishes document level messages to the messaging bus and creates a new JobStatus Document.
```
q-src: The comment is no longer suitable. `JobStatus` should be `EventStatus`.

src/Graviton/RabbitMqBundle/EventSubscriber/DocumentEventPublisher.php
```diff
@@ -0,0 +1,339 @@
...
+     *
+     * @return void
+     */
+    public function postRemove(LifecycleEventArgs $args)
+    {
+        $this->publishEvent($args, 'delete');
+    }
+
+    /**
+     * Creates a new JobStatus document. Then publishes it's id with a message onto the message bus.
```
q-src: The comment is no longer suitable. `Creates a new JobStatus document.` may be completely removed as the instantiation doesn't happen within this method anymore.

narcoticfresh: changed..

q-src:
@narcoticfresh I've finished reviewing your changes and agree with all of them. So from my point of view, we can merge into feature/rabbitmq.

PR #260, opened 2015-09-08
Show actual versions of wrappers and graviton 

PanteraLX:
Aims of this pull request:

* At build time the versions of relevant packages are read with `composer show` and written to a `versions.json` file inside the cache.
* The `X-Version` link contains a summary of the contents of the version file formatted similar to a User-Agent string (ie. `0.9.0; graviton/graviton: 0.25.0; evoja/service-bundle: 0.x.y`)
* These versions are also reported in a new service (which is also exposed in swagger and has a schema)
* The version property of our swagger contains the same data as the X-Version string (unless swagger only supports semver and not strings, then the current wrappers version is to be displayed).

**todo**

* [x] write unit tests for for CoreUtils
* [x] write acceptance tests for the command
* [x] write acceptance tests for the controller
* [x] make sure ``composer check`` does not find heaps of errors
* [x] ensure this runs through on travis
* [x] look into scrutinizer and fix issues you are introducing
* [x] make sure that the above description describes what is actually being implemented
* [x] update docs (seeing this is an important feature for debugging it should prominently be described in the docs at https://gravity-platform-docs.nova.scapp.io/api/)

This is  [EVO-403](https://issue.swisscom.ch/browse/EVO-403)...


hairmare:
Please run phpunit and composer check on a regular basis... They both show issues that probably should not make their way into a PR.

src/Graviton/SwaggerBundle/Service/Swagger.php
```diff
@@ -174,11 +183,13 @@ private function getBasicStructure()
...
         $ret['swagger'] = '2.0';
+        $date = date('Y-m-d');
         $ret['info'] = array(
             // @todo this should be a real version - but should it be the version of graviton or which one?
-            'version' => '0.1',
+            'version' => $this->coreUtils->getVersionById('graviton')['version'],
             'title' => 'Graviton REST Services',
-            'description' => 'Testable API Documentation of this Graviton instance.'
+            'description' => 'Testable API Documentation of this Graviton instance.',
+            'lastUpdate' => $date
```
hairmare: We need to double-check that this doesn't lead to our swagger being inavlid...

composer.json
```diff
@@ -73,6 +73,7 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::removeSymfonyStandardFiles",
             "Graviton\\SwaggerBundle\\Composer\\ScriptHandler::generateSwaggerJson"
+
```
hairmare: remove this  line from changes

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * @param string $rootDir path to root dir
+     *
+     * @return array version numbers of packages
+     */
+    private function getPackageVersions($rootDir)
+    {
+        $versions = array();
+        array_push($versions, $this->getContextVersion($rootDir));
```
hairmare: It would be much more concise to just write this instead:

```php
$versions = [
  $this->getContextVersion($rootDir),
];
```

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -0,0 +1,101 @@
...
+    private function getContextVersion($rootDir)
+    {
+        if (strpos($rootDir, 'vendor')) {
+            //@todo shell exec needs to be refactored
+            $result = shell_exec('cd ' . escapeshellarg($rootDir) . '/../../../../  && composer show -s --no-ansi');
+        } else {
+            $result = shell_exec('composer show -s --no-ansi');
+        }
+        $lines = explode(PHP_EOL, $result);
+        $wrapper = array();
```
hairmare: It's currently possible that this methods returns an empty array. Is that intended?

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -0,0 +1,101 @@
...
+            $packageNames = shell_exec($command);
+        } else {
+            $packageNames = shell_exec('composer show -i');
+        }
+        $packages = explode(PHP_EOL, $packageNames);
+        //last index is always empty
+        array_pop($packages);
+        foreach ($packages as $package) {
+            preg_match_all('/([^\s]+)/', $package, $match);
+            if ((strpos($match[0][0], 'grv') === 0 | $match[0][0] === 'graviton') && !empty($match[0][1])) {
```
hairmare: And what if I want to add ``symfony/symfony`` to the mix?

I'm not happy with this preg_matcher voodoo at all. We should make this nice and configurable somewhere.

I'd rather have to add each and every package that needs listing to a config file than having to ever add something here.

We need to refactor this so it uses an ``app/config/version_service.yml`` file or something along those lines?

src/Graviton/CoreBundle/Controller/VersionController.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Returns all version numbers
+     *
+     * @param Request $request Current http request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response $response Response with result or error
+     */
+    public function allAction(Request $request)
```
hairmare: s/allAction/getAction/

src/Graviton/CoreBundle/GravitonCoreBundle.php
```diff
@@ -58,4 +60,18 @@ public function getBundles()
             new GravitonFileBundle(),
         );
     }
+
+    /**
+     * load compiler pass rest route loader
```
hairmare: update docblocks if you really need to copy-paste them

src/Graviton/CoreBundle/Service/CoreUtils.php
```diff
@@ -15,29 +15,71 @@
...
+     */
+    public function getVersion()
     {
-        //@todo if we're in a wrapper context, use the version of the wrapper, not graviton
-        $composerFile = !empty($composerFile) ? $composerFile : __DIR__ . '/../../../../composer.json';
+        $ver = array();
+        foreach ($this->versions as $version) {
+            $ver[$version['id']]= $version['version'];
+        }
+            return $ver;
```
hairmare: weird indenting is weird

src/Graviton/CoreBundle/Service/CoreUtils.php
```diff
@@ -15,29 +15,71 @@
...
+    }
 
-        if (file_exists($composerFile)) {
-            $composer = json_decode(file_get_contents($composerFile), true);
+    /**
+     * @param string $id add package name
+     * @return array single entry
+     */
+    public function getVersionById($id)
```
hairmare: I think you can remove this now... At least the possible "This id could not be resolved" thing needs to go.

src/Graviton/CoreBundle/Service/CoreUtils.php
```diff
@@ -15,29 +15,71 @@
...
-                    json_last_error()
-                );
+    }
 
-                throw new \RuntimeException($message);
+    /**
+     *
+     * @return array wrapper version
+     */
+    public function getWrapperVersion()
```
hairmare: Same here, proabably obsolete.

src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
```diff
@@ -46,7 +46,6 @@ public function setUp()
             'doctrine_mongodb'
         );
     }
-
```
hairmare: please remove whitespace changes in file

src/Graviton/CoreBundle/Tests/Services/CoreUtilsTest.php
```diff
@@ -14,47 +14,48 @@
  */
 class CoreUtilsTest extends \PHPUnit_Framework_TestCase
 {
+    private $versions = array(
+                array("id" => "self", "version" => "0.25.1"),
```
hairmare: weird indenting, please use 4 spaces per level and not more.

src/Graviton/RestBundle/Tests/Listener/XVersionResponseListenerTest.php
```diff
@@ -24,7 +24,7 @@ class XVersionResponseListenerTest extends \PHPUnit_Framework_TestCase
     public function testOnKernelResponse()
     {
         $response = new Response();
-        $version = '0.1.0-alpha';
+        $version = '0.25.1';
```
hairmare: Why are you changing this test? I  looks a bit smelly that this is the same version that graviton is tagged as in you branch... Please revert the above and maybe use ``disableOriginalConstructor`` instead of ``setConstructorArgs`` when creating a mock below. Doesn't CoreUtils take an array nowadays?

src/Graviton/SwaggerBundle/Command/SwaggerGenerateCommand.php
```diff
@@ -131,7 +131,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $this->filesystem->dumpFile(
             $this->rootDir.'/../web/swagger.json',
-            json_encode($this->apidoc->getSwaggerSpec())
+            json_encode($this->apidoc->getSwaggerSpec(), JSON_PRETTY_PRINT)
```
hairmare: please revert this change, it should be a pr in it's own right if we really need it, but I don't fell we do since it make an already rather large file even bigger for no apparent reason...

You can always do ``curl https://graviton.nova.scapp.io/swagger.json | python -mjson.tool`` on most platforms if you need it pretty printed.

src/Graviton/SwaggerBundle/Service/Swagger.php
```diff
@@ -174,11 +183,12 @@ private function getBasicStructure()
     {
         $ret = array();
         $ret['swagger'] = '2.0';
+        $date = date('Y-m-d');
```
hairmare: More like time of last deploy ;)

Please remove the lastUpdate stuff completely for now.

We can do it in another issue if we still want it and it was optional in the original issue for reasons.

hairmare:
I didn't do any review on the schema stuff yet. I'll look into that as soon as we have some schema stuff that's going in the right direction.

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
janfriedli: The return value is array so if nothing is found it should maybe throw an exception or at least return an empty array?

janfriedli:
Issue [EVO-403](https://issue.swisscom.ch/browse/EVO-403)

src/Graviton/CoreBundle/Service/CoreUtils.php
hairmare: From what I could tell by looking at it this really is only getting called by its tests...

@janfriedli Could you look into this and delete it or let me know where it's being called.

src/Graviton/CoreBundle/Service/CoreUtils.php
janfriedli: This one is called (here)[https://github.com/libgraviton/graviton/blob/5b95bde6d257a146fc833cf030f61d7a1a2a13a9/src/Graviton/SwaggerBundle/Service/Swagger.php#L187]

src/Graviton/CoreBundle/Service/CoreUtils.php
janfriedli: That's true I'll remove it.

janfriedli:
PR which belong to this one
* [wrapper](https://git.swisscom.ch/projects/GRV/repos/graviton-api-szkb-bap/commits?until=refs%2Fheads%2Ffeature%2Fswagger-version-numbers)
* [Doc] (https://github.com/gravity-platform/doc/pull/8)

src/Graviton/CoreBundle/Controller/VersionController.php
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class VersionController extends RestController
+{
+
+    /**
+     * @var coreUtils
```
hairmare: Should be "CoreUtils".

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -0,0 +1,185 @@
...
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class VersionCompilerPass implements CompilerPassInterface
+{
+
+    /**
+     * @var rootDir
+     */
+    private $rootDir;
```
hairmare: Should be `@var string`.

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * @var config
```
hairmare: Should be `@var array` since you put the results of `getContextVersion` in it.

src/Graviton/CoreBundle/Tests/Services/CoreUtilsTest.php
```diff
@@ -14,47 +14,38 @@
...
-     */
-    public function testGetVersion($version, $filePath = '')
-    {
-        $utils = new CoreUtils();
-        $this->assertEquals($version, $utils->getVersion($filePath));
-    }
+    private $versions = array(
+                array("id" => "self", "version" => "0.25.1"),
+                array("id" => "financing", "version" => "0.1"),
+            );
```
hairmare: Needs fixed formatting and a docblock...

PR #261, opened 2015-09-09
expose readonly fields as readonly

q-src:
This pull request adds a `readOnly` property to the item schema as [defined in JSON Schema v4](http://json-schema.org/latest/json-schema-hypermedia.html#anchor15). So the schema of all fields, which are defined as readonly within the `Resources/config/schema/$Service.json`, will be extended by `"readOnly": true`.

### DIFF
--- a/src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
+++ b/src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
@@ -19,9 +19,9 @@
       "description": "Primitive data type of the property"
     },
     "readOnly": {
-      "title": "Readonly",
+      "title": "Read-only",
       "type": "boolean",
-      "description": "Defines whether the property is readonly"
+      "description": "Defines whether the property is read-only"
     },
     "format": {
       "title": "Format",
--- a/src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
+++ b/src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
@@ -18,7 +18,7 @@
       "type": "string",
       "description": "Primitive data type of the property"
     },
-    "readonly": {
+    "readOnly": {
       "title": "Readonly",
       "type": "boolean",
       "description": "Defines whether the property is readonly"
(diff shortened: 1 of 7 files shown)


src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
```diff
@@ -18,6 +18,11 @@
       "type": "string",
       "description": "Primitive data type of the property"
     },
+    "readOnly": {
+      "title": "Readonly",
```
hairmare: It should be "Read-only" in documentation.

src/Graviton/SchemaBundle/Resources/config/schema/SchemaModel.json
```diff
@@ -18,6 +18,11 @@
...
+      "type": "boolean",
+      "description": "Defines whether the property is readonly"
```
hairmare: same here, "read-only"

hairmare:
Please add some validation to the schema tests in https://github.com/libgraviton/graviton/blob/develop/src/Graviton/FileBundle/Tests/Controller/FileControllerTest.php.

We need something in the spirit of other such tests (see CoreBundle for examples) that looks at this new attribute and ensures it is being written correctly.

q-src:
@hairmare I've added a test for the entire schema (like it's done for the app service within the CoreBundle).

hairmare:
please run ``composer check`` and fix issues

narcoticfresh:
@hairmare will this not make our swagger.json invalid again? shouldn't it be `x-readOnly`? or is `readOnly` a standard property?

hairmare:
@q-src I've got one more thing... please check if [this scrutinizer issue](https://scrutinizer-ci.com/g/libgraviton/graviton/inspections/e89913ec-e918-45f0-b1d5-350330aac85c/issues/files/src/Graviton/SchemaBundle/SchemaUtils.php?status=new&orderField=path&order=asc) is relevant. It's possible that it's only a docs issue but it doesn't look like a run off the mill ignorable.

Apart from that this looks mergeable.

PR #264, opened 2015-09-14
Generate swagger without relying on mongodb

hairmare:
This fixes us not being able to ``composer update && dev-cleanstart.sh`` in wrappers by making all the triggered commands work without access to a database.

It also fixes running ``docker-compose run --rm install`` by making a simple ``composer install`` possible without needing a running mongodb.

The change it introduces is that only the default locale ``en`` will be documented in swagger from here on out.

Schema endpoints (``/schema/*/item`` and ``/schema/*/collection``) are not affected.

 This doesn't fix the underlying issues I'm seeing in wrappers after a ``composer update`` that probably pertain to ``doctrine/*`` stuff.

### DIFF
--- a/src/Graviton/SchemaBundle/SchemaUtils.php
+++ b/src/Graviton/SchemaBundle/SchemaUtils.php
@@ -42,10 +42,16 @@ class SchemaUtils
      * @var array service mapping
      */
     private $extrefServiceMapping;
+
     /**
      * @var array [document class => [field name -> exposed name]]
      */
     private $documentFieldNames;
+
+    /**
+     * @var string
+     */
+    private $defaultLocale;
 
     /**
      * Constructor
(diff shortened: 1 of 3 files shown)


src/Graviton/SchemaBundle/SchemaUtils.php
```diff
@@ -123,12 +133,18 @@ public function getModelSchema($modelName, DocumentModel $model)
...
-                return $language->getId();
-            },
-            $this->languageRepository->findAll()
-        );
+        if ($online) {
+            $languages = array_map(
+                function (Language $language) {
+                    return $language->getId();
+                },
+                $this->languageRepository->findAll()
```
lapistano: since *$this->languageRepository* is only used to retrieve every registered language, why not providing the list of languages directly to the constructor? 
Imho will this be way cleaner, than introducing an 'offline' state.

hairmare: How would I inject that without having to introduce something like the above? I has to come from mongodb when available and I can't simply inject the results of the findAll call…

AFAIK this is what repositories are for? :confused:

hairmare: I would have to revert parts of 3eeeb7033a86ebbf5c2f0c7c5adb8f7f408f785a... Let me know if that is the kind of change you are after...

lapistano: will merge it as discussed.

PR #271, opened 2015-09-18
fix pagination, add X-Total-Count header

narcoticfresh:
The `Link` header contents regarding pagination are currently wrong
* It still checked for the GET param `page` to determine the page - this leads to correct headers, but the `prev` part is missing and `last` is rendered on the last page (as it's always assuming we are on page `1`).

I reflected the new correct behavior in `AppControllerTest` - it was not checked for the `prev` part there..

Additionally, I added a new header `X-Total-Count` returning the total number of records. Clients asked for this and it doesn't hurt to add it (the total count from mongo was already fetched to determine the number of pages). I got the name of the header from [here](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#pagination), a [google search](https://www.google.ch/search?q=X-Total-Count) indicates that this header name is used in the world alreay for that purpose ;-)

### DIFF
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -121,6 +121,8 @@ class DocumentModel extends SchemaModel implements ModelInterface
         // define offset and limit
         if (!array_key_exists('skip', $queryBuilder->getQuery()->getQuery())) {
             $queryBuilder->skip($startAt);
+        } else {
+            $startAt = (int) $queryBuilder->getQuery()->getQuery()['skip'];
         }
 
         if (!array_key_exists('limit', $queryBuilder->getQuery()->getQuery())) {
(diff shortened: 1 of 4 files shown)


src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -144,10 +146,13 @@ public function findAll(Request $request)
 
         $totalCount = $query->count();
         $numPages = (int) ceil($totalCount / $numberPerPage);
+        $page = (int) ceil($startAt / $numberPerPage) + 1;
```
lapistano: don't get you math:
Example:
totalCount = 13
numberPerPage = 3
startAt= 4
» $page = 3

but starting with the 4th record the pager is at page 2, no?

narcoticfresh: @lapistano I was confused too.. But the math is right for the most use case

The problem is that the client gives us the `skip` and `limit` value - we then need to determine the `page`.. problem is, `numberPerPage` depends on both values. 

*The rule for the math is* that both params of `limit()` (`skip` and `limit`) must be a multiplication of each other - otherwise, yes, we can not determine a page properly - unless you have an idea for a better formula..

So your example:
```
totalCount = 13
numberPerPage = 3
startAt= 4
» $page = 3
```

is wrong - because IMHO input is wrong (`startAt` is not a multiplication of `numberPerPage`) - how to define a page?

narcoticfresh:
@lapistano so to make it short: if `perPage` = 5, but `skip` = 2 - how to calculate a page?

the simple way would be to check if `skip/perPage` are a float (not an int), and then substract 1.. (?) but don't know if that always works..

src/Graviton/RestBundle/Model/DocumentModel.php
lapistano: ok, I was not aware of the constraint (skip and limit are interlinked) . 
From this point of view you are correct.

PR #273, opened 2015-09-21
service wrapper for swagger sources

samsamann:
This PR provides a service wrapper for swagger sources. This wrapper consists of two components:
- Swagger definition loader: This loader loads remote swagger json files and processed the definition. He is configurable over the service definition in the dependency injection container. At the moment one caching strategy is implemented. This strategy caches the json file for 1 day. With this definition loader we have the ability to propagate the endpoints and schemas of these third party services to the root uri(/).

- ProxyController: This new controller routes the request to the third party API. For this work he needs the definition of the endpoints, which were loaded by the Swagger definition loader.

### Todo's
- [x] write unit tests for the new caching strategy. This contains the strategy himself and the http loader.
- [x] we don't want to test the proxy controller with the petstore API but rather with our swagger.json


lapistano:
@samsamann schaut so aus, als müsste man develop nochmla mergen und die merge conflicts beseitigen. ;)

composer.json
```diff
@@ -49,7 +49,8 @@
         "aws/aws-sdk-php": "~2.7",
         "eo/airbrake-bundle": "~1.0",
         "hwi/oauth-bundle": "~0.3",
-        "graviton/deploy-scripts": ">=0.2.2"
+        "graviton/deploy-scripts": ">=0.2.2",
+        "jenssegers/proxy": "dev-master"
```
narcoticfresh: @samsamann dependencies should be on a tagged release, please change this. current release is v2.2.1, see https://packagist.org/packages/jenssegers/proxy

samsamann: The current release(v2.2.1) requires symfony/http-foundation(v2.5.3). We can not use this version, because we use symfony/symfony(v2.7.4) and this symfony package replace the symfony/http-foundation with the same version.

src/Graviton/CoreBundle/Tests/resources/showcase-complete.json
```diff
@@ -1,5 +1,5 @@
 {
-  "anotherInt": 6555488894525,
+  "anotherInt": 94525,
```
hairmare: These might have been so high for a reason. Why did you need to change them?

lapistano: 32bit vs 64bit border. When on a 32bit system the (int)6555488894525 will always be (int)2147483647 breaking the tests. If the host system must be a 64bit system there imho shall be an extra validation for that on application level (e.g. when the Kernel is initiated), stopping the initiation when requirements are not met.

hairmare: Yupp that's why I chalked this change up as something fishy going on...

narcoticfresh:
@samsamann i really don't see the need for that `CacheStrategy` stuff - why re-invent stuff?.. I changed `HttpLoader` to use the `Doctrine\Cache` component we already have in the framework..

Please update the TODO - now that we don't need to test an own cache strategy (cause we ditched it); what exactly is missing from the HttpLoader tests? It looks quite covering..(?)

narcoticfresh:
so i looked at the existing tests.. it's quite a mess honestly.. 

@samsamann @lapistano isn't it sufficient to have an extensive integration test for `ProxyController`? if this works, it tests all the components with it. agree?

the target would be to write an extensive `ProxyControllerTest` and ditch the other ones as they're no longer needed - that's the thing to do to get this done once and for all..

narcoticfresh:
@samsamann resolve conflicts pls; merge develop again.. i think this PR needs rebasing; but it will be a bit messy to do

PR #274, opened 2015-09-21
Introduced parameter to be able to configure default pagination limit  via parameter.yml

lapistano:
this Pr solves the problem iOs Team has with pagination. It increases the currently set default pagination of 10 records to 1000. 
This should not be the definitive solution, since theu are planning to do proper pageing in the near future.

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -36,3 +36,6 @@ parameters:
     # errbit config
     graviton.errbit.api_key: ~
     graviton.errbit.host: ~
+
+    # pagination defaults
+    graviton.rest.pagination.limit: 1000
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -38,4 +38,4 @@ parameters:
     graviton.errbit.host: ~
 
     # pagination defaults
-    graviton.rest.pagination.limit: 1000
+    graviton.rest.pagination.limit: 10
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -307,6 +307,6 @@ class DocumentModel extends SchemaModel implements ModelInterface
             return $this->paginationDefaultLimit;
         }
 
-        return 1000;
+        return 10;
     }
 }
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -39,10 +39,6 @@ class DocumentModel extends SchemaModel implements ModelInterface
      */
     protected $requiredFields = array();
     /**
-     * @var array
-     */
-    protected $notModifiableOriginRecords;
-    /**
      * @var DocumentRepository
      */
     private $repository;
@@ -50,6 +46,10 @@ class DocumentModel extends SchemaModel implements ModelInterface
      * @var Visitor
      */
     private $visitor;
+    /**
+     * @var array
+     */
+    protected $notModifiableOriginRecords;
     /**
      * @var  integer
      */
@@ -106,7 +106,8 @@ class DocumentModel extends SchemaModel implements ModelInterface
         $startAt = ($pageNumber - 1) * $numberPerPage;
 
         /** @var \Doctrine\ODM\MongoDB\Query\Builder $queryBuilder */
-        $queryBuilder = $this->repository->createQueryBuilder();
+        $queryBuilder = $this->repository
+            ->createQueryBuilder();
 
         // *** do we have an RQL expression, do we need to filter data?
         if ($request->attributes->get('hasRql', false)) {
@@ -161,7 +162,7 @@ class DocumentModel extends SchemaModel implements ModelInterface
     }
 
     /**
-     * @param \Graviton\I18nBundle\Document\Translatable $entity entityy to insert
+     * @param \Graviton\I18nBundle\Document\Translatable $entity entity to insert
      *
      * @return Object
      */
@@ -284,7 +285,9 @@ class DocumentModel extends SchemaModel implements ModelInterface
      */
     protected function checkIfOriginRecord($record)
     {
-        if ($record instanceof RecordOriginInterface && !$record->isRecordOriginModifiable()) {
+        if ($record instanceof RecordOriginInterface
+            && !$record->isRecordOriginModifiable()
+        ) {
             $values = $this->notModifiableOriginRecords;
             $originValue = strtolower(trim($record->getRecordOrigin()));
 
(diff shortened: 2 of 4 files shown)


src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -6,13 +6,12 @@
...
-use Graviton\SchemaBundle\Model\SchemaModel;
-use Symfony\Component\HttpFoundation\Request;
 use Doctrine\ODM\MongoDB\Query\Builder;
+use Graviton\ExceptionBundle\Exception\RecordOriginModifiedException;
 use Graviton\Rql\Visitor\MongoOdm as Visitor;
+use Graviton\SchemaBundle\Model\SchemaModel;
+use Symfony\Component\HttpFoundation\Request;
 use Xiag\Rql\Parser\Query;
-use Graviton\RestBundle\Model\RecordOriginInterface;
-use Graviton\ExceptionBundle\Exception\RecordOriginModifiedException;
```
hairmare: Could you please configure your IDE to not change unrelated stuff?

We are at the point where stuff is getting changed back and forth due to some silly autoformatter conflicts.

This feature branch is full of such changes making it way harder to review than necessary at all.

app/config/parameters.yml.dist
```diff
@@ -36,3 +36,6 @@ parameters:
     # errbit config
     graviton.errbit.api_key: ~
     graviton.errbit.host: ~
+
+    # pagination defaults
+    graviton.rest.pagination.limit: 1000
```
hairmare: Do we even have any test data in fixtures to test this high of a default limit?

hairmare:
I don't need instructions on how to use git flow. Could you replace the description of this PR with what you really want to achieve with the changes.

You might also want to add the new configuration option to composer.json in the incenteev-parameters section so we can set it as needed per deploy.

app/config/parameters.yml.dist
lapistano: we don't

src/Graviton/RestBundle/Model/DocumentModel.php
lapistano: shall I revert this changes?

hairmare: Yes, and please rebase your changes to get them out of history...

app/config/parameters.yml.dist
hairmare: Shouldn't we leave this at 10 here and override it much closer to the customers deploy?

I still think 10 is the much more sensible default...

lapistano: fine for me, as long as the limit will be 1000 for the iPad backends.

hairmare:
Please squash this so we don't have  5095c7c in the tree and also remove all the other noop auto formatting from that your IDE did (ie. removing whitespace next to the ``.`` concat operator, most other whitespace changes, ...)?

lapistano:
You had a valid point with the codeing guideline violations, but I will not change the git history just to make it look pretty.

hairmare:
It's not about coding violations (there were none). It's about having heaps if completely irrelevant changes in the history. 

If at any later stage you wants to see why ``RecordOriginInterface `` was used on https://github.com/libgraviton/graviton/pull/274/files#diff-16e06489fc38bdfb5fcba04f42544688R287, you'll have to step through at least 2 dummy commits in both the blame and the history.

I'll let it slide this time, but be aware that I will start to be more strict about this. If you feel there is need for change please adopt it into the code sniffer package and keep it isolated from function changes.

This still has heaps of noop changes, let me mark them up for you...

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -40,29 +39,33 @@ class DocumentModel extends SchemaModel implements ModelInterface
      */
     protected $requiredFields = array();
     /**
+     * @var array
+     */
+    protected $notModifiableOriginRecords;
+    /**
```
hairmare: Why was this moved? It's a noop change...

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -40,29 +39,33 @@ class DocumentModel extends SchemaModel implements ModelInterface
...
      * @var Visitor
      */
     private $visitor;
-
     /**
-     * @var array
+     * @var  integer
      */
-    protected $notModifiableOriginRecords;
+    private $paginationDefaultLimit;
```
hairmare: The move I pointed above makes this needlessly hard to review... All the whitespace between here and above is also noop.

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -99,12 +102,11 @@ public function setRepository(DocumentRepository $repository)
...
     {
         $pageNumber = $request->query->get('page', 1);
-        $numberPerPage = (int) $request->query->get('perPage', 10);
+        $numberPerPage = (int) $request->query->get('perPage', $this->getDefaultLimit());
         $startAt = ($pageNumber - 1) * $numberPerPage;
 
         /** @var \Doctrine\ODM\MongoDB\Query\Builder $queryBuilder */
-        $queryBuilder = $this->repository
-            ->createQueryBuilder();
+        $queryBuilder = $this->repository->createQueryBuilder();
```
hairmare: noop change

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -281,9 +284,7 @@ protected function doRqlQuery($queryBuilder, Query $query)
      */
     protected function checkIfOriginRecord($record)
     {
-        if ($record instanceof RecordOriginInterface
-            && !$record->isRecordOriginModifiable()
-        ) {
+        if ($record instanceof RecordOriginInterface && !$record->isRecordOriginModifiable()) {
```
hairmare: noop change

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -269,6 +271,7 @@ public function getConnectionName()
     protected function doRqlQuery($queryBuilder, Query $query)
     {
         $this->visitor->setBuilder($queryBuilder);
+
```
hairmare: noop change

lapistano: this is actually part of the coding guidelines to add a newline before the return keyword. 

> Add a blank line before return statements, unless the return is alone inside a statement-group (like an if statement);
> http://symfony.com/doc/current/contributing/code/standards.html

hairmare: Then we should add that one to composer check as well if it's so important. We've been following our own brand of style that doesn't mandate that line, hence it is not an error here.

All I'm trying to say that it gets boring real fast when each pull request contains more noop-CLOC and CLOC than relevant to the change you are after.

lapistano: will keep those changes separate from functional changes next time.

hairmare:
Merged in ada6bba88e06c2787f07f53ac66a7f52dd15d19f

PR #278, opened 2015-09-23
Use setMethods() instead of '_method' requirement

hairmare:
The _method thingy was deprecated many moons ago as per [symfony/symfony/UPGRADE-3.0.md](https://github.com/symfony/symfony/blob/master/UPGRADE-3.0.md#routing).

This should help overall performance due to much less log entries getting written... There are still heaps remaining though.


narcoticfresh:
:+1:

PR #279, opened 2015-09-24
Allow configuration of composer command path

hairmare:
We need to override this on cloudfoundry so it uses the buildpacks composer installation by setting it to `php /tmp/staged/app/php/bin/composer.phar`.

This should make graviton deploy again after #260 broke it, but I'm having a hard time testing it due to most likely unrelated errors when pushing to my own instances :/

### DIFF
--- a/src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
+++ b/src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
@@ -37,6 +37,7 @@ class VersionCompilerPass implements CompilerPassInterface
     public function process(ContainerBuilder $container)
     {
         $this->rootDir = $container->getParameter('kernel.root_dir');
+        $this->composerCmd = $container->getParameter('graviton.composer.cmd');
         $this->config = $this->getVersionConfig();
         $container->setParameter(
             'graviton.core.version.data',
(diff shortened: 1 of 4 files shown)


src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -70,7 +71,7 @@ private function getPackageVersions()
      */
     private function getContextVersion()
     {
-        $output = $this->runCommandInContext('composer show -s --no-ansi');
+        $output = $this->runCommandInContext($this->composerCmd . ' show -s --no-ansi');
```
lapistano: I'd really like to see this properly escaped for shell commands. Preferably in *runCommandInContext()*.

src/Graviton/CoreBundle/Compiler/VersionCompilerPass.php
```diff
@@ -92,7 +93,7 @@ private function getContextVersion()
      */
     private function getInstalledPackagesVersion($versions)
     {
-        $output = $this->runCommandInContext('composer show --installed');
+        $output = $this->runCommandInContext($this->composerCmd . ' show --installed');
```
lapistano: if not done in *runCommandInContext()* then here is the need of escaping as well.

PR #282, opened 2015-09-27
Dynamic key support in schemas and output

hairmare:
This make it able to express that we need to expose an array as an object while using parts of the object as key. This takes part of schema and service output, making such stuff writable still needs taking care of.

Given the following (simplified) field definition:

```js
{
  "name": "field",
  "type": "class:AcmeBundle\\Document\\Field",
  "x-dynamic-key": {
    "document-id": "GravitonDynModuleBundle:Module",
    "repository-method": "findAll",
    "ref-method": "getRef"
  }
}
```

The schema it generates should contain a property for every record id from the result set of `GravitonDynModuleBundle:Module::findAll`.

The swagger-schema shall contain an `additionalProperties` property allowing any id to be used.

It should generate a `Document` class that returns the value of `field` as an associative array based on the id of the MongoDBRef in the `getRef` entry of `AcmeBundle\Document\Field`. It is assumed that the method will return an unprocessed `extref` entry.

Most of this PR is actually only extending the json-definition format to express the above. Since this mostly generates schema info for now, the actual changes to the non-schema parts of the runtime are rather small.

I already have a heap of follow-up stories prepared on this, but need to get this done till the end of the sprint. Most of those stories will pertain to making `x-dynamic-key` fields writable over REST. It looks like some refactoring will be needed to get that up and running which is why this code does not take care of edge cases as much as it could.

Since this has heaps of related PRs to it I'll try and list some here (sorry about some of them being on our internal repos):
* https://github.com/libgraviton/json-schema/pull/4
* [grv/graviton-service-bundle-financing#54](https://git.swisscom.ch/projects/GRV/repos/graviton-service-bundle-financing/pull-requests/54/overview)
* [grv/evoja-checklist-bundle@develop](https://git.swisscom.ch/projects/GRV/repos/graviton-evojachecklistbundle/browse)

If you want to have a go at this you will find a preview on the [`feature/EVO-705-preview`](https://git.swisscom.ch/projects/GRV) branch of our internal wrapper. Again, sorry for linking to closed stuff...

Since there are some customer names in the repos I wasn't able to link to all of them. Please refer to [EVO-705](https://issue.swisscom.ch/browse/EVO-705) for links to additional pull-requests pertaining to this.

### DIFF
--- a/src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
+++ b/src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
@@ -131,6 +131,11 @@ class ExtReferenceListener
 
                 if (isset($item->$topLevel)) {
                     $item->$topLevel = $this->mapField($item->$topLevel, $subField);
+                } else {
+                    // map available things since we found nothing on $topLevel and there might be some refs deeper down
+                    foreach (get_object_vars($item) as $subLevel => $subItem) {
+                        $item->$subLevel = $this->mapField($subItem, $subField);
+                    }
                 }
             } elseif (isset($item->$field)) {
                 $item->$field = $this->convertToUrl($item->$field);
--- a/src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
+++ b/src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
@@ -133,7 +133,7 @@ class ExtReferenceListener
                     $item->$topLevel = $this->mapField($item->$topLevel, $subField);
                 } else {
                     // map available things since we found nothing on $topLevel and there might be some refs deeper down
-                    foreach (get_object_vars($item) as $subLevel => $subItem) {
+                    foreach ($item as $subLevel => $subItem) {
                         $item->$subLevel = $this->mapField($subItem, $subField);
                     }
                 }
--- a/src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
+++ b/src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
@@ -101,6 +101,7 @@ class JsonDefinitionField implements DefinitionElementInterface
             'exposedName'       => $this->getExposedName(),
             'doctrineType'      => $this->getTypeDoctrine(),
             'serializerType'    => $this->getTypeSerializer(),
+            'xDynamicKey'       => $this->getXDynamicKey(),
             'relType'           => null,
             'isClassType'       => false,
             'constraints'       => array_map(
@@ -161,6 +162,17 @@ class JsonDefinitionField implements DefinitionElementInterface
 
         // our fallback default
         return self::$serializerTypeMap[self::TYPE_STRING];
+    }
+
+    /**
+     * @return array|void
+     */
+    public function getXDynamicKey()
+    {
+        $key = $this->definition->getXDynamicKey();
+        if ($key) {
+            return $key;
+        }
     }
 
     /**
--- a/src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
+++ b/src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
@@ -165,7 +165,7 @@ class JsonDefinitionField implements DefinitionElementInterface
     }
 
     /**
-     * @return array|void
+     * @return XDynamicKey|void
      */
     public function getXDynamicKey()
     {
--- a/src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
+++ b/src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
@@ -1,6 +1,8 @@
 <?php
 namespace Graviton\GeneratorBundle\Definition;
 
+use Graviton\GeneratorBundle\Definition\Schema\XDynamicKey;
+
 /**
  * A single field as specified in the json definition
  *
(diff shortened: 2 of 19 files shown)


src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php
```diff
@@ -131,6 +131,11 @@ function ($subItem) use ($subField) {
 
                 if (isset($item->$topLevel)) {
                     $item->$topLevel = $this->mapField($item->$topLevel, $subField);
+                } else {
+                    // map available things since we found nothing on $topLevel and there might be some refs deeper down
+                    foreach (get_object_vars($item) as $subLevel => $subItem) {
```
lapistano: ```foreach($item as $subLevel => $subItem)```  works alike ;)

src/Graviton/GeneratorBundle/Definition/JsonDefinitionField.php
```diff
@@ -164,6 +165,17 @@ public function getTypeSerializer()
     }
 
     /**
+     * @return XDynamicKey|void
+     */
+    public function getXDynamicKey()
+    {
+        $key = $this->definition->getXDynamicKey();
+        if ($key) {
```
lapistano: I like more: 

```if ($key instanceof \Graviton\GeneratorBundle\Definition\Schema\XDynamicKey) {```

File: src/Graviton/GeneratorBundle/Definition/Schema/XDynamicKey.php [hunk 3]
```diff
@@ -0,0 +1,82 @@
...
+    public function getRepositoryMethod()
+    {
+        return $this->repositoryMethod;
+    }
+
+    /**
+     * @param string $repositoryMethod repository-method field
+     * @return $this
+     */
+    public function setRepositoryMethod($repositoryMethod)
```
lapistano: I there a list of available methods? What is a 'repositoryMethod'?

src/Graviton/SchemaBundle/Model/SchemaModel.php
```diff
@@ -167,6 +167,30 @@ public function getReadOnlyOfField($field)
     }
 
     /**
+     * tell us if a model what to be exposed using a key as field
+     *
+     * @param string $field field that we check for dynamic-key spec
+     *
+     * @return boolean
+     */
+    public function hasDynamicKey($field)
```
lapistano: ```hasXDynamicKey``` vs ```hasDynamicKey```

src/Graviton/GeneratorBundle/Definition/Schema/XDynamicKey.php
hairmare: Such a list may be found in [src/Graviton/GeneratorBundle/Resources/config/serializer/Definition.Schema.XDynamicKey.xml](https://github.com/libgraviton/graviton/blob/feature/EVO-705-dynamic-key-support/src/Graviton/GeneratorBundle/Resources/config/serializer/Definition.Schema.XDynamicKey.xml). These are also the same properties I'm adding in https://github.com/libgraviton/json-schema/pull/4.

src/Graviton/SchemaBundle/Model/SchemaModel.php
```diff
@@ -167,6 +167,30 @@ public function getReadOnlyOfField($field)
...
+    }
+
+    /**
+     * Get field used for setting stringy key value
+     *
+     * @param string $field field that we get dynamic-key spec from
+     *
+     * @return object
+     */
+    public function getDynamicKeySpec($field)
```
lapistano: ```getXDynamicKeySpec``` vs ```getDynamicKeySpec```

src/Graviton/GeneratorBundle/Definition/Schema/XDynamicKey.php
hairmare: Also a `repositoryMethod` is usually something like `findAll`. Basically any method that a given repository implements...

lapistano: would be nice to find a short explanation where these methods are defined and a @link annotation to the xml file. Just to let anyone else know where to look for those methods.

hairmare: Basically whatever the DocumentRepository belonging to to the document specified through `document-id` implements...

hairmare: I removed the comment about the xml since it was more or less a red herring.

src/Graviton/SchemaBundle/Model/SchemaModel.php
hairmare: But it's derived from `x-dynamic-key`. I can change this but I want to note that i think it makes more sense as it stands.

lapistano: thought it was a typo.

narcoticfresh:
@hairmare @lapistano 
this has my +1, seems ok..
personally for me, the most important criteria is "does it impact existing codebase". as this is an additional feature, i'm not hesitant to merge.. so go ahead.

the feature could be useful in other places, where we need to inject structures based on calculation. with this, we could let graviton call our 'callback' actions to construct structures (and maybe calculate stuff) in a more generalized fashion than today. that's nice..

PR #285, opened 2015-09-28
Configure COMPOSER_CMD for cf push

hairmare:
Short term fix for local graviton deploys (this was already in the cloud by way of jenkins).

Long term we should support deactivating the rewriting of `versions.yml` with `COMPOSER_CMD=false` or something.


hairmare:
Made obsolete by #286 I'll set this property properly when I remove the version from composer.json.

Which will happen soon anyway as I really don't like to remember to update it with each release.

PR #286, opened 2015-09-28
Made environment variable overpower the parameter »graviton.composer cmd« and added awareness of »COMPOSER_CMD« being "false"

lapistano:


hairmare:
Looks good, will merge when CI is through...

hairmare:
I found it... `getenv('UNDEFINED_VAR') === false` is always true so we need to check for !== 'false' in this context. Tested locally and pushed just now.

PR #288, opened 2015-09-28
Feature/corebundle cleanup

lapistano:
minor clean up, refactoring and execption handling added


src/Graviton/SwaggerBundle/Service/Swagger.php
```diff
@@ -183,11 +185,21 @@ private function getBasicStructure()
...
+        try {
+            $ret['info'] = array(
+                'version' => $this->coreUtils->getWrapperVersion()['version'],
+                'title' => 'Graviton REST Services',
+                'description' => 'Testable API Documentation of this Graviton instance.',
+            );
+        } catch (\RuntimeException $e) {
+            $ret['info'] = [
+                'version' => '',
+                'title' => 'Error occurred',
```
hairmare: Do we really need to have an error message in our generated swagger?

Since this gets generated at build time it would lead to funky looking sawgger docs even if we decide to ignore that there are no version infos..

Generally I don't feel it's smart to take Exception contents and write them to the output. This could lead to path disclosure issues that are very hard to detect and hunt down later on.

lapistano: This is correct. But I did not want to introduce the dependency to a logger. There at least is the possibility of using ErrBit, but I don't know how. Can you tell me?

hairmare: If you don't catch em they get logged... Unless they're ignored in app/config/config.yml...

PR #292, opened 2015-10-06
47612 extref handler

mrix:
We always use new `ExtReference` object for `extref` value and convert it using `SerializerHandler` and `FormDataTransformer`. This allows to remove `ExtReferenceListener` because `extref` values are converted to URL representation automatically. And `ExtReferenceConverter` usage was removed from Doctrine `ExtReferenceType`.

Also this fixes the bug with double converting `extref` (see https://github.com/libgraviton/graviton/blob/v0.31.0/src/Graviton/DocumentBundle/Listener/ExtReferenceListener.php#L135). For the definition:
```json
{
  "id": "RefTest",
  "service": {
    "routerBase": "/testcase/reftest/",
    "fixtures": [
      {
        "id": "test",
        "test": {
          "a": null,
          "b": {"$ref": "http://localhost/core/module/b"},
          "c": {"$ref": "http://localhost/core/module/c"}
        }
      }
    ]
  },
  "target": {
    "fields": [
      {
        "name": "id",
        "type": "string"
      },
      {
        "name": "test.a.ref",
        "type": "extref",
        "exposeAs": "$ref"
      },
      {
        "name": "test.b.ref",
        "type": "extref",
        "exposeAs": "$ref"
      },
      {
        "name": "test.c.ref",
        "type": "extref",
        "exposeAs": "$ref"
      }
    ]
  }
}
```

we get:
```json
{
  "id": "test",
  "test": {
    "b": {"$ref": ""},
    "c": {"$ref": ""}
  }
}
```
because `$ref` in `b` and `c` are encoded twice if `a` is null.


hairmare:
I'll have to check if this doesn't break https://github.com/libgraviton/graviton/blob/develop/src/Graviton/GeneratorBundle/Resources/skeleton/document/Document.php.twig#L168 (in combination with the internal checklist bundle).

mrix:
@hairmare fixed

PR #293, opened 2015-10-06
correctly generate _self Link header part to a single resource on postNoSlash route type

narcoticfresh:
currently, if the user issues a POST to `/file` (no slash at the end), the Self Link header (and thus the `EventStatus` documentUrl that depends on that) wrongly point to a *collection* (`?id=xxx`) instead to a single resource (`/file/<id>`).

This PR fixes that behavior


lapistano:
did this in EVO-771 as well ;), but it is for sure better to make a single PR for that .. will take care of it.

PR #296, opened 2015-10-07
Bump deps

hairmare:
Contains some CVE-2015-5723 as well as some php-7 RC3 changes.


lapistano:
CVE-2015-5723 sounds important. What did this actually fix?

hairmare:
Not much, just some bad default umasks in cache dirs. I don't think we're affected but is sounds bad-ass to use CVE numbers in issue comments.

http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html

PR #298, opened 2015-10-08
20151008 test embedding docs

mrix:
Do not merge this! This is only for example!
We found new critical bug (or specific Doctrine behavior??) in usage of documents mapped as `<document />` in target class in `<embed-one />`. We often use such definitions in our API (eg `Customer` and `Code`).
@narcoticfresh @hairmare this is very very important! At the moment we have no working backends!


narcoticfresh:
@mrix what exactly is the critical bug(?)

hairmare:
I'd like to know where this is cropping up. It's might be due to how symfony/form handles sub objects.

Does it even make sense to expose the id field in an embedded document? The stuff I've looked at recently uses references for the most part and this here might be the reason for that.

mrix:
> @mrix what exactly is the critical bug(?)

When we change any field in `customer` or `consultation`, original embedded document `Code` is removed. For example, we change conslultation status to `closed`, then previous `open` status is deleted form `Code` collection. I dont know how it worked before...

> Does it even make sense to expose the id field in an embedded document? The stuff I've looked at recently uses references for the most part and this here might be the reason for that.

Moment. I'm checking this.

narcoticfresh:
@mrix @hairmare i can't dig in deeply currently.. but this is influenced by the CASCADE setting that is generated in the doctrine xml definition.. maybe changing [this](https://github.com/libgraviton/graviton/blob/develop/src/Graviton/GeneratorBundle/Resources/skeleton/document/Document.mongodb.xml.twig#L31) will help?

hairmare:
The `Code` structure should probably always be a reference and never an embed since those mostly come from the loader. With a quick grep, I'm counting about 20 JSON-Defs that might be affected.

The cases I'm seeing are cases where we are embedded the *n* part of a *1:n* relationship. It looks like such cases need to be a reference or not contain an id.

I haven't been able to identify a quick fix based on the code in this pr though.

I'll look into the cascade settings as well, it seems like those have been there for ages and never really scrutinized.

hairmare:
It doesn't look like this is related to cascade. I still haven't got a better idea than to not expose ids on embedded documents and use references where we need them (like in the `Code` case).

hairmare:
I think this is being caused due to our `embed-*` relations pointing to regular odm documents instead of proper embedded documents.

As soon as I change the `EmbedTestEntity` to `<embedded-document>` in `src/GravitonDyn/EmbedTestEntityBundle/Resources//config/doctrine/EmbedTestEntity.mongodb.xml` I get the following errors:

```
1) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
+        'id' => 'afa957eed96732f0b02b5b2f21dff2f8'
         'data' => 'two'
     )
 )

2) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsReference
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
         'id' => 'one'
-        'data' => 'one'
     )
 )

3) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testHashAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
+        'id' => '7baaaf2a6fcb38a5b42353a89d85519f'
         'data' => 'two'
     )
 )
```

In large parts what we are stumbling upon is really down to how mongodb and odm work.

The only proposal to fix this I can find is not a quick fix and also not something we can do without introducing some degrees of bc breakage.

* Ensure that embedded documents are expressed as `embedded-document` and do not expose an `id` field.
* Switch cases where we absolutely need to reference data to use JSON-References through the `extref` mechanism.

I think we'll have to analyse each individual definition that could be affected by such a change. In some cases we will probably need to introduce some redundancy so everything is still more or less searchable. 

In other cases clients might seriously have to step up their `$ref` game to be able to still use some services.

It looks like we need to look into this a bit. From what I saw, just downgrading doctrine-odm and it's parts won't just make this issue go away since it's trigger probably wasn't just the `doctrine/*` update but also some of the newly writable services using `recordOrigin`. The latter feature made this error visible by using human-readable ids in embedded documents. In most other cases I looked at the ids were either hidden or not human readable.

hairmare:
Introducing a MappedSuperclass like [this](http://stackoverflow.com/a/13961284) might also be a solution to parts of this issue. I'm not sure how those work when it comes to data consistency though.

src/Graviton/EmbedTestBundle/Tests/EmbedDocumentTest.php
```diff
@@ -0,0 +1,142 @@
...
+                'embedded' => [
+                    '_id' => $embedId,
+                    'name' => $embedName,
+                ],
+            ],
+            $this->dm->getDocumentCollection(Document::class)->findOne(['_id' => 'test'])
+        );
+    }
+
+    public function testDocument()
```
mrix: Hm... This test works fine. And IDs of embedded document are not converted to UUID after saving.

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -200,8 +200,8 @@ public function updateRecord($documentId, $entity)
         // In both cases the document attribute named originRecord must not be 'core'
         $this->checkIfOriginRecord($entity);
         $this->checkIfOriginRecord($this->find($documentId));
-        $entity = $manager->merge($entity);
-        $manager->flush();
+        $manager->persist($entity);
```
mrix: @hairmare Why did we use `$manager->merge($entity);` here?
Embed's hash is overwritten during merging record (check https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L1979).
I reverted `$manager->persist($entity);`, but there is an issue - we cannot save empty array into `embed-many`.

hairmare: This was switched to use `persist` to make upserting of embed-many records possible in 9669459bf95a36fd98f3a5f82af45b2d56601907. I think it was about making it possible to save empty arrays.

narcoticfresh:
@mrix indeed, the change to `merge()` was a rather quick one as we had problems with empty arrays and (i'm not sure anymore) other persists of some array changes (like remove 2nd member within 3 members).. the change to `merge()` solved this.. 

@mrix btw, damien today said you're using ur own version of the backend at the moment.. was this issue really just introduced with the current develop (doctrine bump) and the current master doesn't have this issue?

mrix:
@narcoticfresh yes, i downgraded doctrine version to `"doctrine/mongodb-odm": "1.0.0-BETA13"` on our backend instance. As I see in git log, API `v0.15.0` does not include the commit `Bump all deps due to CVE-2015-5723...` where `doctrine/mongodb-odm` was upgraded to `1.0.2`

narcoticfresh:
@mrix assuming that you have a test in this PR that demonstrates the error - can u try to push a composer.lock with the older doctrine releases as deps to see if that fixes the test?

narcoticfresh:
@mrix i was playing around in order to find a solution.. no luck so far..
so yes, remove the merge() leaves us with the 'save empty ArrayCollection problem'.. i tried different `strategy`-settings ([see here](http://doctrine-mongodb-odm.readthedocs.org/en/latest/reference/collection-strategies.html#collection-strategies)).. and it seems any other strategy then `set` makes it even worse - `set` at least only struggles with empty collections..

let us know here what luck you have trying with the downgrade..

hairmare:
@narcoticfresh I did some downgrade tests last thursday and was able to reproduce this in older versions. We may need a different set of eyes to verify the issue with older versions though.

mrix:
@narcoticfresh I was wrong. Downgrading doctrine does not fix the bug.
I was testing only on such fields `document.subdocument.embedded`.
But the behavior is different between `document.subdocument.embedded` and `document.embedded`. The new tests show the difference.


PhphUnit output using `1.0.2`:
```
[vagrant@vagrant-centos7 graviton]$ vendor/bin/phpunit src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php
PHPUnit 4.8.9 by Sebastian Bergmann and contributors.

F.F.F

You should really fix these slow tests (>500ms)...
 1. 18934ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsEmbedded
 2. 1151ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsReference
 3. 831ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsDeepReference
 4. 692ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsDeepEmbedded
 5. 588ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testHashAsEmbedded


Time: 22.42 seconds, Memory: 68.00Mb

There were 3 failures:

1) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
-        'data' => 'two'
+        'id' => 'one'
+        'data' => 'one'
     )
 )

/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:118

2) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsDeepEmbedded
Failed asserting that 404 matches expected 200.

/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:62
/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:220

3) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testHashAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
+        'id' => 'f5527e2f07183762b8f45c5f45330a2a'
         'data' => 'two'
     )
     'documents' => Array (
         0 => stdClass Object (
-            'id' => 'three'
+            'id' => '8d514d900e6e358fb4c29545ecbc4d8c'
             'data' => 'three'
         )
     )
 )

/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:320

FAILURES!
Tests: 5, Assertions: 88, Failures: 3.
```

Using `1.0.0-BETA13`:
```
[vagrant@vagrant-centos7 graviton]$ vendor/bin/phpunit src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php
PHPUnit 4.8.9 by Sebastian Bergmann and contributors.

F...F

You should really fix these slow tests (>500ms)...
 1. 23812ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsEmbedded
 2. 1448ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsDeepReference
 3. 1067ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsReference
 4. 720ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsDeepEmbedded
 5. 707ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testHashAsEmbedded


Time: 27.96 seconds, Memory: 67.50Mb

There were 2 failures:

1) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
-        'data' => 'two'
+        'id' => 'one'
+        'data' => 'one'
     )
 )

/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:118

2) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testHashAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
+        'id' => '9357919a4a79339991f8b3027a229f4d'
         'data' => 'two'
     )
     'documents' => Array (
         0 => stdClass Object (
-            'id' => 'three'
+            'id' => '81a1ff7f17ec3660a59d23839f7b76fb'
             'data' => 'three'
         )
     )
 )

/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:320

FAILURES!
Tests: 5, Assertions: 95, Failures: 2.
```

hairmare:
I switched back to `1.0.2` and got the data part somewhat working by splitting documents in to embeddable and referenceable classes over on the [`feature/298-make-embedded-documents-be-real-embedded-copies` branch](https://github.com/libgraviton/graviton/tree/feature/298-make-embedded-documents-be-real-embedded-copies).

```
$ vendor/bin/phpunit src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php
PHPUnit 4.8.13 by Sebastian Bergmann and contributors.

F.F.F

You should really fix these slow tests (>500ms)...
 1. 1590ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsEmbedded
 2. 969ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsReference
 3. 913ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsDeepReference
 4. 722ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testHashAsEmbedded
 5. 697ms to run Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest:testDocumentAsDeepEmbedded


Time: 5.31 seconds, Memory: 82.75Mb

There were 3 failures:

1) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
+        'id' => 'a47dea750e503b43b8962735f110b02d'
         'data' => 'two'
     )
 )

/Users/hairmare/Documents/git.repos/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:118

2) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsDeepEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'deep' => stdClass Object (
         'document' => stdClass Object (
-            'id' => 'two'
+            'id' => '01cb09719dbb37ae98bd90247ef4d55c'
             'data' => 'two'
         )
     )
 )

/Users/hairmare/Documents/git.repos/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:217

3) Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testHashAsEmbedded
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'id' => 'test'
     'document' => stdClass Object (
-        'id' => 'two'
+        'id' => 'f18aff5a780236359cb1f429ad045eaf'
         'data' => 'two'
     )
     'documents' => Array (
         0 => stdClass Object (
-            'id' => 'three'
+            'id' => '92a1cbd54a793d6da4faf46eacfc92d1'
             'data' => 'three'
         )
     )
 )

/Users/hairmare/Documents/git.repos/graviton/src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php:320

FAILURES!
Tests: 5, Assertions: 87, Failures: 3.
```

I'm still seeing issues with the id and am now trying to uncouple the visible `id` from it's mongodb `_id` counterpart in the embedded cases.

My branch also breaks a lot of stuff elsewhere that will need taking care of before it's ripe for rebasing it into a useable pull request.

narcoticfresh:
replaced by #314

PR #299, opened 2015-10-08
EVO-771 multipart form data

lapistano:
Introduces the ability to upload a file and its' meta data in one go.

### DIFF
--- a/src/Graviton/FileBundle/FileManager.php
+++ b/src/Graviton/FileBundle/FileManager.php
@@ -194,7 +194,7 @@ class FileManager
     {
         $action = [];
 
-        if (!empty($metaData) && array_key_exists('action', $metaData) && !empty($metaData['action'])) {
+        if (!empty($metaData) && !empty($metaData['action'])) {
             foreach ($metaData['action'] as $command) {
                 if (!empty($command) && !empty($command['command'])) {
                     $fileMetadataAction = clone $this->fileMetadataAction;
--- a/src/Graviton/FileBundle/FileManager.php
+++ b/src/Graviton/FileBundle/FileManager.php
@@ -1,0 +1,156 @@
+<?php
+/**
+ * Handles file specific actions
+ */
+
+namespace Graviton\FileBundle;
+
+use Gaufrette\File;
+use Gaufrette\FileSystem;
+use Graviton\RestBundle\Model\DocumentModel;
+use GravitonDyn\FileBundle\Document\FileMetadata;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+/**
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class FileManager
+{
+    /**
+     * @var FileSystem
+     */
+    private $fileSystem;
+
+    /**
+     * FileManager constructor.
+     *
+     * @param FileSystem $fileSystem file system abstraction layer for s3 and more
+     */
+    public function __construct(FileSystem $fileSystem)
+    {
+        $this->fileSystem = $fileSystem;
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


src/Graviton/FileBundle/Resources/config/services.xml
```diff
@@ -13,14 +13,20 @@
...
     </parameters>
     <services>
         <service id="graviton.aws_s3.client" class="%graviton.aws_s3.client.class%" factory-class="%graviton.aws_s3.client.class%" factory-method="factory">
             <argument>%graviton.aws_s3.client.args%</argument>
         </service>
 
-        <service id="graviton.file.controller.file" class="%graviton.file.controller.file.class%" parent="graviton.rest.controller">
+        <service id="graviton.file.file_manager" class="%graviton.file.file_manager.class%">
             <argument type="service" id="gaufrette.file_service_filesystem"/>
+            <argument type="service" id="gravitondyn.file.document.filemetadataaction"/>
```
hairmare: It looks like you need to inject a repository-factory so this will also run before dynamic-bundles have been generated. 

The easiest way to reproduce travis' error is probably by running dev-cleanstart.sh.

src/Graviton/FileBundle/FileManager.php
```diff
@@ -0,0 +1,215 @@
...
+     *
+     * @param array $metaData Set of meta data to be parsed for commands.
+     *
+     * @return array
+     */
+    private function transcodeAction(array $metaData)
+    {
+        $action = [];
+
+        if (!empty($metaData) && array_key_exists('action', $metaData) && !empty($metaData['action'])) {
```
q-src: `!empty($metaData) && !empty($metaData['action']` would do it, since `empty()` checks on array keys itself

src/Graviton/FileBundle/FileManager.php
```diff
@@ -0,0 +1,215 @@
...
+            foreach ($metaData['action'] as $command) {
+                // remove potentially dangerous chars
+                $commandString = preg_replace('@[^a-zA-Z0-9_-]@', '', $command['command']);
```
q-src: I would remove this check since the current workers use the `command` field only for checking if they have to do something. So they just check if this field equals `archive` or `print`, for example. 
If a future worker would somehow process this command in a way it could be dangerous (e.g. execute the command or parts of it on a shell), the worker should do this check itself - because the worker knows which chars are actually dangerous.

src/Graviton/FileBundle/FileManager.php
```diff
@@ -0,0 +1,215 @@
...
+                if (!empty($command) && $command['command'] == $commandString) {
+                    $fileMetadataAction = clone $this->fileMetadataAction;
+                    $reflection = new \ReflectionProperty(get_class($this->fileMetadataAction), 'command');
+                    $reflection->setAccessible(true);
+                    $reflection->setValue($fileMetadataAction, $commandString);
```
q-src: Can't you just use `$fileMetadataAction->setCommand($commandString)` instead of using Reflection?

lapistano: totally did not see this method .. thx. Will replace it.

src/Graviton/FileBundle/FileManager.php
lapistano: some days, one does learn something .. thx. ;)

src/Graviton/FileBundle/FileManager.php
```diff
@@ -0,0 +1,209 @@
...
+        foreach ($files as $key => $fileInfo) {
+            $entityClass = $model->getEntityClass();
+            $record = $model->insertRecord(new $entityClass());
+            $inStore[] = $record->getId();
+
+            /** @var \Gaufrette\File $file */
+            $file = $this->saveFile($record->getId(), $fileInfo['content']);
+
+            // update record with file metadata
+            $meta = new FileMetadata();
```
q-src: How about outsourcing this into `FileDocumentFactory`? Moreover you could populate the `FileMetaData` object there and then also add the `FileMetaDataAction`.

PR #304, opened 2015-10-14
fix HEAD handling

narcoticfresh:
i know, we all didn't want to support HEAD requests in the first place ;-/
but as it turns out, some browsers make them and there is no way to disable that.

Graviton currently handles them in a rather inconsistent way - it answers it with an 2** status code, but actually processes the data request in `allAction()` in `RestController` but then has an empty response body. I've seen that it gets data by the delayed answer, the presence of the `x-total-count` header in the response and dev.log output.

this PR handles HEAD the same as we handle OPTIONS. this suffices as the only thing the client (m$ie) is looking for in the HEAD response is the presence of a `Content-Type` header (afaik), which we provide in the OPTIONS request.

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
@@ -638,6 +638,27 @@ class AppControllerTest extends RestTestCase
     }
 
     /**
+     * requests on OPTIONS and HEAD shall not lead graviton to get any data from mongodb.
+     * if we page limit(1) this will lead to presence of the x-total-count header if
+     * data is generated (asserted by testGetAppPagingWithRql()). thus, if we don't
+     * have this header, we can safely assume that no data has been processed in RestController.
+     *
+     * @return void
+     */
+    public function testNoRecordsAreGeneratedOnPreRequests()
+    {
+        $client = static::createRestClient();
+        $client->request('OPTIONS', '/core/app/?limit(1)');
+        $response = $client->getResponse();
+        $this->assertArrayNotHasKey('x-total-count', $response->headers->all());
+
+        $client = static::createRestClient();
+        $client->request('HEAD', '/core/app/?limit(1)');
+        $response = $client->getResponse();
+        $this->assertArrayNotHasKey('x-total-count', $response->headers->all());
+    }
+
+    /**
      * test getting schema information from canonical url
      *
      * @return void
--- a/src/Graviton/RestBundle/Routing/Loader/ActionUtils.php
+++ b/src/Graviton/RestBundle/Routing/Loader/ActionUtils.php
@@ -58,7 +58,6 @@ class ActionUtils
         }
 
         $route = new Route($pattern, $defaults, $requirements);
-
         $route->setMethods($method);
 
         return $route;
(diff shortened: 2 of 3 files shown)


src/Graviton/CoreBundle/Tests/Controller/AppControllerTest.php
```diff
@@ -638,6 +638,27 @@ public function testGetAppSchemaInformation()
...
+     * have this header, we can safely assume that no data has been processed in RestController.
+     *
+     * @return void
+     */
+    public function testNoRecordsAreGeneratedOnPreRequests()
+    {
+        $client = static::createRestClient();
+        $client->request('OPTIONS', '/core/app/?limit(1)');
+        $response = $client->getResponse();
+        $this->assertArrayNotHasKey('x-total-count', $response->headers->all());
```
hairmare: Is this really what we need to support? Should a client that just wants to know the total count of records be able to do exactly that using the `HEAD` method? I'm aware that this usage of `HEAD` won't really work nicely in combination with CORS, but I think we had some clients a bit ago that where using `HEAD` requests for stuff not related to CORS.

narcoticfresh: @hairmare AFAIK no client does HEAD requests knowingly.. they are just issued automatically by Internet Explorer in some cases before the preflight OPTIONS.. so MSIE really does HEAD, then OPTIONS and then the actual GET..

Even if a client *would* do that for the solely purpose to get the record count, that would be quite unnecessary as the record count is anyway included in the reponse of the actual GET.

narcoticfresh:
@hairmare one more thought: if a client really just wants to count (and no actual data), issuing a HEAD request at the moment makes Graviton to prepare 'default pagination size' records but not render it. an expensive request.

if a client wants to have the tinyest amount of data with record count, it's still cheaper to issue a `GET /?limit(1)` to the corresponding service as it only prepares 1 entity. i've seen that `/?limit(0)` doesn't seem to work obviously..

narcoticfresh:
@hairmare hmm.. kinda off topic: at the moment, passing `limit(0)` means "no limit".. is this what we want? would be convenient as the "stupid" clients are passing `limit(99999999)` now and "could* use `limit(0)`...

.....*or*... should we make limit(0) return no records at all..? offering a slim way to just get record count?

hairmare:
@narcoticfresh I was under the impression that we already have (web-)clients that do `HEAD` requests. Another case where they make sense is to check if an auth-token (received through legacy) is still valid or would result in a failed precondition response.

I'm not sure if the cases that clients use `HEAD` really exist or are sane. I just remember that I was talking about `HEAD` requests to someone some time ago. Maybe we should ask around?

narcoticfresh:
i think we discussed HEAD once when they were faulty answered by legacy backend.. and now it came to me (via Huck), that HEAD requests go awfully long - that's what I'm trying to fix.. he suggested we should try the client to use Chrome so no HEAD is issued.. so i think this side is not actively using HEAD..

I'll ask in the chat..

hairmare:
@narcoticfresh Let me know what conclusion was reached and I'll merge this as needed.

narcoticfresh:
@hairmare it has been clarified..

* the 'infinite loop, cupertino' centric client doesn't use HEAD at all
* the browser centric client uses HEAD for a purpose that is actually served by OPTIONS. they need it to get Link _self header information about a single resource. they don't need X-Total-Count in the HEAD response.

so IMHO we can merge this now..

hairmare:
Thanks for looking into this.

PR #305, opened 2015-10-14
remove autoloader wtf

narcoticfresh:
just something i've seen.. is this useful for anything? maybe an old cloud workaround? i think `/app/htdocs` does not exist anywhere, not even in the cloud (it's `/app/app/src`).. so can remove?


hairmare:
:+1: will merge when CI is through

PR #307, opened 2015-10-15
Feature/evo 1044 schema title

PanteraLX:


src/Graviton/GeneratorBundle/Resources/skeleton/model/schema.json.twig
```diff
@@ -6,6 +6,10 @@
   "description": "@todo replace me",
 {% endif %}
 
+{% if json is defined %}
+    "title": {{ json.getTitle()|json_encode() }},
```
hairmare: It might be more pragmatic to do this as follows:

```twig
    "title": "{{ json.getTitle() }}",
```

getTitle will always return a string anyway.

hairmare:
Merged in #312

PR #309, opened 2015-10-16
EVO-949 - fix translation of non-existant translation domains at container build time

narcoticfresh:
this tries to solve the translation problems we observed mostly in the cloud. but it worked mostly locally.

**why did it not work in the cloud?**
at container build time, the translator resource files in the bundles that trigger the load (like `core.de.odm`) are searched and put as a *fixed* list in the *compiled* container.

as we added translations for new domains, they were never loaded as those files were never present at container build time. 

**why did it work locally (mostly)?**
locally, after we created some domains, it also created the trigger file (like `hans.de.odm`). as we develop, we often run a script like `dev-cleanstart.sh` - which rebuilds the container among other things. once we did that, translations for that domain did work.

so -> it never worked in the cloud because there, the container is never rebuilt after initial deployment.

**what this PR introduces**
* change the behavior of `PostPersistTranslatableListener`<br />instead of touching and/or removing files, this class now just dispatches an `Event` object.
* introduce the class `I18nUtilsCaching`<br />this class handles everything that concerns the handling of the cache as needed and feeds the `Translator` with the added files at runtime.
* introduce `I18nCachingListener`<br />Listens to the events from `PostPersistTranslatableListener` and calls the corresponding methods on `I18nUtilsCaching`.
* introduce a `Translator` that replaces the `FrameworkBundle\Translator`<br />It just adds one line where it calls `I18nUtilsCaching` as a away to inject the runtime added files into the Translation loader process.
* an integration test of the whole thing as well as a unit test just for the `I18nUtilsCaching` class.

**gains**
i really hope this change brings some performance improvements, also while running the unit tests. before, we touched and/or removed files at every doctrine persist event in `PostPersistTranslatableListener` and significant stuff has been done there.

now, all those `Event`s (and the resulting cache invalidations) are processed only once at the time the `Kernel` dispatches the `kernel.terminate` event. additionally, i remove the cached catalogue in just 1 `Finder` iteration.

**varia changes**
* the integration test of this feature *needs* to be on a *clean* cache. If we re-run the same test suite on an already existing cache, our runtime translation resources could already be picked up locally from the run before. so i introduced the cleaning of the `test` env in the phpunit bootstrap.
* i configured the `DoctrineCacheBundle` as I want to cache stuff I generate. If not, I must create the resulting array in every request. Note that this is the "Doctrine cache library", *not* the cache of doctrine that we use itself.

### DIFF
--- a/src/Graviton/I18nBundle/Resources/config/services.xml
+++ b/src/Graviton/I18nBundle/Resources/config/services.xml
@@ -23,6 +23,7 @@
     <parameter key="graviton.i18n.request.class">Symfony\Component\HttpFoundation\Request</parameter>
     <parameter key="graviton.i18n.listener.postpersisttranslatable.class">Graviton\I18nBundle\Listener\PostPersistTranslatableListener</parameter>
     <parameter key="graviton.i18n.utils.class">Graviton\I18nBundle\Service\I18nUtils</parameter>
+    <parameter key="graviton.i18n.cacheutils.class">Graviton\I18nBundle\Service\I18nCacheUtils</parameter>
     <parameter key="graviton.i18n.form.transformer.translatabletodefaultstring.class">Graviton\I18nBundle\Form\DataTransformer\TranslatableToDefaultStringTransformer</parameter>
     <parameter key="graviton.i18n.form.type.translatable.class">Graviton\I18nBundle\Form\Type\TranslatableType</parameter>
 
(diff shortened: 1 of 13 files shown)


src/Graviton/I18nBundle/Resources/config/services.xml
```diff
@@ -19,12 +19,18 @@
...
     <parameter key="graviton.i18n.loader.doctrineodm.class">Graviton\I18nBundle\Translation\Loader\DoctrineODMLoader</parameter>
     <parameter key="graviton.i18n.request.class">Symfony\Component\HttpFoundation\Request</parameter>
     <parameter key="graviton.i18n.listener.postpersisttranslatable.class">Graviton\I18nBundle\Listener\PostPersistTranslatableListener</parameter>
     <parameter key="graviton.i18n.utils.class">Graviton\I18nBundle\Service\I18nUtils</parameter>
+    <parameter key="graviton.i18n.cacheutils.class">Graviton\I18nBundle\Service\I18nCacheUtils</parameter>
     <parameter key="graviton.i18n.form.transformer.translatabletodefaultstring.class">Graviton\I18nBundle\Form\DataTransformer\TranslatableToDefaultStringTransformer</parameter>
     <parameter key="graviton.i18n.form.type.translatable.class">Graviton\I18nBundle\Form\Type\TranslatableType</parameter>
+
+    <!-- should we move this to somewhere else? -->
+    <parameter key="translator.class">Graviton\I18nBundle\Translator\Translator</parameter>
```
lapistano: translator.class  vs  graviton.i18n.translator.class?

narcoticfresh: ehm, no.. i want to override the default translator here.. that's the purpose of this line..

lapistano: ah, i did not get this.. thx

PR #310, opened 2015-10-17
Feature/evo 771 enhancements

lapistano:
completes the multipart/form-data FileUpload by adding it to PUT and fixes the ingored 'link' section of send request metadata.


src/Graviton/FileBundle/Controller/FileController.php
```diff
@@ -218,4 +219,106 @@ private function determineRoutes($routeName, array $files, array $routeTypes)
...
+    {
+        // split content
+        $contentType = $request->headers->get('Content-Type');
+        list(, $boundary) = explode('; boundary=', $contentType);
+
+        $content = $request->getContent();
+        list(, $metadataInfo, $fileInfo) = explode($boundary, $content);
+        $attributes = array_merge(
+            $request->attributes->all(),
+            $this->extractMetaData(explode("\r\n", ltrim($metadataInfo)))
```
q-src: I'd put the `explode(...)` into `extractMetaData()`, since the explosion is part of the extraction (same for line 266). Then you could just call `$this->extractMetaData($metadataInfo)`, which is less confusing.

PR #313, opened 2015-10-19
introduce means to change rendered urls by configuration

narcoticfresh:
Graviton rightfully doesn't care what urls it generates (like `extref`s or self urls).. As long as it's resolvable by `Router` and the request actually hits us, we are fine..

In all future scenarios, Graviton will be proxyed by some component. Be it something like Tyk or Kong or some L-Backend - some component will be in front. This arises the need of content rewriting in the JSON. 

At our first client, it seems it goes to a point where we'll put L-Backend (a shrinked version of it only containing the `ProxyBundle` and `/api/auth`) on the same host as Graviton (= meaning same hostname). 
Under that scenario it would be optimal if Graviton generates the URL from the L-Backend - and as they are on the same host and L-Backend proxies the request) - no content and/or content rewriting is necessary anymore. Provided we can generate different URLs in the output - that is, that this PR introduces.

On a vanilla Graviton, there will be no changes nor will this be activated. But it facilitates proxying scenarios in the future massively..

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -8,6 +8,10 @@ parameters:
     graviton.security.services.whitelist: {}
     graviton.rest.special_mimetypes: {}
 
+    graviton.router.host: null
+    graviton.router.port_http: null
+    graviton.router.port_https: null
+
     graviton.security.oauth.github.client_id: ~
     graviton.security.oauth.github.client_secret: ~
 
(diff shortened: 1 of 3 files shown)


lapistano:
@narcoticfresh should this be merged any time soon?

narcoticfresh:
@lapistano yope

hairmare:
mayhaps?

narcoticfresh:
hopefully so one day maybe.. i mean.. who knows.. it *could* happen

PR #315, opened 2015-10-21
Translatable array support

mrix:
This is a part of EVO-581. I have a task where I have to add definitions containing arrays of translatable fields.

I add `field[]` item to `TranslatableInterface::getTranslatableFields()` to mark `field` as array of translatable.
Maybe it makes sense to add new method like `TranslatableInterface::getTranslatableArrayFields()`. Please tell me if you think the second option is more reasonable.


narcoticfresh:
@mrix i tagged https://github.com/libgraviton/GravitonTestServicesBundle as v0.14.0, please change your deps to that and solve conflicts..

mrix:
@narcoticfresh I merged `develop` branch and updated test services dependencies.

PR #318, opened 2015-10-22
make PUT on /file *not* return a Location header

narcoticfresh:
no service in Graviton returns a `Location` header on a `PUT` request. but `/file` does. this PR makes `/file` behave as any other service in this regard.

some clients wrongly try to follow present `Location` headers in all methods except `POST` - sending this will lead a browser like Chrome to try to make OPTIONS preflight request and/or directly GET the file - this is not what we want, especially with a service like `/file` that potentially serves big objects..

### DIFF
--- a/src/Graviton/FileBundle/Controller/FileController.php
+++ b/src/Graviton/FileBundle/Controller/FileController.php
@@ -160,13 +160,7 @@ class FileController extends RestController
 
         $response->setStatusCode(Response::HTTP_NO_CONTENT);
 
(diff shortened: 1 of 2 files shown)


narcoticfresh:
@lapistano history browsing in github is a bit tricky.. but if you check what release is used on nova cloud currently - v0.31.0 - you see that there, PUT doesn't send a Location header..

so this was freshly introduced somewhere since the v0.31.0 tag..
https://github.com/libgraviton/graviton/blob/v0.31.0/src/Graviton/FileBundle/Controller/FileController.php

PR #314, opened 2015-10-20
Make embedded documents be defined as real <embedded-document>

hairmare:
Basically this contains work on making all doctrine `<embed-one>` and `<embed-many>` relationships point to a document that has been defined as `<embedded-document>`. This is #298 and [EVO-1099](https://issue.swisscom.ch/browse/EVO-1099).

When done this should fix #298. It has been quite tricky to find a solution that does not break BC everywhere, but I believe that this is almost there.

To check this you can do the following while ignoring build failures.

```bash
rm -rf src/GravionDyn/ && sh dev-cleanstart.sh
rm -rf app/cache/ && vendor/bin/phpunit src/Graviton/CoreBundle/Tests/Controller/EmbeddingDocumentsTest.php
```

This PR contains my unrebased work on the theme and will need rebasing after everything works. I was planning on cleaning this up and squashing it before making a pr. I'm making one now since I'll be traveling and this can't wait until I return.

* [x] make generator generate both embedded and regular documents (using the suffix `Embedded` for embeddable variants)
* [x] make docs generated from json-defs use embedded versions of documents throughout
* [x] make sure that the `id` field in embedded documents is a simple string and not a id required by docgtrine
* [x] add a replacement id (`realId`) so doctrine still works
* [x] hide the `realId` using the serializer so it never gets exposed
* [x] change direction of class relationships from `DocumentEmbedded extends Document` to `Document extends DocumentEmbedded` since doctrine/mongodb-odm > 1.0.0 does not support disabling id fields in subclasses (note that composer.json is still locked to a beta, which it should not be in the end)
* [ ] change generator so `id` field only gets added to `Document` and not `DocumentEmbedded`
* [ ] consider getting rid of `realId` in `Document` by moving it to `DocumentEmbedded` (possibly by way of a mapped superclass containing everything else)
* [x] make fixtures loading also work with embedded structures (they always write references as it seems)
* [x] make sure swagger generation works again (This broke and I haven't found the time to look into it)
* [ ] make it work for `deep.nested.stuff` (probably has to do with rewriting `class:GravitonDyn\...[]` style types to `class:GravitonDyn\...Embedded[]`, I've concentrated on simple cases so far)
* [ ] check if `PersonContactEmbedded` and `HashEmbedded` are really needed (I had lots of stuff on my working copy when adding them and they might be obsolete again, `PersonContactEmbedded` will most likely not work due to how it's id is defined (the doctrine/mongodb-odm > 1.0.0 issue))
* [x] check that integrations tests work again
[...]

PR #321, opened 2015-10-23
EVO-1099: fix embedded documents

narcoticfresh:
this was a heated topic and a long journey to get here. this one replaces #314 to have a cleaner git history. i'm sorry for the stolen commits guys - I tried like 10 times to rebase but it had so many conflicts that I decided it's easier to start in a new branch, reset to develop and re-commit all changes in a consolidated fashion.

this is an important change that we need for graviton 1.0 and the first client delivery.. only with this PR, we finally handle saving and handling of all relations (be it ref or embed) correctly on the MongoDB level.

it introduces
* changes to GeneratorBundle to generate an *Embedded version for every document alongside the normal Document class (with all that comes with it - meaning also repos, services and so on)
* correctly generate the services. If now a service has a field defined like "field.subfield", the class FieldEmbedded is used for that hash, containing field "subfield"
* Some Bundles (Person and File) used some wrong classes that also get corrected.

testing
* Since the beginning of this feature, we had an initial test (`EmbeddingDocumentsTest`) from @mrix that demonstrated the wrong behavior. We worked until this test was all good, what it is now. This test now is part of our test suite; the test service definitions for that are now part of the `TestBundle`.


composer.json
```diff
@@ -61,7 +61,7 @@
         "libgraviton/codesniffer": "~1.3",
         "lapistano/proxy-object": "dev-master",
         "johnkary/phpunit-speedtrap": "~1.0",
-        "graviton/test-services-bundle": "^0"
+        "graviton/test-services-bundle": "*"
```
lapistano: this will checkout 'dev-develop' is this ok? Why not tag a new version?

narcoticfresh: hm, it shouldn't.. `*` means "any version", meaning a tag..
the composer.lock also looks like that, 
```

Comment (lapistano) [inline composer.json]: ok. I guess I was confused by the 'minimum-stability: dev' configuration in the configuration.

File: src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentBase.php.twig [hunk 2]
```diff
@@ -0,0 +1,254 @@
...
+use JMS\Serializer\DeserializationContext;
+
+{% set translatableFields = [] %}
+{% set preTranslatedFields = [] %}
+
+/**
+ * {{ base }}Document\{{ document }}Base
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+* @license  http://opensource.org/licenses/gpl-license.php GNU Public License
```
lapistano: indentation

src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentEmbedded.php.twig
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * document class for {{ base }}Document\{{ document }}Embedded
+ */
+
+namespace {{ base }}Document;
+
+/**
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+* @license  http://opensource.org/licenses/gpl-license.php GNU Public License
```
lapistano: indentation (every docblock in file)

PR #324, opened 2015-10-26
fix extref in hash

narcoticfresh:
so.. this one adds the ability to save and display extrefs inside hashes ('free form objects')
obviously, we have no idea where they can be.. so we need to recurse that hash..

**what it does**
* recurse the hash (**important:** remember, that we *only* recurse the current hash! so in a big service like *consultation*, only a subset of objects are hashes - and there, we iterate that one smaller subhash)
* if we find an `$ref` field, transform it to a mongodb ref and save it like that under a new name (it needs to be a new name as `$ref` would need an `$id` field on the *same level!*)
* on output, transform it back to `$ref` - we can spot our fields by the unique keyword fieldname
* for all extref stuff, we use the in-place `ExtRefConverter`

**drawbacks**
remember: this is a workaround. ~~but a bad one actually. once we get `consultations` saved like this, we need to manually migrate them to the newer format of *real* extrefs (because the presence of the `_dynExtRef` fieldname!)~~
*update*: i renamed the field we save to internally to `ref` - the same way it would be saved if it would be a real extref. this way, we don't need to migrate as long as the references will be named `$ref` in the service.

i thought about it and it should be fairly safe to do that. *only drawback now*: you *cannot* have a field named "ref" anywhere in the service or it'll try to convert it to a `$ref` - so if it doesn't contain a valid url, it will fail.. but we don't have this anyway..

**testing**
showcasecontroller test has been enhanced to test this..

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Controller/ShowcaseControllerTest.php
+++ b/src/Graviton/CoreBundle/Tests/Controller/ShowcaseControllerTest.php
@@ -258,6 +258,52 @@ class ShowcaseControllerTest extends RestTestCase
     }
 
     /**
+     * test if we can save & retrieve extrefs inside 'free form objects'
+     *
+     * @return void
+     */
+    public function testFreeFormExtRefs()
+    {
+        $minimalExample = $this->postCreationDataProvider()['minimal'][0];
+
+        $document = json_decode(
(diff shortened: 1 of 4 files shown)


narcoticfresh:
@lapistano see my update above in the comment.

PR #325, opened 2015-10-27
EVO-1281 - make /core/version urls the same as all others (having / at the end)

narcoticfresh:
as said.. /core/version should be /core/version/ (with trailing slash) so it's consistent with all other services)


janfriedli:
According to @hairmare i did not implement it with a trailing slash because it's not a collection. /core/version is different to other services...

narcoticfresh:
@janfriedli hm, that's actually a valid point.. it's the resource `version` of the `core` collection.. i see..  but then it's kinda misleading, that `GET /core/` shows nothing.. 

@lapistano should we revert the change?

PR #326, opened 2015-10-30
Wrong self-version in core/version

PanteraLX:
The core/version service returns the wrong version of the root package ('self'). We are reading the self-version out of a composer show command. Since there is no ('handwritten') version tag in the composer.json file anymore, composer gets the information from somewhere else. Composer adds a 4th version after the patch version for a reason that i could not figure out. Nevertheless, this 4th identifier adds no further information and is not SemVer-valid, so it has to be removed.

### DIFF
--- a/src/Graviton/CoreBundle/Tests/Services/CoreVersionUtilsTest.php
+++ b/src/Graviton/CoreBundle/Tests/Services/CoreVersionUtilsTest.php
@@ -31,10 +31,14 @@ class CoreVersionUtilsTest extends RestTestCase
      */
     public function testGetVersionNumber($versionString, $expectedVersion)
     {
-        $this->coreVersionUtils = $this->getMockBuilder('Graviton\CoreBundle\Service\CoreVersionUtils')
-            ->disableOriginalConstructor()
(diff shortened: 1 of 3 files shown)


src/Graviton/CoreBundle/Tests/Services/CoreVersionUtilsTest.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * @dataProvider provideVersionStringForGet
+     *
+     * @param string $versionString   String which should be checked
+     * @param string $expectedVersion Expected Version string
+     * @return void
+     */
+    public function testGetVersionNumber($versionString, $expectedVersion)
+    {
+        $this->coreVersionUtils = $this->getMockBuilder('Graviton\CoreBundle\Service\CoreVersionUtils')
```
lapistano: It is most of the time a very bad idea to mock the SUT. This usually leads to self full filling tests.
Please do use a fully instantiated CoreVersionUtils object and mock the constructor parameters if necessary.

lapistano:
pls see why Q'nA checks have failed.

PR #327, opened 2015-10-30
Feature/fileupload metadata override fix

lapistano:
https://issue.swisscom.ch/browse/EVO-1328

### DIFF
--- a/src/Graviton/FileBundle/FileManager.php
+++ b/src/Graviton/FileBundle/FileManager.php
@@ -239,11 +239,13 @@ class FileManager
      * @param integer      $fileSize Size of the uploaded file
      * @param array        $fileInfo Additional info about the file
      * @param FileDocument $fileData File data to be updated
+     *
+     * @return void
      */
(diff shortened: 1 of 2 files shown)


src/Graviton/FileBundle/FileManager.php
```diff
@@ -232,24 +233,44 @@ private function getRecord(DocumentModel $model, FileDocument $fileData = null,
...
+     * @return void
      */
-    private function initOrUpdateMetadata(FileDocument $file, $fileSize, array $fileInfo, array $actions)
+    private function initOrUpdateMetadata(FileDocument $file, $fileSize, array $fileInfo, FileDocument $fileData = null)
     {
         $meta = $file->getMetadata();
+        $actions = [];
+
+        if (!empty($fileData)) {
+            $actions = (!empty($fileData->getMetadata()->getAction()->toArray()))
```
q-src: Can't you just do `$fileData->getMetadata()->getAction()->toArray()` directly (Same for line 254)?

lapistano: Unfortunately not .. every part of $fileData->getMetadata()->getAction() can theoretically return »null« which will break the call of toArray() in the end.

PR #328, opened 2015-11-02
validate boolean fields correctly

samsamann:
This PR provides a new form type called 'strictBoolean'. This gives us the ability to handle boolean fields correcty. Now, a required boolean field can only have false or true as value.


src/Graviton/DocumentBundle/Form/DataTransformer/BooleanTransformer.php
```diff
@@ -0,0 +1,88 @@
...
+    public function reverseTransform($value)
+    {
+        if (!isset($this->propertyPath)) {
+            throw new RuntimeException('propertyPath must be set');
+        }
+
+        $originData = $this->submittedData;
+        if (is_array($this->submittedData) && preg_match('@^\[+[0-9]\]$@', $this->propertyPath)) {
+            $index = (int) trim($this->propertyPath, '[]');
+            $originData = $this->submittedData[$index];
```
lapistano: even, if »submittedData« is an array it is not said there is an element named as the value of $index. So $originData will be null and the content valid.

src/Graviton/DocumentBundle/Form/Type/StrictBooleanType.php
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param FormBuilderInterface $builder The form builder
+     * @param array                $options The options
+     * @return void
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $dataTransformer = clone $this->transformer;
+        $dataTransformer->setPropertyPath($builder->getPropertyPath());
+        $dataTransformer->setSubmittedData($options['submitted_data']);
```
lapistano: $options['submitted_data'] might not exist and the script will proceed only logging a Notice. Is this ok?

samsamann: even, if $options['submitted_data'] is not set on the type, this option will still exist with ``null``. But I agree the default value is not very good.

src/Graviton/DocumentBundle/Form/DataTransformer/BooleanTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->submittedData = $submittedData;
+    }
+
+    /**
+     * @inheritDoc
+     *
+     * @param mixed $value The value in the original representation
+     * @return mixed The transformed value
+     */
+    public function transform($value)
```
lapistano: from the parents DocBlock: 
> By convention, transform() should return an empty string if NULL is passed.

I think the transformer should honour that.

src/Graviton/DocumentBundle/Form/DataTransformer/BooleanTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+        return $value;
+    }
+
+    /**
+     * @inheritDoc
+     *
+     * @param mixed $value The value in the transformed representation
+     * @return mixed The original value
+     */
+    public function reverseTransform($value)
```
lapistano: from the parents DocBlock: 
> By convention, transform() should return an empty string if NULL is passed.

I think the transformer should honour that.

PR #336, opened 2015-11-23
Bump dependencies

hairmare:
~~I'll tag a release as soon as this is in develop.~~ I'm trying again to bump all the things. If it still fails we should probably investigate.


lapistano:
@hairmare what is this on waiting for? is it still necessary?

hairmare:
@lapistano It had some [weird fails in CI](https://travis-ci.org/libgraviton/graviton/builds/92665263) when it last ran. I just rebased it today to see what happens.

PR #340, opened 2015-12-07
fixe like() searches on identifier fields

narcoticfresh:
i took a look at EVO-1782. this led me first to believe it's not that trivial.

after digging deeper and creating [this issue on the mongo-odm project](https://github.com/doctrine/mongodb-odm/issues/1294), it seems that there is an easy fix in our context.

as it turns out, we *wrongly* specify `type='string'` on identifier fields. this makes `mongo-odm` to use the `StringType` instead of the `CustomidType` what would be used otherwise and *also results in a string based ID in MongoDB*.

This is important as `StringType` casts everything to `string` (obviously) for ID based searches - `CustomidType` doesn't do that, it preserves the `\MongoRegex` instance.

So, removing `type='string'` from our generated services indeed allows the user to issue `like()` RQL expression on the `id` field! And it doesn't change anything in the DB, our IDs are still normal `String` fields.


hairmare:
Looks good. I'll merge when CI is through... I'm assuming we will want to tag a release containing this as well as my changes from #341 for demo-time tomorrow.

PR #341, opened 2015-12-07
Remove antimattr/mongodb-migrations-bundle

hairmare:
This replaced the bundle with some code under our own control. For the most part I needed to do this since it was rather hard to override some of the internals of the commands in the bundle.

This PR uses the symfony dic to configure the *raw* command from antimattr/mongodb-migrations in the simplest possible fashion. When calling the commands using ``graviton:mongodb:migrate`` it injects the needed config as a helper and makes sure to do so early enough.


narcoticfresh:
@hairmare ok; i'll merge this when CI is through

PR #342, opened 2015-12-10
EVO-1383 fix datetime deserialization

mrix:
This fixes the input format of date fields (see http://symfony.com/doc/current/reference/forms/types/datetime.html#input).
Current implementation returns date values as strings in `$form->getData()`.
Such behavior doesn't allow to implement services like calculators, where we have to respond immediately with given data.

I added the new test. Without my changes the test throws the exception:
```
DatetimeDeserializationTestController.php on line 41:
Graviton\ExceptionBundle\Exception\SerializationException {#2283
  -statusCode: 500
  -headers: []
  #message: """
    exception 'JMS\Serializer\Exception\LogicException' with message 'Expected object but got string. Do you have the wrong @Type mapping or could this be a Doctrine many-to-many relation?' in /sync/graviton/vendor/jms/serializer/src/JMS/Serializer/SerializationContext.php:68\n
    Stack trace:\n
    #0 /sync/graviton/vendor/jms/serializer/src/JMS/Serializer/GraphNavigator.php(143): JMS\Serializer\SerializationContext->isVisiting('2015-12-10 12:0...')\n
    #1 /sync/graviton/vendor/jms/serializer/src/JMS/Serializer/GenericSerializationVisitor.php(140): JMS\Serializer\GraphNavigator->accept('2015-12-10 12:0...', Array, Object(JMS\Serializer\SerializationContext))\n
```

PR #345, opened 2015-12-29
bump all deps

narcoticfresh:
i'm mainly after `graviton/json-schema`, but let's update all our deps..


hairmare:
Makes sense. I'll merge when CI is through. It looks like you are bumping the test services for #342 as well. Let's see what happens there :)

PR #352, opened 2016-01-18
add additional field to the file service

samsamann:
With this PR we add an additional key/value store named additionalProperties to the metadata.
We need this new field, because we want print and archive PDF and other files. Our printing and archive system needs more information about which printer must be used and where can the file be stored for example.

### DIFF
--- a/src/Graviton/FileBundle/FileDocumentFactory.php
+++ b/src/Graviton/FileBundle/FileDocumentFactory.php
@@ -39,17 +39,25 @@ class FileDocumentFactory
     /**
      * Provides an instance of FileMetadata
      *
-     * @param string $id             Identifier of the file.
-     * @param int    $size           Size of the file.
-     * @param string $filename       Name of the file.
-     * @param string $mimetype       Mime-Type of the file.
-     * @param array  $actions        List of actions to be executed.
-     * @param string $additionalInfo Additional file information.
+     * @param string $id              Identifier of the file.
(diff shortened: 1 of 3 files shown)


hairmare:
Didn't we just do this in #347?

samsamann:
we must use a key/value store and not only a string, because our printing system needs several information for example frameNr or formNr.
I now it's not really nice, but at the moment we have a customer who use the print action. So we can't consume an second severice for these purposes.

lapistano:
Why was the idea about passing the data as serialized JSON string dismissed?

hairmare:
It could technically be done in the field from #347 but that certainly feels like this:

dawg

lapistano:
sure :) but then there is no other way but to add a new field

narcoticfresh:
our big pain was that #347 was not discussed with us prior.. there we could've said that it should be an array to allow more then just a string - as we know, the tool using that field cannot make any code changes for half a year (for whatever reason).. and i totally agree with @hairmare that putting `json_encode`d strings inside that field is a pain in the ass..

we shall introduce this new array field and make the other system using the plain string field to use this new array field, allowing us to deprecate the simple string field.

hairmare:
I still hate this field thing... It makes for one of those badly documented key/value silos that end up with way too much important data in them and are hard to keep clean.

The only way forward I see would be to split the printing functionality into it's own service. The expense of having a separate print service is so miniscule that it would be well worth it.

PR #353, opened 2016-01-27
evo-2472 and evo-2476 security authentication

janfriedli:


src/Graviton/CoreBundle/Resources/config/services.xml
```diff
@@ -17,9 +17,11 @@
         <parameter key="graviton.core.service.coreversionutils.class">Graviton\CoreBundle\Service\CoreVersionUtils</parameter>
         <parameter key="graviton.core.links" type="collection">
             <parameter>graviton.core.static.version.get</parameter>
+            <parameter>graviton.security.static.whoami.get</parameter>
```
lapistano: would like to see that graviton.core has an alias for the »graviton.security.static.whoami.get« parameter to be able to define it on global level without the need to redeploy the CoreBundle. E.g move <parameter key="graviton.core.links" type="collection"> to the configuration definition of the CoreBundle to be able to define it in file config.yml

src/Graviton/CoreBundle/Resources/config/services.xml
```diff
@@ -17,9 +17,11 @@
...
         </parameter>
         <parameter key="graviton.core.main.path.whitelist" type="collection">
             <parameter>/core/version</parameter>
+            <parameter>/person/whoami</parameter>
```
lapistano: same here

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -681,4 +699,18 @@ private function getRouteName(Request $request)
 
         return $routeName;
     }
+
+    /**
+     * Get the user who's logged in
+     *
+     * @return Mixed | false
```
lapistano: pls be strict on the returned type!
in case there is no token available throw an exception accordingly and catch it in the callee to handle it correctly.

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -95,11 +105,12 @@ public function setRepository(DocumentRepository $repository)
...
      * {@inheritDoc}
      *
-     * @param \Symfony\Component\HttpFoundation\Request $request Request object
+     * @param Request $request The request object
+     * @param Object  $user    The user
      *
      * @return array
      */
-    public function findAll(Request $request)
+    public function findAll(Request $request, $user = false)
```
lapistano: Same here pls expect a user object or NULL (to define the optional character of the argument)

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -313,4 +328,15 @@ private function getDefaultLimit()
...
     }
+
+    /**
+     * @param Boolean $active active
+     * @param String  $field  field
+     * @return void
+     */
+    public function setFilterByAuthUser($active, $field)
+    {
+        $this->filterByAuthUser = is_bool($active) ? $active : false;
```
lapistano: $this->filterByAuthUser = is_bool($active) ?: false;
dies the same but leaner ;)

src/Graviton/SecurityBundle/Controller/WhoAmIController.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Return Consultant object
+     *
+     * @return \Symfony\Component\HttpFoundation\Response $response Response with result or error
+     */
+    public function whoAmIAction()
+    {
+        if ('anonymous' !== $this->getUser()->getUsername()) {
```
lapistano: will cause a NOTICE in the current implementation.. see comments above

PR #354, opened 2016-01-28
EVO-388: Create a shortcut to access gravitons app/console in wrapper context

PanteraLX:
When we want to access gravitons app/console in a wrapper context, we have to dig deep in the wrappers vendor folder. This PR should create a link to the console file in vendor/bin (the link is named graviton).


.gitignore
```diff
@@ -12,8 +12,7 @@ web/explorer
 mongodata/
 src/GravitonDyn/*
 app/config/parameters.yml
-app/config/parameters_local.xml
-bin
+app/config/parameters_local.xmlgit
```
lapistano: typo?

bin/graviton
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env php
+<?php
+
+$path = __DIR__ . '/../app/console';
```
lapistano: sure this works? 
this script will be copied in to __DIR__/vendor/bin and the console script will be located in __DIR__/vendor/graviton/graviton/bin ... I do not see where the $path variable does take this into account. pls verify in a wrapper situation.

PanteraLX: As far as I know is vendor/bin/graviton only a simlink to vendor/graviton/graviton/bin/graviton. I tested it in wrapper context and it worked

lapistano: nice .. merci

PR #355, opened 2016-02-01
Allow injected SYMFONY__ envs to contain json

hairmare:
They are only interpreted as json if they start with either `{` or `[` as to not disturb the old behaviour where they where but simple strings.


narcoticfresh:
@hairmare push should be the same as pr in this case, let's not wait for travis

PR #357, opened 2016-02-02
add more envs to incenteev map

narcoticfresh:
we need this to make the 'whoami' configurable for cloudfoundry environment.. I'll need to tag ASAP again (and bump..)..

I'm not sure about the current Docker environment.. we need to be able to always return a dummy consultant, *not* the string "anonymous". and as those params are directly written into the compiled container (not as getParameter() calls but 1:1), we **can also not** override them via the `SYMFONY__` envs.. so what to do there?

but first, let's merge this for CF..

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "ddf5c643a06e8893de7a8aa357b991d5",
-    "content-hash": "018e3254777ea41931160ba705244321",
+    "hash": "01ad0b9dd992f516cda15bef5f170952",
+    "content-hash": "391717ced6c13ebc0320576baca87e9d",
     "packages": [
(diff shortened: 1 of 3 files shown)


composer.lock
```diff
@@ -225,7 +225,7 @@
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/dbtlr/php-airbrake/zipball/69fd4782219ba22fcfe4c278de1eebbbe38f0950",
+                "url": "https://api.github.com/repos/dbtlr/php-airbrake/zipball/999763ca1580d1fa7bffc0b6dfe1cf0152078e86",
```
narcoticfresh: @hairmare hm, why on earth does the zip dist change? i just updated the hash.. (`update --lock`).. and at the same time, the `source.reference` stays the same..

hairmare: I think that's an artefact of the github api... Given that errbit has deploy issues anyway we can get away with it. At best we should update php-airbrake soonish anyhow, it recently merged some good stuff.

PR #360, opened 2016-02-05
EVO-2695 Add Choice options

jacobchristensen79:


hairmare:
@narcoticfresh This looks good to me. Maybe you know abou parts that might be affected by the CDATA container introduced in the `value`. I think you did the original work on constraints.

narcoticfresh:
@hairmare @jacobchristensen79 
logically it shouldn't make a difference if we have literals or `CDATA` once the `Container` is compiled..

My question is more, do we have any unit tests covering this? I see that https://github.com/libgraviton/GravitonTestServicesBundle/blob/develop/Resources/definition/ShowCase.json has a `strict` validator, but there's nothing using data that tests the `CDATA` escaping..

hairmare:
Having a test for one of these cases make lots of sense. You might need to add something in [TestServicesBundle](https://github.com/libgraviton/GravitonTestServicesBundle) so you get something testable.

lapistano:
Toby and I started with Jacobs' agreement on writing tests.

tobster67:
I checked in the modified tests in the Jacobs Feature-Branch feature/evo-2695-choices to let him check them out. 
Basically Bastian and I modified the Graviton\CoreBundle\Tests\Controller\ShowcaseControllerTest and the JSON-Files that are used by the several tests.
The new test testErrorsInFixtures() is loading a json containing an illegal fixture for a choices field with a Choice Constraint  and checks if the error message is as expected.
To get the tests working, modify the ShowCase.json in vendor/graviton/test-services-bundle/Resources/definition/ which is part of https://github.com/libgraviton/GravitonTestServicesBundle . 
There is a Pull-Request for this here:
https://github.com/libgraviton/GravitonTestServicesBundle/pull/27

hairmare:
I just tagged GravitonTestServicesBundle so you can composer update it here.

This also seems to be in need of a rebase or merge.

tobster67:
part of the fix of the failed build is the pull request for the bugfix in GravitonTestServicesBundle https://github.com/libgraviton/GravitonTestServicesBundle/pull/28 
The latest Version of GravitonTestServicesBundle has to be referenced in the composer.json. So I am waiting for the GravitonTestServicesBundle pull request to be merged before going on with the fix here.

hairmare:
merged and tagged libgraviton/GravitonTestServicesBundle#28

hairmare:
eec1334 appears to mostly do the same thing as #366. It might be better to merge the "update all the things" in a different PR than the choice changes.

tobster67:
If i run the tests locally, I get one failure that has nothing to do with the choices Test - maybe a change in symphony:

1) Graviton\I18nBundle\Tests\Form\Type\TranslatableTypeTest::testGetParent
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'form'
+'Symfony\Component\Form\Extension\Core\Type\FormType'

...workspace/graviton/src/Graviton/I18nBundle/Tests/Form/Type/TranslatableTypeTest.php:32

I think this should be fixed in https://github.com/libgraviton/graviton/pull/366 "Update all the things", but it will most likely prevent this build from completing. So are we in a deadlock here? @hairmare Should I fix it in this branch to get this done finally?

hairmare:
Yes, the fix is in https://github.com/libgraviton/graviton/commit/530aa45510ce4e6830aedf7c0e787061a96dd993.

It looks like you have deps in `vendor/` that differ from what is specified in `composer.lock`. 

I don't see any deadlock, just run `composer install` before running `phpunit` and everything should test out fine.

tobster67:
OK - nice :-) Then it will work on Travis as well. No need for me to commit again. If that Travis thing would only be faster!

tobster67:
@hairmare: Sorry to be so nasty about that: The Composer cannot update this test because its part of the Graviton/src/i18nBundle:
https://github.com/libgraviton/graviton/blob/feature/evo-2695-choices/src/Graviton/I18nBundle/Tests/Form/Type/TranslatableTypeTest.php. 
This is not in the vendor. So composer install doesn't help here. Merging feature/evo-2695-choices with feature/update-all-the-things as You proposed before would do the job.

hairmare:
I'll merge this as soon as travis is showing some green integration tests and redo #366 afterwards

hairmare:
We might need to add some more tests for this feature later on as it seems to only test for errors right now...

hairmare:
Just FYI: I tagged v0.44.0 with this and did the wrapper updates needed...

PR #365, opened 2016-02-11
EVO-2707 Cookie Strategy update.

jacobchristensen79:


hairmare:
In what cases should this code be triggered? At a glance it looks like this will get triggered if the reverse-proxy sends us garbage with the right amount of formatting. Do we really want this? Shouldn't graviton object to auth headers containing unexpected contents?

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
```diff
@@ -70,4 +102,11 @@ protected function validateField($header, $fieldName)
             }
         }
     }
+
+    /**
+     * @param Container $container
+     */
+    public function setContainer(Container $container) {
```
hairmare: If just getting at the parameters above is the reason for injecting the whole container we should switch to just injecting the needed params.

That will also take care of the `if ($container->hasParameter()) $param = $container->getParameter(); else throw Exception();` spiel that is missing above.

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
```diff
@@ -33,10 +36,39 @@ protected function extractFieldInfo($header, $fieldname)
...
-        throw new \InvalidArgumentException('Provided request information are not valid.');
+        // Cookie value can be a coma separated string
+        $extractField = $this->container->getParameter('graviton.security.core_extract_username');
+
+        // Get Original requested Value.
+        if (!$extractField) {
+            return $bagValue;
+        }
+
+        $pattern = "/((?m)(?<=\b{$extractField}=)[^,]*)/i";
```
lapistano: lines 55 and following are way to specific to our current usecase to be implemented in the AbstractHttpStrategy class. pls move this to the specific strategy.

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
lapistano: +1

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
jacobchristensen79: I wanted this to be usable for any implemented strategy. And lines 51 will only be run if so is enabled in parameters ( as so if we wish to have this for header or cookie ).

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
jacobchristensen79: Has parameter is needed when you do not know if something is set, in this case we do know and it will be set. It's optional to have this enabled or not. I do not wish to throw an exception due to we may not wish to enable this feature for all our customers.

src/Graviton/SecurityBundle/Authentication/Strategies/AbstractHttpStrategy.php
```diff
@@ -19,8 +19,6 @@
  */
 abstract class AbstractHttpStrategy implements StrategyInterface
 {
-
-    protected $strategyMatch;
```
jacobchristensen79: It was an unused value

app/config/parameters.yml.dist
```diff
@@ -7,8 +7,8 @@ parameters:
     graviton.security.authentication.header_required: false
     graviton.security.authentication.test_username: false
     graviton.security.authentication.allow_anonymous: true
-    graviton.security.authentication.strategy: graviton.security.authentication.strategy.header
-    graviton.security.authentication.strategy_key: x-idp-username
+    graviton.security.authentication.strategy: graviton.security.authentication.strategy.cookie
+    graviton.security.authentication.strategy_key: graviton_user
```
jacobchristensen79: We change to have by default the Cookie strategy and default key value for the cookie. The strategy it self control how to extract the username.

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -720,7 +720,7 @@ public function getSecurityUser()
     {
         /** @var PreAuthenticatedToken $token */
         if (($token = $this->tokenStorage->getToken())
-            && ($user = $token->getUser()) instanceof SecurityUser ) {
+            && ($user = $token->getUser()) instanceof UserInterface ) {
             return $user;
         }
```
jacobchristensen79: Making the RestBundle independent. Removing reference to Graviton Security Bundle.

src/Graviton/SecurityBundle/Authentication/Strategies/CookieFieldStrategy.php
```diff
@@ -35,6 +50,44 @@ public function __construct($field)
...
+     * @param string       $idName          save to request attrivute name
+     * @return void
+     */
+    public function setDynamicParameters(RequestStack $requestStack, $extractUsername, $extractCoreId, $idName)
+    {
+        $this->requestStack = $requestStack;
+        $this->extractUsername = $extractUsername;
+        $this->extractCoreId = $extractCoreId;
+        $this->clientIdName = $idName;
     }
```
jacobchristensen79: I set these values in service, no in parameters to avoid problems on deployments. But still somehow configurable.

src/Graviton/SecurityBundle/Authentication/Strategies/CookieFieldStrategy.php
```diff
@@ -35,6 +50,44 @@ public function __construct($field)
...
+    /**
+     * Symfony Container
+     *
+     * @param RequestStack $requestStack    request object
+     * @param string       $extractUsername identifier in posted params
+     * @param string       $extractCoreId   client specific identifier
+     * @param string       $idName          save to request attrivute name
+     * @return void
+     */
+    public function setDynamicParameters(RequestStack $requestStack, $extractUsername, $extractCoreId, $idName)
```
lapistano: Why hiding dependencies in a setter, when it is possible to implement a constructor?

jacobchristensen79: Meant to be back-compatible and optional. But true, it can be moved to constructor and test's updated. I moved the previous functionality implemented in the Abstract class to the Strategy it self and left it as setter.

PR #366, opened 2016-02-11
Update all the things

hairmare:
I needed to change on test since it symfony seems to have introduced some changes.

I did some digging but couldn't find the actual changes. Tests say everything is fine though.

all_the_bump

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ services:
 sudo: false
 before_script:
 - >
-    if [ $(phpenv version-name) != "hhvm" ] && [ $(phpenv version-name) != "7.0" ]; then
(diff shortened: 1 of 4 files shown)


hairmare:
I'm seeing some breakage with php 5.5 and am currently dropping 5.5 support in this branch properly. I think we should consider dropping it now since it will make travis so much faster.

5.6 will have a rather long support term since it is the last one in the 5 series so we might be able to support that in a more LTSy fashion.

Given the large performance improvements we really should consider moving fast on getting php 7 out of the door. Right now I'm even contemplating going live with mongofill as it seems to perform much better than the native driver on 5.6.

hairmare:
meme_face_mad

PR #368, opened 2016-02-17
EVO-2891: Add a ProxyBundle configuration for including / excluding the base path of thirdparty APIs

q-src:
This PR adds a configuration which defines whether graviton should expose thirdparty endpoints with or without the thirdparty API's base path. Since exposing this base path does not make sense in most cases, the default behavior is removing it.

In addition, the following fixes are applied:
1. On `/3rdparty/{prefix}/{endpoint}` requests, use the thirdparty API defined by the prefix in the URL or throw a `NotFoundException` when no such prefix is defined (current behavior is always using the last configured thirdparty API by completely ignoring the prefix in the URL)
2. Path normalization (removing one path `{parameter}` at the end of the path for the endpoint overview) for swagger sources is removed (as discussed with @samsamann). Reason: Removing the `{parameters}` at the end of the exposed URL is confusing, since `{parameters}` in the middle of the path cannot be removed.
3. Always remove the `transfer-encoding` and `trailer` headers when forwarding thirdparty API responses, since graviton does not re-chunk chunked thridparty API responses.
4. Some ENTB phpcbf fixes

### DIFF
--- a/src/Graviton/GeneratorBundle/Tests/Definition/JsonDefinitionArrayTest.php
+++ b/src/Graviton/GeneratorBundle/Tests/Definition/JsonDefinitionArrayTest.php
@@ -8,7 +8,6 @@ namespace Graviton\GeneratorBundle\Tests\Definition;
 use Graviton\GeneratorBundle\Definition\DefinitionElementInterface;
 use Graviton\GeneratorBundle\Definition\JsonDefinitionArray;
 
-
 /**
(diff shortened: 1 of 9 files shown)


lapistano:
unfortunately there was no README.md in the first place. Could you pls add one describing the Bundle, how to configure it and what each config option means. thx

q-src:
@lapistano I've just added a basic readme file.

PR #369, opened 2016-02-17
EVO-2938 - add ref to event status

narcoticfresh:
In talks it has been discussed that it would be nice to enable the user to lookup certain events by the originating REST resource that created a certain `/event/status`.

So this PR adds a `$ref` to every `/event/status/` record. It points to the exact document that was created/updated and allows for more precise RQL searches to find relevant status entries.

I also added more testing so the object that the `Listener` persists gets inspected more in detail.


hairmare:
:+1: make sense, will merge after CI :sleeping:

hairmare:
There seem to be some issues on travis. Could you please check them @narcoticfresh

PR #372, opened 2016-02-19
set additional information field to non-requiered to keep versatility…

phil2211:
The availability must be ensured by the client. If not, the service will loose its generic state

### DIFF
--- a/src/Graviton/FileBundle/Resources/definition/File.json
+++ b/src/Graviton/FileBundle/Resources/definition/File.json
@@ -84,7 +84,7 @@
         "type": "string",
         "title": "property name",
         "description": "additional property name",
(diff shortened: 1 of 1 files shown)


hairmare:
Are you sure these need to be optional? The whole additionalProperties items should be optional but their actual contents should not.... while you need not specify an additionalProperties item, if you specify one it should always contain both a name and value.

I think we should split additionalProperties into its own definition to make this possible (if it doesn't already work, I didn't test).

lapistano:
I agree with @hairmare; by extracting the "metadata.additionalProperties.0.*" definitions into a separate definition file it is possible to handle the whole field as optional, but when added each field of it with a separate level of mandatory.

PR #376, opened 2016-02-23
Update Schema output to print using Serialiser config.

jacobchristensen79:


hairmare:
I have a feeling that this touches way to much to achieve getting rid of `realId` in schemas. I think they are quite ok where they already are and should only be removed from schemas.

lapistano:
it's still work in progress, right?

PR #377, opened 2016-02-23
Hide realId from schema output

hairmare:
This is a rather quick solution. I tried removing the realId earlier, but need to do it so late due so I don't break everywhere else this is used.

I'm also checking that the field is an id since that enables us to still use the name realId for other purposes if we need to.

This replaces #376 with something a bit more surgical but also a bit hacky...


hairmare:
Closing in favor of #376 since that one also contains my changes from here.

PR #378, opened 2016-02-29
'loose up' some deps in composer.json #2

narcoticfresh:
like the other, but with forced debug: false.. just tryin


narcoticfresh:
retest this please

narcoticfresh:
@hairmare damn it.. still the same.. [without coverage](http://jenkins.rel-pub.services.evoja.ch/job/Graviton/5/), checking out, generating and testing takes ~12 minutes(!). [with coverage](http://jenkins.rel-pub.services.evoja.ch/job/Graviton/4/) it's 2 hrs 10 minutes ;-// wtf??

hairmare:
It's already alot better than where we started. At least individual tests take less than an hour on travis now.

I do lack the ability to tell which changes actually made the difference due to them mostly being aborted on travis :/

composer.json
```diff
@@ -45,12 +45,12 @@
         "graviton/php-rql-parser": ">=2.0.0-alpha15",
         "graviton/rql-parser-bundle": "~0.10.0",
         "knplabs/knp-gaufrette-bundle": "^0.2@dev",
-        "aws/aws-sdk-php": "~2.7",
+        "aws/aws-sdk-php": "~3",
```
hairmare: Is this safe? Do we need to test this against actual storage?

narcoticfresh: @hairmare yes, we should test that actually.. but it's not the focus here.. as long as we get a recent `phpunit/phpunit`, it's fine - we can also revert that change if it's not blocking the new phpunit release

narcoticfresh:
@hairmare what makes you think that fixture loading is the part that takes long?

hairmare:
@narcoticfresh Are you still on this?

Fixture loading takes long since the FixturesBundle creates a new kernel for loading each fixture. At a quick glance I didn't find a way to deactivate the debug flag on those kernels.

narcoticfresh:
@hairmare well.. no, I'm not really on it anymore. Too much time was spent, it is hard to justify to spend more time - seeing that I don't really know where I should investigate further to squeeze more time out of time unit test run.

the goal of this PR was to have a decent (<1 hr) runtime of unit tests in *our* Jenkins - but it seems that is not doable at the moment..

what should we do with the stuff here? anything you see that we could merge into develop? if yes, I could rebase and resolve the base conflicts..

narcoticfresh:
i'm going to close this one..  #380 seems to get the useful stuff out of this..

PR #381, opened 2016-03-07
Fast import of initial YML data files

jacobchristensen79:


hairmare:
What does this do that https://github.com/libgraviton/import-export can't do? The tool already has a facility to load data directly into MongoDB. That importing the yaml files is slow is due to the fact that it does so via HTTP. This is an important feature because it ensures that we only add data through the yaml files that can be changed after it is deployed on production...

src/Graviton/CoreBundle/Command/DirectImportCommand.php
```diff
@@ -0,0 +1,399 @@
...
+        // Do this by rest
+        $restObjects = [];
+
+        /** @var SplFileInfo $file */
+        foreach ($finder as $file) {
+            $fileName = str_replace($location, '', $file->getPathname());
+
+            $progress->advance();
+
+            $contents = explode('---', $file->getContents());
```
hairmare: There is `webuni/front-matter` that does frontmatter like a baus.

src/Graviton/CoreBundle/Command/DirectImportCommand.php
```diff
@@ -0,0 +1,399 @@
...
+            if (!$contents || count($contents)!==3) {
+                $errors[$fileName] = 'Content of file is incorrect, could not parse it with --- ';
+                continue;
+            }
+
+            $target = trim(str_replace('target:', '', $contents[1]));
```
hairmare: We might introduce a `file:` attribute in frontmatter for loading to `/file` soonish so this needs to become more general.

narcoticfresh:
@jacobchristensen79 @hairmare 
was there an answer why we do this while we have `graviton-import-export`?
as there is no PR description it's hard to see what the purpose is(?)

hairmare:
@narcoticfresh There is an error when loading initialdata with graviton-import-export in vagrant boxes. I'm currently trying to figure out why the built-in server hangs up after a while and then only responds with 500.

narcoticfresh:
@hairmare  did the vagrant users try to load them synchronously (`-s`)?

hairmare:
@narcoticfresh yes, nevertheless the built in server just breaks down after loading about half of the files ;(

src/Graviton/CoreBundle/Command/DirectImportCommand.php
jacobchristensen79: Agree, but I think File should not be in target but more generic to the item to be inserted ( As we also have items and file may be also files...)

src/Graviton/RestBundle/Validator/Form.php
```diff
@@ -101,6 +103,48 @@ public function checkForm(
...
+                if (!$child->isValid()) {
+                    $errors[$child->getName()] = $this->getErrorMessages($child);
+                }
+            }
+        }
+
+        return $errors;
+    }
+
+    /**
```
jacobchristensen79: Now we get all form errors, not only something went wrong but what.

jacobchristensen79:
I have tested this importing the new initial data. It's working but we have to review the data files for incorrect params.

PR #383, opened 2016-03-09
flexibilize monolog log destination via incenteev

narcoticfresh:
We want to expose the log file destination for our Docker CD infrastructure..
Once tagged, I'll open PRs for the wrappers..

I tested this locally with a branch on a wrapper - setting the parameter to `php://stdout` indeed makes `ERROR`s pop up on the stdout of the `*-backend` docker container using `docker logs`

I guess we leave the loglevel on `error` on `prod` environment?

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -16,7 +16,7 @@ monolog:
     handlers:
         main:
             type:   stream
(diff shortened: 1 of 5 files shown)


hairmare:
LGTM, waiting on CI

PR #384, opened 2016-03-09
update to new php-amqplib & bundle

narcoticfresh:
You may see the following messages from composer:

```
Package videlalvaro/php-amqplib is abandoned, you should avoid using it. Use php-amqplib/php-amqplib instead.
Package oldsound/rabbitmq-bundle is abandoned, you should avoid using it. No replacement was suggested.
```

User `videlalvaro` has moved his projects to the `php-amqplib` organization and other maintainers took over. So we shall change our dependencies accordingly..

This *may* conflict with #383 (`composer.lock` changes) upon merge of either one, see what comes first..

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,14 +49,14 @@
         "eo/airbrake-bundle": "~1.0",
         "hwi/oauth-bundle": "~0.3",
         "graviton/deploy-scripts": ">=0.2.2",
-        "oldsound/rabbitmq-bundle": "~1.7",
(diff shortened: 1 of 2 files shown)


hairmare:
Please redo the lockfile after #383 gets merged, CI is almost through on that one.

narcoticfresh:
conflict resolved..

narcoticfresh:
@jacobchristensen79 you u merge as @hairmare is not here today?

PR #385, opened 2016-03-09
Host swagger through php

hairmare:
This makes it easier to deploy on docker and also removes the silliness that was graviton:swagger:copy.


narcoticfresh:
@jacobchristensen79 and this seems to be fine too, why not merge it?

PR #387, opened 2016-03-14
Validation of Id on ExtRef relation.

jacobchristensen79:
Configurable option to enable the validation on extref $ref references to existing objects.

### DIFF
--- a/src/Graviton/RestBundle/Validator/Constraints/ExtReference/ExtReferenceValidator.php
+++ b/src/Graviton/RestBundle/Validator/Constraints/ExtReference/ExtReferenceValidator.php
@@ -69,7 +69,7 @@ class ExtReferenceValidator extends ConstraintValidator
         }
 
         // Does ID exists in DB.
-        if ($this->validateId && !in_array($value->getRef(),['App', 'TestCaseReadOnly'])) {
+        if ($this->validateId && !in_array($value->getRef(), ['App', 'TestCaseReadOnly'])) {
             $db = $this->documentManager->getConnection()->selectDatabase('db');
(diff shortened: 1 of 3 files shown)


hairmare:
AFAIK there are some cases where we only want the format of the extref to be valid even though it may point to a 404. It looks like that won't work anymore if we validate all the things.

hairmare:
I'm still not sure that we should not allow records pointing to a 404. Also, this certainly needs some tests to cover the functionality as well as check the error message.

src/Graviton/RestBundle/Validator/Constraints/ExtReference/ExtReferenceValidator.php
```diff
@@ -46,6 +65,19 @@ public function validate($value, Constraint $constraint)
             !in_array($value->getRef(), $constraint->collections, true)
         ) {
             $this->context->addViolation($constraint->message, ['%collection%' => $value->getRef()]);
+            return;
         }
+
+        // Does ID exists in DB.
+        if ($this->validateId && !in_array($value->getRef(),['App', 'TestCaseReadOnly'])) {
```
narcoticfresh: wtf? `TestCaseReadOnly`

narcoticfresh:
@jacobchristensen79 sorry for joining the discussion so late ;-)

I can't see if this still enables to save `extref`s that lead to non-existing documents ( @hairmare 's question) . Cause that feature is crucial - just imagine the mess in initialdata import ;-) 

So what does it actually change? Do you intend to set the config flag to `false` anyway all the time and just have it around for debug scenarios?

PR #388, opened 2016-03-14
Feature/search extension

alebon:
Added RqlTranslator to transform SearchNodes into LikeNodes which can be easy used to achieve search functionality by configuring fields as searchable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "php-jsonpointer/php-jsonpointer": "~1.1@rc",
         "exercise/htmlpurifier-bundle": "~0.1",
         "egulias/email-validator": "~1.2",
-        "graviton/php-rql-parser": ">=2.0.0-alpha15",
+        "graviton/php-rql-parser": ">=2.0.0-alpha16",
         "graviton/rql-parser-bundle": "~0.10.0",
         "knplabs/knp-gaufrette-bundle": "^0.2@dev",
         "aws/aws-sdk-php": "~2.7",
--- a/src/Graviton/RestBundle/Service/RqlTranslator.php
+++ b/src/Graviton/RestBundle/Service/RqlTranslator.php
@@ -6,11 +6,15 @@
 namespace Graviton\RestBundle\Service;
 
 use Graviton\Rql\Node\SearchNode;
+use MongoDate;
 use Xiag\Rql\Parser\AbstractNode;
 use Xiag\Rql\Parser\DataType\Glob;
 use Xiag\Rql\Parser\Node\AbstractQueryNode;
 use Xiag\Rql\Parser\Node\Query\LogicOperator\AndNode;
 use Xiag\Rql\Parser\Node\Query\LogicOperator\OrNode;
+use Xiag\Rql\Parser\Node\Query\ScalarOperator\EqNode;
+use Xiag\Rql\Parser\Node\Query\ScalarOperator\GeNode;
+use Xiag\Rql\Parser\Node\Query\ScalarOperator\LeNode;
 use Xiag\Rql\Parser\Node\Query\ScalarOperator\LikeNode;
 use Xiag\Rql\Parser\Query;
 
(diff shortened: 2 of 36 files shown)


lapistano:
Missing documentation in https://github.com/gravity-platform/doc/. Pls make sure the whereabouts are explained there.

composer.json
```diff
@@ -42,7 +42,7 @@
         "php-jsonpointer/php-jsonpointer": "~1.1@rc",
         "exercise/htmlpurifier-bundle": "~0.1",
         "egulias/email-validator": "~1.2",
-        "graviton/php-rql-parser": ">=2.0.0-alpha15",
+        "graviton/php-rql-parser": ">=2.0.0-alpha16",
```
lapistano: what about stabilising the sql-parser? Why is it still in alpha?

hairmare: Because we never took the time to start with stabilizing and v1 was release prematurely... It's probably about time to start doing beta releases though.

hairmare:
On the missing docs... I already started doxing the json files possibilities in this PR: https://github.com/gravity-platform/doc/pull/24

wegenmic:
Whenever I start a search on an object that has no searchable attributes defined, using the search syntax returns all elements of that type instead of no elements.
We should change that behaviour so that we can only have actual results when there are searchable attributes defined.

src/Graviton/RestBundle/Service/RqlTranslator.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        $orNode = new OrNode();
+
+        foreach ($searchFields as $searchField) {
+
+            foreach ($searchNode->getSearchTerms() as $searchTerm) {
+                $searchGlob = new Glob('*' . $searchTerm . '*');
+                $likeNode = new LikeNode($searchField, $searchGlob);
+                $orNode->addQuery($likeNode);
+            }
```
wegenmic: The foreach loop
`            foreach ($searchNode->getSearchTerms() as $searchTerm) {
                # handle normal text fields
                $searchGlob = new Glob('*' . $searchTerm . '*');
                $likeNode = new LikeNode($searchField, $searchGlob);
                $orNode->addQuery($likeNode);`
only handles searches for normal text / string fields. Since we should also be able to search for number and dates, I would suggest something like the following
`            foreach ($searchNode->getSearchTerms() as $searchTerm) {
                # handle normal text fields
                $searchGlob = new Glob('*' . $searchTerm . '*');
                $likeNode = new LikeNode($searchField, $searchGlob);
                $orNode->addQuery($likeNode);

                # handle numbers
                $searchNumber = (int)$searchTerm;
                $numberNode = new EqNode($searchField, $searchNumber);
                $orNode->addQuery($numberNode);

                # handle dates
                $dateNode = new AndNode();
                $searchFrom = new MongoDate(strtotime($searchTerm."T00:00:00.000Z"));
                $searchTo = new MongoDate(strtotime($searchTerm."T23:59:59.999Z"));
                $dateFrom = new GeNode($searchField, $searchFrom);
                $dateTo = new LeNode($searchField, $searchTo);

                $dateNode->addQuery($dateFrom);
                $dateNode->addQuery($dateTo);
                $orNode->addQuery($dateNode);
            }`

src/Graviton/RestBundle/Service/RqlTranslator.php
```diff
@@ -43,6 +47,29 @@ public function translateSearchNode(SearchNode $searchNode, $searchFields = arra
                 $searchGlob = new Glob('*' . $searchTerm . '*');
                 $likeNode = new LikeNode($searchField, $searchGlob);
                 $orNode->addQuery($likeNode);
+
+                if (is_numeric($searchTerm)) {
```
hairmare: Can we still search for a string containing only numbers with this? This might be a problem due to the fact that `is_numeric("1234") === true`.

alebon: Yes, it will match. I think the important part here is the EqNode which will be correctly transformed by the MongoOdm visitor (QueryBuilder). And if it is only a part of the search term, let's say the DB value is "myNameis233" and you search for "233" the first LikeNode translation will match this case.

PR #389, opened 2016-03-15
ALIDA auth header

flavioaiello:
Due to the ALIDA auth a new header is needed. This will be later replaced with oauth2.

### DIFF
--- a/src/Graviton/RestBundle/Resources/config/services.xml
+++ b/src/Graviton/RestBundle/Resources/config/services.xml
@@ -218,6 +218,10 @@
             <call method="addAllowHeader">
                 <argument>X-REST-Token</argument>
             </call>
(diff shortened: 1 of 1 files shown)


lapistano:
We should discuss, if it is possible to remove X-REST-Token then.

flavioaiello:
The discussion is ongoing, so this is just a workaround

hairmare:
WTF do we need another one of these weird non standardized auth headers? Isn't `X-REST-Token` already enough? Who felt the need to invent another one of these crappy things? It would be much better if someone finally gave us the time and resources to do a proper OAuth2/JWT implementation...

wuschelchopf:
@hairmare it was me :D We found out today, that it's required (at least until GRV gets time for the OAuth2-Auth) that BECO sends the auth-Token in client-specific-header-field-name. The fastest fix was that graviton accepts all known variations.

damien

PR #390, opened 2016-03-21
Made fetch of external Sources more robust

lapistano:


src/Graviton/SecurityBundle/Tests/Authentication/Strategies/CookieFieldStrategyTest.php
```diff
@@ -50,7 +50,7 @@ public function setUp()
      *
      * @return void
      */
-    public function testApply($fieldValue)
+    public function dtestApply($fieldValue)
```
hairmare: This hunk does not look like it was changed on purpose.

PR #394, opened 2016-03-25
Added support for OPTIONS http method

lapistano:
necessary for the BECO to preflight the whoami service


hairmare:
This should be done as its own action that does not also return a body to keep preflight requests fast and minimal. Have a look at [how it's done](https://github.com/libgraviton/graviton/blob/develop/src/Graviton/CoreBundle/Resources/config/routing.xml) for `/core/version` in  to see how it could be implemented.

lapistano:
obsolete by #395

PR #400, opened 2016-04-05
Feature/evo 4532 different loc 4 consultant

lapistano:
the consultant username was not stored in the username, but in the finnova_id field

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ app/config/parameters_local.xml
 app/files/
 app/config/deploy.yml
 versions.yml
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ app/config/parameters_local.xml
 app/files/
 app/config/deploy.yml
 versions.yml
-.idea/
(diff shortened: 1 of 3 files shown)


wegenmic:
looks good to me

.gitignore
```diff
@@ -16,3 +16,4 @@ app/config/parameters_local.xml
 app/files/
 app/config/deploy.yml
 versions.yml
+.idea/
```
hairmare: Please add this line to your local `~/.gitignore` and not to the project itself...

src/Graviton/SecurityBundle/Tests/Authentication/Strategies/CookieFieldStrategyTest.php
```diff
@@ -131,9 +131,9 @@ public function testApplyExtract($username, $fieldValue)
     public function stringExtractProvider()
     {
         return array(
-            'testing extract username' => array("testUser", "username=testUser;finnova_id=someId123"),
-            'testing extract rev username' => array("someOtherUser", "finnova_id=someId123;username=someOtherUser"),
-            'trailing simicolon test' => array("test-mdm", "finnova_id=someId123;username=test-mdm;"),
+            'testing extract username' => array("someId123", "username=testUser;finnova_id=someId123"),
+            'testing extract rev username' => array("someId123", "finnova_id=someId123;username=someOtherUser"),
+            'trailing simicolon test' => array("someId123", "finnova_id=someId123;username=test-mdm;"),
```
hairmare: These test were worth much more before the change. I think we should have changed `$fieldValue` to make these tests so what we want and still be expressive after the change.

PR #401, opened 2016-04-05
Added extensive logging to actually know which references are wrong

lapistano:


src/Graviton/RestBundle/Service/RestUtils.php
```diff
@@ -101,11 +108,24 @@ public function getServiceRoutingMap()
      */
     public function serializeContent($content, $format = 'json')
     {
-        $result = $this->getSerializer()->serialize(
-            $content,
-            $format,
-            $this->getSerializerContext()
-        );
+        $result = '';
```
hairmare: Won't this lead to data missing without much feedback to the client? Personally I would prefer if we return the client an actual error instead of just replacing the deserialized stuff with `""` as it fails.

lapistano: we thought aber this as well, but returning another structure than the expected on will cause the client to break .. and this is not an option atm ..
Further the chance to have the info in the log (and I am monitoring it) gives us the chance to actually do s.th about circular references in the structures and see that those get resolved.

PR #408, opened 2016-04-15
http 200 causes cors preflights to cache content

lapistano:
changed the returned http status for options requests to 204 to preve…nt browser from caching cors preflights


hairmare:
Preventing CORS request is usually considered ok since the are not expected to change dynamically all the time. What gives?

PR #413, opened 2016-04-29
add new 'ignored' status to /event/status worker status

narcoticfresh:
for EVO-5106..

adds a new possible value for `status.status` in `/event/status`. The new `ignored` status shall be set by a worker if he decides to not do anything on a given work item, as opposed to just set it to `done`. The status `done` without having actually done anything is misleading, we need to be able to distinguish that case.


narcoticfresh:
@hairmare ping.. did this travis error came up before? seems to happen on php7 mongofill install - and has nothing to do with the change this PR introduces

PR #415, opened 2016-05-02
EVO-5038 Adding dynamic id to CodeEmbedded Documents

jacobchristensen79:


src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentEmbedded.php.twig
```diff
@@ -31,7 +31,15 @@ protected $realId;
 */
 public function getId()
 {
-return $this->id;
+{% if document == 'Code' %}
+    if (!$this->id && isset($this->group, $this->number)) {
+    return $this->group.'-'.$this->number;
```
hairmare: Shouldn't the id field contain always contain a concat of group and number?

jacobchristensen79:
This should be solved by the DB load data and this should not be needed. Back-up plan B

PR #393, opened 2016-03-23
Update all the things with composer

hairmare:
* [x] update all the things to their most current state
* [x] fix resulting issues (small stuff with phpcs and phpunit mock behaviour)
* [x] update dependencies in composer.json to make dep graph smaller and allow faster updating in the future as well as during travis runs

PR #416, opened 2016-05-09
form validation: print "extra fields" names

narcoticfresh:
One of the most annoying client-facing Graviton errors is _This form should not contain extra fields_, as it's not clear which fields are those extra fields.

I revisited this, digged into `Form` again and it seems that there's an easy way to determine those fields (we tried this some time ago and it wasn't that easy). Now, those extra fields are added to the `cause` object.

It seems that this Symfony feature is intended for the Twig integration as those params have Twig placeholder names. So there's a need to change those names.

_Why is it done like that?_
As opposed to "normal" `Validators`, this error is part of the generic `ConstraintViolation` class in Symfony. The only way to override/customize the error message is upon creating the `Form` instance, but also this wouldn't allow to incorporate the parameter value IMHO.

I made it generic in a way that any `parameter` array member can be added to the error message, but for now we only want to see `extra_fields`.


narcoticfresh:
@hairmare umh, there is the memory_limit error again on PHP7.. I'll try again after #393 - should that make things better I hope?

hairmare:
I tried combating the memory issues in d58573c87bd19957196b49edd0169b3e5b731eca and it seems to have worked. That still needs some downstream testing though and I'm not sure when we'll be able to land it since noone seems to want to merge it.

hairmare:
I'm preparing a simpler replacement...

PR #423, opened 2016-05-24
Set index as Single Field Indexes

jacobchristensen79:
In generator, on building XML definition for MongoDB use single field indexes


hairmare:
I feel we should write a `<indexes>` tag and then add multiple indexes to it. That will keep the xml much less complex and also follows what seems to be newer symfony best-practices.

PR #425, opened 2016-05-26
Fast return to avoid null values if object contain same reference to …

jacobchristensen79:
…translation id

### DIFF
--- a/src/Graviton/I18nBundle/Listener/I18nSerializationListener.php
+++ b/src/Graviton/I18nBundle/Listener/I18nSerializationListener.php
@@ -74,6 +74,10 @@ class I18nSerializationListener
         }
 
         $hash = \spl_object_hash($object);
+        // If main object already build return.
+        if (array_key_exists($hash, $this->localizedFields)) {
+            return;
+        }
         $this->localizedFields[$hash] = [];
 
         $translatable = $object->getTranslatableFields();


src/Graviton/I18nBundle/Listener/I18nSerializationListener.php
```diff
@@ -74,6 +74,10 @@ public function onPreSerialize(PreSerializeEvent $event)
         }
 
         $hash = \spl_object_hash($object);
+        // If main object already build return.
+        if (array_key_exists($hash, $this->localizedFields)) {
+            return;
```
hairmare: If we enter here our data is in really bad shape. For some reason we have data that gets us into here. We should not have any data that gets us into this loop since using referenced data in the way that we are seeing here is a very bad pattern and leads to heaps of data loss in the long run.

jacobchristensen79: We do not enter data here. We do build here an Array of fields for a Type of Object that have translated fields. On post Serialise we insert data. 
So at this point of the code we do not to do a fetch get translatable fields for each object if already have the translatable fields for that one, so just return and no need to build again the array.

hairmare:
We are playing a dangerous game here! We really need to figure out how this new `return` statement can never be reached because introducing inconsistent data is really easy as the underlying problem leading to this pr shows.

PR #426, opened 2016-05-27
Generate all languages from DB into Translations

jacobchristensen79:


src/Graviton/I18nBundle/Command/CreateTranslationResourcesCommand.php
```diff
@@ -97,12 +111,46 @@ function ($language) use ($output, $domains) {
...
                     $domains,
                     function ($domain) use ($output, $language) {
+                        /** @var Language $language */
                         $file = implode('.', [$domain, $language->getId(), 'odm']);
                         $this->filesystem->touch(implode(DIRECTORY_SEPARATOR, [$this->resourceDir, $file]));
                         $output->writeln("<info>Generated file $file</info>");
+
+                        $locale = $language->getId();
+                        $count = $this->generateResourceTranslations($domain, $locale);
+                        $output->writeln("<info>Generated {$count} translations for {$domain}:{$locale}</info>");
```
hairmare: s/Generated/Loaded/

I find it rather weird that symfony doesn't load all of this on it's own after we supply translation resource files. Maybe something changed in the i18n component, maybe cache scrubbing during the deploy at the right time would also solve this without implementing it ourselves.

jacobchristensen79: According to Symfony, http://symfony.com/blog/new-in-symfony-2-6-improvements-for-the-translation-component
We could do a --force but this is not working using our current wrapper installation. If you have a better approach please provide it as this is currently a blocker and Critical needed solution.

hairmare:
Cache scrubbing seems to have worked...

PR #427, opened 2016-06-03
EVO-5807 - dont render empty extref objects at all

narcoticfresh:
**Important**: @tobster67 this is *your* PR now, please see it gets merged ;-)

it has been found that some clients have problems when we render an empty extref object (*an object having only `$ref` as attribute, but is optional and the client didn't send it*) as empty object. this PR shows a way to change specifically that - **but leaving current test expectations untouched**.

as I'm not sure if this needed/wished and the client impact (to others) is unknown to me, I started it to show how it could be done - so please take this over.

so this PR changes (given an empty property: `app.module.$ref`)
* what has been rendered as `"app": {"module": {}, "secondField": 33}` 
* will be rendered `"app": {"secondField": 33}` 
so if `module` is an object to only have `$ref` and it's empty, it will not be rendered at


narcoticfresh:
This one is not needed it seems, closing

src/Graviton/CoreBundle/Tests/Controller/NullExtrefControllerTest.php
```diff
@@ -208,7 +224,38 @@ public function dataTestData()
                             ],
                         ],
                     ],
-                ]
+                ],
+                [
+                    'requiredExtref'      => ['$ref' => 'http://localhost/core/app/admin'],
+                    'optionalExtrefArray' => [],
```
narcoticfresh: @tobster67 so IMHO, the tests `testPutMethod` and `testPostMethod` here in `NullExtrefControllerTest` already assert the new behavior..

they insert `null` `extref`s into a service, fetch them again and compare. now, they also check that the empty objects are not rendered anymore..

PR #430, opened 2016-06-11
EVO-4951 - Massively improve PUT/POST/PATCH performance by validation change and optimizations

narcoticfresh:
When we started to document the `/consultation` structure, we quickly seen a big problem in Graviton: The complexity and time needed to validate a service structure grows exponentially to its size. So with our "Big Bang 1" changes we wanted (we scaled it back then), Graviton took around ~30 seconds to insert a 1.2MB consultation on a PUT request.

Even today, a big sized consultation needs around 6-7 seconds to PUT on docker instances.

**What does change**
This PR aims to make this way quicker (**from 6-7 seconds to ~0.9 seconds** when all cached) by
* Completely changing the way Graviton validates user input. We ditch the Symfony Form approach and validate the input based on the JSON Schema (v4) using the validator. This approach is quick and allows us to *completely cache a schema*. The schema does not need to be changed *depending on user input*, which is a major pain in the current approach
* Optimize several needless actions in the current workflow (like needless Doctrine hydrations) 
* Try to cache as much as possible (single request scope) 

**Impact**
There *should* be no changes from the user perspective in validations (some will be noted here). The main difference will be changed validation messages (the ones sent with `400 Bad Request`):
* All validation texts (like "Cannot be blank" or whatever) will change
* The `extra fields are not allowed` will produce more errors, as 1 error per extra field is generated
* Missing properties in sub-sub-structures only produce one warning for the parent, not for all children, so produces less errors.
* Also, our `schema` output will change. It will be complete v4 compliant - but maybe some clients will need to be adapted (I'm looking at you, Graviphoton).

This is just semantics and the impact small - we are not aware of any clients actively parsing the return values on `400 Bad Request`, so we can freely change them.

*But of course*, we need to adapt our test suite.

**Testing**
We have very good integration tests testing all needed validation behavior. Already the tests have provided insights into the missing things in this PR, that's why they are still failing. 

For example, the `strict boolean` tests already confirmed that our boolean handling is good from the get go or the `optional embed` test correctly failed showing that we are currently missing this.

So we will mostly add tests for the newly integrated constraints.

It is expected that once this is done and the Symfony Form based code can be removed from Graviton, that overall code coverage may rise.

**TODO List**
- [x] Make sure validation is the same (extref :white_check_mark:, Constraints :white_check_mark:, parent/child relations :white_check_mark:, additionalFields :white_check_mark:)
- [X] "NotBlank" constraint! See PersonContact in PersonBundle
- [x] Make sure schema is cached in a working way (invalidated by Language change! What about x-dynamic-key stuff?) **done** - on `Language` modification, all schemas containing translatables are invalidated; `x-dynamic-key` also the specific schema is invalidated that uses that collection. This behavior is tested by an integration test.
- [x] Eliminate senseless hydrations in *Actions in RestController
- [x] Make sure Translatables get only persisted once per request, in an unique way
- [x] ~~Make sure x-dynamic-key specs in Swagger.json is rendered as $ref to the generated schema (online one)~~ Fixed with flexible rendering by wrapping in a new Entity.
- [x] make tests for all new 'hardened' constraints
- [x] Clean everything that has to do with old Form approach
- [X] See where SchemaModel is still used and how **Update**: It is, `DocumentModel` extends from it. Nice.
- [x] if 2 fields are defining an inline object, like 'position.x' and 'position.y' and both are required; it doesn't say that the property 'position' is required when it is omitted -> tests are rightfully failing because of this! -> pending @mrix 
- [x] Fix different behavior from `develop` - for example `realizationCompletion` has 2 required properties, but when it is sent as `{}` (empty object); current `develop` accepts its whereas we deny it
- [x] Make sure all checks Form->checkJsonRequest() & checkJsonPatchRequest() are moved to another class to be done there
- [x] Make sure `null` can be sent instead of a string if not required - currently gives `NULL value found, but a string is required` -> every non-required primitive shall have `type: ["[its type", "null"]` in schema
- [x] Current test assume that if an object that references another gets deleted, the referenced object also gets deleted.. Isn't that something we didn't actually wanted but were forced to do it? Current behavior (this PR) would let the referenced entity alone.. See  `Graviton\CoreBundle\Tests\Controller\EmbeddingDocumentsTest::testDocumentAsDeepReference
Failed asserting that 200 matches expected 404.` **Update:** Decision together with @hairmare and @phil2211 - objects referenced in a embedded rendered "ref" relation **shall not be deleted**, change test accordingly. (make it also to insert something afterwards again with same ID to see if that works as expected)
- [x] change all test expectations in regard to validation text output (expect the correct error output as it changed)
- [x] Fix `/file` to work the same again
- [x] Update `FileController` and any others that inherit from `RestController` and are dependant of constructor arguments and/or validation after Form stuff is cleaned out
- [x] Make sure whatever the `NotNull` constraint did that is used in various service definitions still does what it's supposed to do or is not needed anymore (maybe @mrix knows more?) => Still does, is used in the `DocumentMap` classes in compiler passes and leads to `required: true`, that is still correct
- [x] Make sure that `IdInPost` validator is applied again correctly - also check not-generated services (like `/core/app` and so on)
- [x] Make sure the `readOnly: true` attribute on properties has the same effect again (see usage in `FileController` or `EventStatus`)
- [x] Move validation logic from PUT/POST/PATCH to an own function taking `$document` and `$model`, returning `$record` and throwing the validation exception if necessary..
- [X] **Important**: Check impact on Graviphoton and fix it before merging, coordinate deploy of backend and Graviphoton to clients. **Update** I checked graviphoton, it doesn't seem to break anything. I thought the datepicker widget was missing, but it seems there never was one. New schema even adds features in `enum` case, so it displays a dropdown with possible values now.
- [x] Make a last profile of a consultation PUT and check for low hanging fruits

client changes
- [ ] Values for numeric fields *must* be sent as numeric value. Strings are not accepted anymore (this has been validated and confirmed to not be a problem for Consultation service)
- [ ] For some reason, the property "$ref" (in extref objects) in PATCH operations was assumed to be named "ref" (no dollar sign). Now, in PATCH, the field must named "$ref".


narcoticfresh:
@mrix I need your help/input on this PR ;-)
What you rightfully fixed at the time in #224 (embedded is optional by default) needs to be addressed again.

Your fix in #224 was only necessary in the way you did it as Symfony Form couldn't handle required properties of not-required objects correctly (so you needed to remove the fields at runtime).

**This has changed** in the new approach - a correctly generated JSON schema (`required` arrays) does correctly validate! But currently I'm unable to change that as the `JSONDefinition` classes there make it rather hard for me.

I want to integrate the following rule:
* An embedded (like a field `object.field`) is per default required
* *Unless* there is a field `object`of type `hash` with `required` unset or `false` -> Then `object` becomes optional

I have big troubles to find out where I can put that logic. The important field is of course here -> https://github.com/libgraviton/graviton/blob/develop/src/Graviton/GeneratorBundle/Definition/JsonDefinitionHash.php#L73.

But I have troubles finding out how to adapt [JsonDefinition::createFieldHierarchyRecursive](https://github.com/libgraviton/graviton/blob/develop/src/Graviton/GeneratorBundle/Definition/JsonDefinition.php#L292) resp. [JsonDefinition::processFieldHierarchyRecursive](https://github.com/libgraviton/graviton/blob/develop/src/Graviton/GeneratorBundle/Definition/JsonDefinition.php#L314) - as in `JsonDefinitionHash` I'm already too late as I lost the definition at least in the case of a `object.field` hash - not in a direct `class:` type `name: object`, there the required is transported correctly it seems.

Can u help us out here?

src/Graviton/I18nBundle/Listener/I18nSerializationListener.php
```diff
@@ -55,6 +73,7 @@ public function onPreSerialize(PreSerializeEvent $event)
         $object = $event->getObject();
 
         // Doctrine try to map value fields that may not exists.
+        // @TODO why is this done here? clearify and maybe remove(?)
```
jacobchristensen79: This was done to have in one place the error cause when doctrine tried to map the embedded object and this did not exists. We talked about this, it caused a 500 exception and we did never no what exactly was going on. Now, currently, we can see what "id" is missing and therefor find a solution fast. Mainly because we use mongodb as a relational db in many cases and allow us to save none related objects using extref. So this solved 500 blank page errors with no log.

narcoticfresh:
@hairmare @narcoticfresh
for us to remember as a review rule for this PR: any changes on unit test asserts in regard to validation output is only permitted in this cases:

* **text change** - the output (text of the validation) has changed
* **removed** - if asserts on validation text was **removed** in the PR -> it's only acceptable if the removed asserts was output that referenced subobjects. so if the test sends data to a service that has `object.field` (and `object` and `object.field` are required) - but the client does **omit** `object`, new validation will only complain about the missing `object`! the missing `field` is only validated when `object` is sent! -> so the assert for `field` **can be removed** in that case
* **added** - there are cases when more validation output is sent now.. as of now, i can't find a case sadly..

src/Graviton/I18nBundle/Listener/I18nSerializationListener.php
hairmare: FWIW this was implemented in #404 and I didn't understand why at the time. Too me it looks like we are ignoring an edge case that needs handling.

Maybe @lapistano knows why this was merged as he clicked the button at the time and I think I remember that this happened under the duress of fixing bugs that were pebkac for the most part.

narcoticfresh:
@hairmare 
so i just removed the WIP label - a big moment ;-)

yesterday, I verified this branch in all 3 wrappers. I loaded the tenant specific initialdata and the example consultations in the `consultation_documentation_init` branch of the consultation bundle.

that resulted in 2 PRs there: 1 in initialdata (some fields where sent as `int` to `string` fields whereas now we need `string` for `string` fields, and 1) in the `swissquant-bundle` for the `RestController` constructor signature.

After all this time, feature-wise, I feel pretty confident about this branch (especially after seeing the branch in the wrappers yesterday).. *but* I'm not quite sure how to proceed as there still could be some impact we don't know about.

So I'm not quite sure when to merge - how do the changes look to you?

src/Graviton/RestBundle/Tests/Utils/ObjectSlicerTest.php
```diff
@@ -21,7 +21,7 @@ class ObjectSlicerTest extends \PHPUnit_Framework_TestCase
      */
     public function testError()
     {
-        $this->setExpectedException('InvalidArgumentException', 'Value must be an array or an object');
+        $this->expectException('InvalidArgumentException', 'Value must be an array or an object');
```
hairmare: This has too many args & needs an additional call to `expectExceptionMessage` to conform to the new syntax.

src/Graviton/SchemaBundle/Constraint/Builder/ChoiceConstraintBuilder.php
```diff
@@ -0,0 +1,66 @@
...
+            $options,
+            function ($carry, $option) {
+                if ($option->name == 'choices') {
+                    return explode('|', $option->value);
+                }
+            }
+        );
+
+        // is this a numeric field? convert values
+        if (in_array('integer', $property->getType()->getTypes())) {
```
hairmare: Scrutinizer sez "The method `getTypes` cannot be called on `$property->getType()` (of type `string`).". On a quick glance I actually concurr. There are more occurrences of this later on... Do you see whats going on?

src/Graviton/SchemaBundle/Document/Schema.php
```diff
@@ -166,43 +237,67 @@ public function getDescription()
...
+            } else {
+                $typesToSet[] = $type;
             }
-            $this->type = $setType;
-            $this->setFormat($type);
-        } else {
-            $this->type = $type;
         }
+
+        $this->type = new SchemaType($typesToSet);
```
hairmare: Scrutinizer is complaining about the type properties `@var` annotation but I can't figure out how to comment on line 30 of this file that needs the actual change.

hairmare:
Did you see that `Graviton\DocumentBundle\DependencyInjection\Compiler\Utils\AbstractField` still uses `Symfony\Component\Form\FormConfigBuilder` in some capacity? It looks like the DocumentBundles DI still has some formy stuff going on. Is that on purpose?

narcoticfresh:
@hairmare i'll have a look at the Scrutinizer report and do some fixes where possible in a timely fashion..

hairmare:
@narcoticfresh don't waste too much time on it, it looks like most of it is just small docs stuff.

src/Graviton/SchemaBundle/Document/Schema.php
narcoticfresh: @hairmare i think this and your previous comment have the same root cause, tried to address in a747b8a

narcoticfresh:
@hairmare 

> Did you see that Graviton\DocumentBundle\DependencyInjection\Compiler\Utils\AbstractField still uses Symfony\Component\Form\FormConfigBuilder in some capacity?

good catch, removed that.. wasn't used anymore..

after checking i agree, most scrutinizer stuff is not worth the effort.. your previous points were corrected in a747b8a

hairmare:
The following 3 files also still seem to `use` form stuff:

* `src/Graviton//ExceptionBundle/Exception/ValidationException.php`
* `src/Graviton//ExceptionBundle/Listener/RqlSyntaxErrorListener.php`
* `src/Graviton//ExceptionBundle/Listener/ValidationExceptionListener.php`

It doesn't look like they are doing anything formy though.

hairmare:
I think this is good to merge. We'll certainly want to do a develop merge before the actual merge to develop and we also need to coordinate when we release this so it gets tested also taking into account that we should get this merged ASAP since this has impact on quite some stuff we've been doing.

narcoticfresh:
@hairmare 

> I think this is good to merge.

very nice, thanks for the review..

i'll keep this up to date with develop if necessary.. right now it looks like we could merge this next week (maybe Tuesday/Wednesday), i'll keep you posted..

src/Graviton/DocumentBundle/DependencyInjection/Compiler/Utils/DocumentMap.php
```diff
@@ -403,4 +456,82 @@ function (\DOMElement $element) {
...
+        $exposedPrefix = '',
+        callable $callback = null
+    ) {
+        $result = [];
+        foreach ($document->getFields() as $field) {
+            if ($this->getFlatFieldCheckCallback($field, $callback)) {
+                $result[$documentPrefix . $field->getFieldName()] = $exposedPrefix . $field->getExposedName();
+            }
+
+            if ($field instanceof ArrayField) {
```
lapistano: don't like the if/ifelse/else tree .. what about injecting a FlattenFieldName**FIELDTYPE**Strategy if it is not possible to make the $field aware of its flattened name?

lapistano: just found out, that **processDocument** does create the fields used here. What about extending the used field classes and  add 'flattenedFieldname' awareness to them (like ContainerAwareInterface) and remove the if/elseif/else tree in favor of a $field->getFlatName()  ?

src/Graviton/GeneratorBundle/Definition/JsonDefinitionHash.php
```diff
@@ -96,6 +102,39 @@ public function getType()
...
+    {
+        return ($this->definition === null);
+    }
+
+    /**
+     * in an 'anonymous' hash situation, we will check if any children are required
+     *
+     * @return bool if required or not
+     */
+    public function getRequired()
```
lapistano: renaming proposal : isRequired()

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -243,47 +255,131 @@ public function find($documentId)
     /**
      * {@inheritDoc}
      *
-     * @param string $documentId id of entity to update
-     * @param Object $entity     new entity
+     * @param string $documentId   id of entity to update
+     * @param Object $entity       new entity
+     * @param bool   $returnEntity true to return entity
      *
      * @return Object
```
lapistano: * @return Object|null

src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -243,47 +255,131 @@ public function find($documentId)
...
+
+        $entity = $this->manager->merge($entity);
+
+        $this->manager->persist($entity);
+        $this->manager->flush($entity);
+
+        if ($returnEntity) {
+            return $entity;
+        }
     }
```
lapistano: + return null;

src/Graviton/SchemaBundle/Document/SchemaAdditionalProperties.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * sets properties
+     *
+     * @param Schema|boolean $properties properties
+     *
+     * @return void
+     */
+    public function setProperties($properties)
```
lapistano: I would understand adding a function **addProperty()** to add a new property, but **setProperties()** does not make sense since the class **SchemaAdditionalProperties** is just a value object anyways. Just create a new one, when the whole set of registered properties change.

src/Graviton/SchemaBundle/Document/SchemaAdditionalProperties.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Graviton SchemaAdditionalProperties Document
+ */
+
+namespace Graviton\SchemaBundle\Document;
+
+use \Doctrine\Common\Collections\ArrayCollection;
```
lapistano: where is the ArrayCollection used?

src/Graviton/SchemaBundle/Document/SchemaEnum.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * sets properties
+     *
+     * @param array $values enum values
+     *
+     * @return void
+     */
+    public function setValues(array $values)
```
lapistano: same as with ** SchemaAdditionalProperties**. Instances of this class are value objects and imo shall not be changed, but newly created.

src/Graviton/SchemaBundle/Document/SchemaType.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * sets types
+     *
+     * @param array $types types
+     *
+     * @return void
+     */
+    public function setTypes(array $types)
```
lapistano: same as with ** SchemaAdditionalProperties**

src/Graviton/DocumentBundle/DependencyInjection/Compiler/Utils/DocumentMap.php
narcoticfresh: @lapistano the Field itself is never aware of it's position in the tree after it has been processed here - that what makes this part so hard. I would really not touch all that stuff more..

narcoticfresh: @lapistano oh i forgot.. all this part (`getFieldNamesFlat()`) was not created in this PR - I moved this around to make it easier accessible by other CompilerPasses - see this commit: https://github.com/libgraviton/graviton/commit/b122f60e2833afb793453c4ebe6875b9522513df#diff-804310df2d1756036e9532d0d084769f

lapistano: commit looks good .. thx

narcoticfresh:
@lapistano i addressed your feedback in 06a5868

Regarding the Documents in `Graviton\SchemaBundle\Document` (like `SchemaAdditionalProperties`, `SchemaEnum`): I needed to create those to have Serializer Handlers on them (which are in `Graviton\SchemaBundle\Serializer\Handler`) as those need special treatment on serializing.

The only reason for those getters is also the Serializer, to have both the possibility to serialize or deserialize them. Of course, we never actually deserialize a Schema (and I'm not really sure if it would work anyway); but it's just there to comply to the normal way we write the Serializer xml's.. 

IMHO they can stay and don't hurt - `add*()` functions are not really needed as nobody uses those Documents in a way that would benefit from "adders"..

PR #431, opened 2016-06-13
File hash based on File content

jacobchristensen79:


hairmare:
I find it very alarming that you are implementing new features in GravitonFileBundle without any additional tests in `Graviton\FileBundle\Tests\Controller\FileControllerTest`.

jacobchristensen79:
Like the rest, added now also the metadata check, both in manager and controller.
If I modify a functionality I first check that the previous tests still work as expected and then I add the test for the new code. There is indeed 2 ways of doing so, adding the tests of the expected result and then develop this is always the best option, here the tests was there and I added new functionality and I wished to have the same result.

PR #433, opened 2016-06-15
Update phpunit and sensio/generator-bundle

hairmare:
The phpunit was split out of @narcoticfresh's work in #430 to make that more lean and I'll take care of merging it there when we get this to develop. I realized that the time is ready to bump sensio/generator-bundle while I was looking into this and was able to bodge our code to conform to what the new bundle gives us. I haven't been able to get https://github.com/sensiolabs/SensioGeneratorBundle/pull/497 merged which would have also taken care of the remaining warnings left after this hack.

To test this in a graviton based wrapper you will need to update this in a fashion similar to the incantation `composer update graviton/* phpunit/* sensio/*` so it does not pull unwanted changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "hwi/oauth-bundle": "~0.4",
         "graviton/deploy-scripts": "~0.4",
         "graviton/json-schema": "~0.8",
-        "hadesarchitect/json-schema-bundle": "~0.1.0",
+        "hadesarchitect/json-schema-bundle": "=0.1.0",
         "php-jsonpatch/php-jsonpatch": "~1.2.2",
         "antimattr/mongodb-migrations": "^1.0",
         "jenssegers/proxy": ">=3.0.0-beta2",
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "symfony/monolog-bundle": "~2.4",
         "sensio/distribution-bundle": "~3.0.12",
         "sensio/framework-extra-bundle": "~3.0",
-        "sensio/generator-bundle": "~2.3",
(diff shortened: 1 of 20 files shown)


jacobchristensen79:
Hi, after install I had to change a few injections of service: json_schema.validator
 It worked after changing it to: hades.json_schema.validator ( changed in version 0.1.0 to 0.1.1 )

hairmare:
I agree, when I do a `composer update hadesarchitect/json-schema-bundle` I need to apply those changes. This PR does not try to update `hadesarchitect/json-schema-bundle` though.
I was not able to reproduce where this is trying to pull the updated bundle. Could you retry after wiping your `vendor/` and `app/cache` dirs and ensuring you do not have local changes in `composer.lock`?

narcoticfresh:
@jacobchristensen79 @hairmare 

yes, `hadesarchitect/json-schema-bundle` is not part of this PR update and it's bump is not in the composer.lock in this PR...

**it should not be bumped** - the new tag has more changes (not only the service identifier) and we will remove the package completely in the near future anyway.. and on top, the new bundle tag doesn't add any value anyway if bumped ;-)

narcoticfresh:
@jacobchristensen79 i updated the dep on the hades bundle.. now it should not be updated to the faulty/changed version in the wrapper anymore..

hairmare:
FWIW I would have *fixed* this in the same way @narcoticfresh did in 57029a443264638db6d8587e9e3ccf520171be63 but was not able to even reproduce the issue. I'm assuming it was raised due to running `composer update` instead of `composer update graviton/* phpunit/* sensio/*` as needed by this pr. Sorry for not doxing that.

PR #434, opened 2016-06-22
Feature/evo 5985 real upload change

jacobchristensen79:
Second option to story

### DIFF
--- a/src/Graviton/FileBundle/Resources/definition/File.json
+++ b/src/Graviton/FileBundle/Resources/definition/File.json
@@ -80,6 +80,12 @@
         "description": "Additional information for this file."
       },
       {
+        "name": "metadata.hash",
+        "type": "string",
+        "title": "Unique file content hash value",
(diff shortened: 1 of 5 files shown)


src/Graviton/FileBundle/Resources/definition/File.json
```diff
@@ -80,6 +80,12 @@
         "description": "Additional information for this file."
       },
       {
+        "name": "metadata.hash",
```
narcoticfresh: @jacobchristensen79 should be `readOnly: true` field no?

jacobchristensen79: no, the "new" feature is actually that they also can post this directly from the apps, calculate it from client side or server side if empty ( I think that this is incorrect as the rest of "why" this hash code, but that's another story ).
If file sent, is hash? no, then calculate. if no file, no hash change.

PR #436, opened 2016-06-29
When using guzzle to upload files these did not have filename and the…

jacobchristensen79:
… only way to send params like metadata was in header. This fixes both.

### DIFF
--- a/src/Graviton/FileBundle/FileManager.php
+++ b/src/Graviton/FileBundle/FileManager.php
@@ -12,6 +12,7 @@ use GravitonDyn\FileBundle\Document\File as FileDocument;
 use Symfony\Component\HttpFoundation\File\Exception\UploadException;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use GravitonDyn\FileBundle\Document\FileMetadata;
 use Symfony\Component\HttpKernel\Exception\HttpException;
@@ -378,7 +379,10 @@ class FileManager
             $fInfo = finfo_open(FILEINFO_MIME_TYPE);
             $contentType = finfo_file($fInfo, $file);
             if (!$contentType) {
-                throw new HttpException(400, "Could not determine Content type of file: ".$fileName);
+                throw new HttpException(
+                    Response::HTTP_NOT_ACCEPTABLE, 
+                    'Could not determine Content type of file: '.$fileName
+                );
             }
         }
 
--- a/src/Graviton/FileBundle/FileManager.php
+++ b/src/Graviton/FileBundle/FileManager.php
@@ -380,7 +380,7 @@ class FileManager
             $contentType = finfo_file($fInfo, $file);
             if (!$contentType) {
                 throw new HttpException(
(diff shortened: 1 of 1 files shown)


src/Graviton/FileBundle/FileManager.php
```diff
@@ -343,23 +353,30 @@ private function extractFileFromContent($fileInfoString)
...
-        preg_match('@name=\"([^"]*)\";\sfilename=\"([^"]*)\"\s*Content-Type:\s([^"]*)@', $fileInfo[0], $matches);
-        $originalName = $matches[2];
+        preg_match('/name=\"(.*?)\"[^"]/i', $fileInfo[0], $matches);
+        $name = isset($matches[1]) ? $matches[1] : '';
+
+        preg_match('/filename=\"(.*?)\"[^"]/i', $fileInfo[0], $matches);
+        $fileName = isset($matches[1]) ? $matches[1] : '';
+
+        preg_match('/Content-Type=\"(.*?)\"[^"]/i', $fileInfo[0], $matches);
+        $contentType = isset($matches[1]) ? $matches[1] : '';
```
narcoticfresh: @jacobchristensen79 in all 3 `preg_matches` you fallback to `''` if no match.. in name/filename this might be ok, but for the mime type this is imho not good.. this ends up that we're sending the user the `Content-Type: ` when he downloads the file and I don't know how the browser/client may react.. Maybe fall back to `application/octet-stream` in case of no header? or use the magic function to determine the file type?

src/Graviton/FileBundle/FileManager.php
```diff
@@ -343,23 +353,30 @@ private function extractFileFromContent($fileInfoString)
...
+
         $dir = ini_get('upload_tmp_dir');
         $dir = (empty($dir)) ? sys_get_temp_dir() : $dir;
-        $file = $dir . '/' . $originalName;
+        $file = $dir . '/' . $fileName;
```
narcoticfresh: @jacobchristensen79 please use `DIRECTORY_SEPARATOR` instead of `/` here to remain platform neutral

src/Graviton/FileBundle/FileManager.php
```diff
@@ -343,23 +354,39 @@ private function extractFileFromContent($fileInfoString)
...
 
+        // FileType Content-Type
+        preg_match('/Content-Type=\"(.*?)\"[^"]/i', $fileInfo[0], $matches);
+        if (isset($matches[1])) {
+            $contentType = $matches[1];
+        } else {
+            $fInfo = finfo_open(FILEINFO_MIME_TYPE);
+            $contentType = finfo_file($fInfo, $file);
+            if (!$contentType) {
+                throw new HttpException(400, "Could not determine Content type of file: ".$fileName);
```
narcoticfresh: @jacobchristensen79 this exception you throw actually is an exception in `JsonExceptionListener` and lands here https://github.com/libgraviton/graviton/blob/develop/src/Graviton/CoreBundle/Listener/JsonExceptionListener.php#L42

so the errror is rendered in html, not as json - that shouldn't be the case.. i don't know why this `400` check there exists..

jacobchristensen79: The 400, yes because we have another "render" for form 400 errors.
Should I change the error to be a 406 ( Not Acceptable ) ? 
then we have a 
{"code":0,"message":"Could not determine Content type of file: image.jpg"}

PR #437, opened 2016-07-04
try to disable xdebug when not needed on travis

narcoticfresh:
if you check https://travis-ci.org/libgraviton/graviton/builds/142129543, it seems to be a bit quicker.. so let's put it there ;-)


hairmare:
Have you considered doing this in the `composer install` step? It looks like you could get way more improvement out of this if you do it before `composer install` runs.

PR #449, opened 2016-07-23
EVO-6307 - Record origin (exception) constraint

narcoticfresh:
So this is for EVO-6307 - exceptions to the record origin rules

As you will see, this not only adds exceptions to recordOrigin. Instead, the whole checkin of the recordOrigin rules are now inside a Schema based validator - it's a better fit for how validation works now. And it needed to be changed as the previous logic was all inside `DocumentModel` in which we cannot validate too much.

It has the benefit that we don't need to fetch any additional data from the database anymore to do the validation. 

There are still the same rules.. A request will be denied if any of this orcurrs:
* A record with recordOrigin core wants to be deleted
* A record with the 'blacklist' recordOrigin (like `core`) wants to be inserted
* A record with recordOrigin core wants to be altered on an attribute that is not flagged as an exception (`recordOriginException: true` in the field definition)

**Changed behavior**
Now if a client GETs a `recordOrigin: core` record and does a PUT with it **unchanged**, it will be accepted as the entity wasn't changed. I think this is actually a more sane behavior as just denying it.

**Testing**
I replaced the scattered `recordOrigin: core` related tests with a more concise set of integration tests that test all known constellations.


src/Graviton/RestBundle/Tests/Model/DocumentModelTest.php
```diff
@@ -1,113 +0,0 @@
-<?php
```
tobster67: This test is completely removed instead of modified / extended according to the new possibilities. Is there an adequate replacement that I am just not seeing?

narcoticfresh: Yes, here we have the integration test..
https://github.com/libgraviton/graviton/blob/7d08251ec2178013406559655d757e47216ee14a/src/Graviton/SchemaBundle/Tests/Controller/RecordOriginConstraintTest.php

tobster67: didn't notice that this is an even better integration Test 👍

PR #450, opened 2016-07-25
EVO-5751 adding tests and command to build the mongodb search index

tobster67:
* adding post install command for index generation in travis.yml and dev-cleanstart.sh
* adding simple functional tests for the search
* adding a test for the command graviton:generate:build-indexes


src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php
```diff
@@ -42,6 +42,30 @@ public function setUp()
...
+    public function testSearchIndex()
+    {
+        $client = static::createRestClient();
+        $client->request('GET', '/core/module/?search(module)&select(key)');
+        $this->assertEquals('retirement', $client->getResults()[0]->key);
+        $this->assertEquals('realEstate', $client->getResults()[1]->key);
+        $this->assertEquals('investment', $client->getResults()[2]->key);
+        $this->assertEquals('requisition', $client->getResults()[3]->key);
+        $this->assertEquals('payAndSave', $client->getResults()[4]->key);
+
```
tobster67: just to explain why this test makes sense: Without an mongodb-text index, the first result of that search would be "AdminRef", since it doesn't contain the search term 'module' in its path (path is part of the defined index). So the test will fail then. The service will just return all datasets.

hairmare:
We seem to be having some mongodb version issues :( When I run the new tests against `mongo:latest` from the docker hub (aka mongodb 3.2.8) I get the following fails.

```
$ php56 vendor/bin/phpunit --testsuite=integration --filter=ModuleControllerTest
PHPUnit 5.4.5 by Sebastian Bergmann and contributors.

FF..................................                              36 / 36 (100%)

Time: 1.58 minutes, Memory: 762.75MB

There were 2 failures:

1) Graviton\CoreBundle\Tests\Controller\ModuleControllerTest::testSearchIndex
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'retirement'
+'realEstate'

/Users/hairmare/Documents/git.repos/graviton/src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php:60

2) Graviton\CoreBundle\Tests\Controller\ModuleControllerTest::testSearchWeightedIndex
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'retirement'
+'realEstate'

/Users/hairmare/Documents/git.repos/graviton/src/Graviton/CoreBundle/Tests/Controller/ModuleControllerTest.php:84

FAILURES!
Tests: 36, Assertions: 111, Failures: 2.
```

It took me quite some time to figure out that downgrading mongodb to `mongo:2` from the docker hub (aka 2.6.12) makes the errors go away.

Further testing then confirmed that tests also run on `mongo:3.0` (currently at 3.0.12) as well as `mongo:3.3` (at 3.3.10). Not being happy with these results I went a bit further and can now confirm that the breaking change was introduced with `v3.2.8`. I tried to confirm this with upstream but failed at finding anything in JIRA (https://jira.mongodb.org/browse/SERVER).

For this issue we need to figure out how we would like to handle this. I see multiple options:
- downgrade to an older mongo on production (3.2.7)
- upgrade to an unstable version (3.3.x)
- add a version check in graviton and "downgrade" the feature

I hope we can carve out some time to look into this tomorrow during/after story time.

hairmare:
It looks like updating mongo is easy on travis: https://github.com/travis-ci/travis-ci/issues/3694#issuecomment-155186560

This is currently running in https://travis-ci.org/libgraviton/graviton/builds/147219615 so we can see if it pulls the broken mongo on travis letting us reproduce this even moar.

hairmare:
Oll'right... I did even more testing and was able to get this running locally. Turns out MongoDB can be picky about versions. In my case updating the client to be the same version made everything work nicely. We might need to take care of matching versions on prod somehow since the come from different docker volumes.

Imma merge this now and if it breaks down the line we can still fix it. The main weirdness I was encountering was that the index was being created but just didn't work.

PR #452, opened 2016-07-26
EVO-5174 extending event/status status and a test for the new field

tobster67:
This extends the status Object inside the event/status Service with a translatable description field.
This should be used by the Workers to write something meaningful into it, that can be read by the upcoming Worker Dashboard.


src/Graviton/RabbitMqBundle/Resources/definition/EventStatus.json
```diff
@@ -69,6 +69,14 @@
         ]
       },
       {
+        "name": "status.0.description",
+        "type": "string",
+        "title": "Description",
+        "description": "Description of the status",
```
wegenmic: it's not the description of the status itself, but more the description of what the specific worker does

PR #453, opened 2016-07-26
Feature/evo 5751 add embedded textIndexes

jacobchristensen79:
Added the possibility to configure Text Indexes to embedded objects


composer.lock
```diff
@@ -5364,16 +5364,16 @@
     "packages-dev": [
         {
             "name": "graviton/test-services-bundle",
-            "version": "v0.21.0",
+            "version": "v0.20.0",
```
narcoticfresh: @jacobchristensen79 why u roll back the test bundle?

PR #455, opened 2016-07-27
Bump MongoDB on travis and enforce unit/integration test split

hairmare:
* [x] install mongodb 3.2 in test env (matches our current production)
* [x] move integration tests from unit testsuite to integration testsuite ae5e3dc1a13c3665202480961219c812cc5836f7
* [x] update configuration to point to a nonexistent mongodb when not doing integration tests (so new tests in the unit testsuite will fail if they try to access the database)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: php
-sudo: false
+sudo: required
 services:
 - mongodb
 - rabbitmq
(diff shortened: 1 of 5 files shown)


narcoticfresh:
@jacobchristensen79 i hope this will be green after the next run - please get it merged, then resume for remaining bumps in #467

PR #458, opened 2016-07-27
Rename header Allow to Access Controle Allow Methods

jacobchristensen79:


src/Graviton/ProxyBundle/Controller/ProxyController.php
```diff
@@ -163,6 +163,12 @@ protected function cleanResponseHeaders(HeaderBag $headers)
     {
         $headers->remove('transfer-encoding'); // Chunked responses get not automatically re-chunked by graviton
         $headers->remove('trailer'); // Only for chunked responses, graviton should re-set this when chunking
+
+        // Change naming to Graviton output Allow options
+        if ($allow = $headers->get('Allow', false)) {
```
lapistano: afaik there is a >corsresponselistener< why does this not apply?

src/Graviton/ProxyBundle/Controller/ProxyController.php
```diff
@@ -163,6 +163,12 @@ protected function cleanResponseHeaders(HeaderBag $headers)
...
+            $headers->set('Access-Control-Allow-Methods', $allow);
+            $headers->remove('Allow');
+        }
     }
```
jacobchristensen79: In CorsResponseListener
we do set: corsMethods to Access-Control-Allow-Methods
So we can also just add the set to use the key corsMethods.

lapistano:
this is just a proxy, which should not mess with the responses of the servers behind .. no fix!

PR #461, opened 2016-07-29
Feature/evo 6659 strict search dot

jacobchristensen79:
Required in this specific search criteria to escape the value dot's as not considered part of the string

### DIFF
--- a/src/Graviton/RestBundle/Model/DocumentModel.php
+++ b/src/Graviton/RestBundle/Model/DocumentModel.php
@@ -235,7 +235,10 @@ class DocumentModel extends SchemaModel implements ModelInterface
             foreach ($innerQuery->getQueries() as $key => $innerRql) {
                 if ($innerRql instanceof SearchNode) {
                     if (!$hasSearch) {
-                        $searchString = implode(' ', $innerRql->getSearchTerms());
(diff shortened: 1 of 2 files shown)


src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -235,7 +235,10 @@ private function buildSearchQuery(XiagQuery $xiagQuery, MongoBuilder $queryBuild
             foreach ($innerQuery->getQueries() as $key => $innerRql) {
                 if ($innerRql instanceof SearchNode) {
                     if (!$hasSearch) {
-                        $searchString = implode(' ', $innerRql->getSearchTerms());
+                        $searchString = '';
+                        foreach ($innerQuery->getSearchTerms() as $string) {
```
hairmare: Don't you want `$innerRql->getSearchTerms()` here?

jacobchristensen79: I need to do " " if it's a 333.444.555  search string query,  else it fetches all 333 or 444 or 555   so if I do a search= "333.444.555" it will search by this, same as if there is a search pattern with some.dot.here string.

PR #462, opened 2016-08-02
Add simple integration tests that exercises the dotted search feature

hairmare:


hairmare:
This seems to work on travis with 5.6 (the errors with php7 are most likely unrelated to the changes). What I don't get is why this is failing locally, i probably need to cleanstart again.

jacobchristensen79:
Just saw that the search is done on app/core, this has no text search indexes.

hairmare:
It's on `/core/module/` and uses the `key` field which should have an index.

hairmare:
replaced with #466

PR #463, opened 2016-08-03
EVO-6336 ommit doubled service display

tobster67:
* this fixes the problem of showing the whoami service in the services list twice

the display of the services is done by the method determineServices who takes as argument an array that holds all routes of graviton & the adapter. The whoami route is additionally defined as "additionalRoute" in app/config.yml which leads to the doubled  display of it here.


tobster67:
checks are currently failing cause the check in php7 runs out of memory. Problem will be fixed soon. Please feel free to merge anyway :)

PR #465, opened 2016-08-03
Fix search with space

jacobchristensen79:
We didn't consider people searches with name "space" surname. Tests will follow.


hairmare:
@jacobchristensen79 please add some kind of test, we'll never get to the bottom of these search changes if we don't test them.

hairmare:
whoopsie, I should have added WIP on this :/

PR #466, opened 2016-08-03
Add simple integration tests that exercises the dotted search feature

hairmare:
Same as #462 but against develop since that was merged.


tobster67:
Tried this locally. Nice Test. The field key has a weighted searchIndex. What about the WIP-tag? Want to merge this...

PR #467, opened 2016-08-04
"bump all the things"

jacobchristensen79:
Vendor library update

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -71,16 +71,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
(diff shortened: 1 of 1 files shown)


hairmare:
Needs some work on the generator parts since those seem to have changed their signature again. We've been here before, our stuff is intertwined with upstream generator parts in an unholy fashion.

It might make sense to do a partial update if you where after some specific bump.

narcoticfresh:
@jacobchristensen79 let's close this, please make new one after #455 merge..

PR #469, opened 2016-08-05
Remove Airbrake

jacobchristensen79:
Moving to other provider


hairmare:
Is any "other provider" coming or can this be finished and/or closed?

PR #470, opened 2016-08-09
Downgrade symfony and jms/serializer

hairmare:
These downgrade combat various issues we caught way too late.


hairmare:
Workaround for https://github.com/symfony/symfony/issues/19582 ;)

PR #474, opened 2016-08-17
Feature/evo 7007 selfversion with git

tobster67:
During build-time, git should be available. Since the versions.yml File is written during the build, triggered by a post-install-cmd from composer right before the whole stuff is copied to its location in the container this should also work in the docker environment.

Refactored the whole process, so that there is a new command available:
`php app/console graviton:core:generateversions`
The formerly defined Service CoreVersionUtils was never used again in the entire application. So this obviously made no sense and was kicked out. Its substitute is the mentioned command. A functional test was added for the command.

This is how the self-version is fetched:
* first the getContextVersion-Method gets the hash from the commit that was just checked out
* with that hash it looks to which tag this commit belongs to - taking this as the version
* if that fails, because there is no tag set yet it takes dev- and the current branchname as version


src/Graviton/CoreBundle/Command/GenerateVersionsCommand.php
```diff
@@ -67,6 +73,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $container = $this->getApplication()->getKernel()->getContainer();
         $rootDir = $container->getParameter('kernel.root_dir');
         $this->composerCmd = $container->getParameter('graviton.composer.cmd');
+        $this->gitCmd = $container->getParameter('graviton.git.cmd');
```
lapistano: what about:

`
$this->contextDir = $rootDir . ((strpos($rootDir, 'vendor'))? '/../../../../' :  '/../');
`

tobster67: why not 👍

src/Graviton/CoreBundle/Command/GenerateVersionsCommand.php
```diff
@@ -103,24 +110,33 @@ public function getPackageVersions()
...
+            // if empty, set dev- and current branchname to version:
+            if (!strlen($version)) {
+                $version = 'dev-' . trim($this->runGitInContext('rev-parse --abbrev-ref HEAD'));
+            }
 
+            $wrapper['id'] = 'self';
+            $wrapper['version'] = $version;
+        } else {
+            throw new CommandNotFoundException(
+                'getContextVersion: '. $this->gitCmdCmd . ' not available in ' . $this->contextDir
```
lapistano: $this->gitCmd vs $this->gitCmdCmd

tobster67: oh yeah ...

src/Graviton/CoreBundle/README.md
```diff
@@ -23,6 +24,19 @@ desiredVersions:
   - graviton/graviton
 ```
 
+#### How to (re)generate versions.yml
+* Make sure that composer and git are available on your machine either in $PATH or at a location that can be defined in the app/config/parameters.yml:
+```
+...
+    graviton.composer.cmd: composer
+    graviton.git.cmd: usr/local/git
```
lapistano: /usr/local/git 
/bin/git

pls use the absolute path

PR #477, opened 2016-08-19
pass select() directly to mongodb

narcoticfresh:
this is a major performance boost as we discussed in the common hipchat rooms - passing all fields in select() directly to mongo. 

the consequence on select()ing on embedded objects has been well communicated and it has been decided by mgmt that it's wanted nevertheless 
--> this also explains the changes to the ShowCaseControllerTest in regard to the select() operation there.

What changes in behavior
* User can not select on fields that are behind referenced objects. This is communicated. Workaround is to just select the higher levels or - if needed - we change the objects to embeds.
* "id" is now always rendered even if not in select(), as well as arrays in the path. This should not disturb any clients IMHO.

so here it is.. 

i added a small decorator to the JsonExceptionListener that renders a special message if we can detect that the user shaved off at the wrong level, disabling us from rendering an referenced object..


src/Graviton/RestBundle/Model/DocumentModel.php
```diff
@@ -357,12 +356,36 @@ public function insertRecord($entity, $returnEntity = true, $doFlush = true)
...
-     * @param string $documentId id of entity to find
+     * @param string  $documentId id of entity to find
+     * @param Request $request    request
      *
      * @return Object
      */
-    public function find($documentId)
+    public function find($documentId, Request $request = null)
     {
+        if ($request instanceof Request) {
```
lapistano: statement not necessary .. there already is the type hint.

narcoticfresh: this is may way to *not* write `if (!is_null($request))` as it's optional.. like `instanceof` more.. what gives to change? ;-)

PR #479, opened 2016-08-24
fix wrong id handling in recordorigin constraint

narcoticfresh:
triggered through EVO-6960, we found an error how the `RecordOriginConstraint` checks if a current record exists or not. Falsely, it assumed there will always be an ID in payload - but as `RestUtils` in its check, also `ConstraintUtils` must fall back to the `Request` attribute `id` to make sure. Actually, the `Request` attribute is the one that must win.

We still have the check in place (in `RestUtils`) that will deny a request with a different id in payload and in the request URL..

This PR fixes the issue and adds an integration test to avoid regression..

### DIFF
--- a/src/Graviton/SchemaBundle/Constraint/ConstraintUtils.php
+++ b/src/Graviton/SchemaBundle/Constraint/ConstraintUtils.php
@@ -8,6 +8,8 @@ namespace Graviton\SchemaBundle\Constraint;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Graviton\JsonSchemaBundle\Validator\Constraint\Event\ConstraintEventSchema;
 use Graviton\RestBundle\Service\RestUtils;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
  * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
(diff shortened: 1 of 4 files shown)


src/Graviton/SchemaBundle/Constraint/ConstraintUtils.php
```diff
@@ -79,11 +89,25 @@ public function getSerializedEntity($documentClass, $recordId)
      */
     public function getCurrentEntity()
     {
+        $currentRecordId = null;
+
+        // first, let's the one from the payload..
```
lapistano: can't we not just set the $currentRecordId  if there is no id in the attributes of the current request instead of overwriting it?

src/Graviton/SchemaBundle/Constraint/ConstraintUtils.php
```diff
@@ -79,11 +89,25 @@ public function getSerializedEntity($documentClass, $recordId)
...
+        if (isset($this->currentData->id)) {
+            $currentRecordId = $this->currentData->id;
+        }
+
+        // if we have a request, it must override it..
+        if ($this->requestStack->getCurrentRequest() instanceof Request &&
```
lapistano: don't like these train wrecks .. can't we do 
'$request = $this->requestStack->getCurrentRequest()'

and use this?  #estetics I know but much better to read ;)

PR #482, opened 2016-08-29
update json schema dep to new major..

narcoticfresh:
`justinrainbow/json-schema` has a new major, this PR bumps it to this new one.
Mostly the same, but the `Factory` has gone and some things changed with the `RefResolver` (which also doesn't exist anymore)..

I want this because the silly validation message `String value found, but a null is required` is now correct to the user by mentioning both types ;-)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -79,6 +79,7 @@ doctrine_mongodb:
             mappings:
                 GravitonCoreBundle: ~
                 GravitonRabbitMqBundle: ~
+                GravitonI18nBundle: ~
             filters:
                 softdeleteable:
                     class: Gedmo\SoftDeleteable\Filter\ODM\SoftDeleteableFilter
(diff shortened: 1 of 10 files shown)


hairmare:
@narcoticfresh please check what went wrong with the integration tests.

narcoticfresh:
@hairmare yes i started.. it's a weird error of a weird test that only happens when `integration` testsuite is executed 'standalone'.. the strange test is from @jacobchristensen79 , he'll take a look..

app/config/config.yml
```diff
@@ -79,6 +79,7 @@ doctrine_mongodb:
             mappings:
                 GravitonCoreBundle: ~
                 GravitonRabbitMqBundle: ~
+                GravitonI18nBundle: ~
```
narcoticfresh: @hairmare & @jacobchristensen79 
adding this should fix it - running again..

don't really know why though.. any reason why we don't just put `auto_mapping: true` as described [here](http://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/index.html)?

hairmare: I think I recall us changing that for some reason. I think there were some performance issues because we had to look though alot of GravitonDyn with `auto_mapping`. After some digging I wasn't able to locate the commit though.

PR #493, opened 2016-09-08
Remove unused libraries and Symfony to 2.8.8

jacobchristensen79:
Due to memory issues I started removing un-used vendors.


composer.json
```diff
@@ -41,13 +41,10 @@
         "graviton/rql-parser-bundle": "~0.11.0",
         "knplabs/knp-gaufrette-bundle": "~0.3",
         "aws/aws-sdk-php": "~2.8",
-        "eo/airbrake-bundle": "~1.0",
-        "hwi/oauth-bundle": "~0.4",
         "graviton/deploy-scripts": "~0.4",
         "graviton/json-schema": "~1.0",
         "php-jsonpatch/php-jsonpatch": "~3.0",
         "antimattr/mongodb-migrations": "^1.0",
-        "jenssegers/proxy": ">=3.0.0-beta2",
```
narcoticfresh: @jacobchristensen79 isn't this needed by sw*ssquant and/or our swaggerloader/proxy stuff here or does the sw*ssquant-bundle pull that in for themselves?

jacobchristensen79: Looks like this is used inside "ProxyBundle" but there is no test for it. Updating composer.

PR #494, opened 2016-09-09
[WIP] Feature/evo 7278 created by proposal

jacobchristensen79:
We need a silent DB insert on who modified and when a document on last time. 
Initially we intended to use the standard createdAt/By and updatedAt/By but due to delete the document on "update" I could not use it that way. So after discussion I propose the simple solution to run it using modifiedAt/By. 
This can be queried in RQL but value will not be visible in any output ( Unless specifically "exposed" ).


jacobchristensen79:
I would have liked to use doctrine listener.

jacobchristensen79:
Just found how to use EventListener for Mongo and updated code.

narcoticfresh:
close after discussion with @jacobchristensen79 of different way to implement this

PR #499, opened 2016-09-19
Bugfix: don't show empty arrays in JSON when deselected

tobster67:
Bug: When querying the RestAPI and selecting fields with "select(foo)" there are still all embedded Arrays from an entity shown as empty Arrays, even when they are not selected.
* The generated doctrine Hydrators always populate the embedded Arrays in an entity even when there is no value for them resulting from the database query. They populate them with an empty array instead of setting them to Null.
* This PR adds an ExclusionStrategy to the JMS-Serializer. If a client is selecting some fields from datasets then only these are serialized. The others are skipped. This applies to the wrongly populated Arrays mentioned above as well and solves the problem.


src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
```diff
@@ -0,0 +1,62 @@
...
+    public function shouldSkipProperty(PropertyMetadata $property, Context $context)
+    {
+        // we are only dealing with the first level of the JSON here
+        if ($context->getDepth() > 1) {
+            return false;
+        }
+
+        $select = false;
+        $selectVars=[];
+        foreach ($_GET as $getVar => $val) {
```
narcoticfresh: @tobster67 you should not really use superglobals as `$_GET` directly in the code and re-parse RQL stuff.. can u not inject/get the Xiag RQL component here?

Maybe you can let something (like the RQL listener) inject the Xiag RQL component on the `Request` (as request attribute), then inject the `request_stack` here and re-fetch the `Request` attribute here..(?)

narcoticfresh:
@tobster67 i checked again and it seems `XiagQuery` is already a request attribute.. Please check how `DocumentModel` [does it's things in this line](https://github.com/libgraviton/graviton/blob/v0.75.0/src/Graviton/RestBundle/Model/DocumentModel.php#L147).

So this means you can just inject `request_stack` ([which is this](http://api.symfony.com/3.1/Symfony/Component/HttpFoundation/RequestStack.html)) into your `ExclusionStrategy`, this provides the `getCurrentRequest()` function, then you can get the `XiagQuery` easily and call `getSelect()` (or whatever it is)..

Should make it way nicer ;-)

src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
tobster67: thanks. Will do so.

tobster67:
@narcoticfresh It did make it a lot nicer and actually less complicated :)

src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function shouldSkipProperty(PropertyMetadata $property, Context $context)
+    {
+        // we are only dealing with the first level of the JSON here
+        if ($context->getDepth() > 1) {
+            return false;
+        }
+
+        $return = false;
+        $xiagQuery = $this->requestStack->getCurrentRequest()->get('rqlQuery');
```
narcoticfresh: @tobster67 please check if `$this->requestStack->getCurrentRequest()` is actually `instanceof Request`, that's not always the case.. you'll have `cannot call method on non-object` errors`.. 

**And please please** run the unit tests locally (and `composer check`) before you commit, it saves a lot of unnecessary commits in case of failure..

src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
```diff
@@ -0,0 +1,115 @@
...
+                    $this->selectedFields = $select->getFields();
+                    // get the nested fields as well
+                    $nestedFields = [];
+                    foreach ($this->selectedFields as $key => $field) {
+                        if (strstr($field, '.')) {
+                            $nestedFields=array_merge($nestedFields, explode('.', $field));
+                            unset($this->selectedFields[$key]);
+                        }
+                    }
+                    $this->selectedFields = array_merge($this->selectedFields, $nestedFields);
```
tobster67: yes You would probably get the first someProperty as well. I will revamp that one.

tobster67: Actually this would only apply if someProperty would be an array, cause the select query is applied to the mongo odm and all none array values that are not selected result as Null at the time they arrive here. But still - good point. Will add a test to check this as well.

PR #500, opened 2016-09-20
Make it configurable to hide a field if its embedded extref is empty

tobster67:
Problem: Some clients of the RestAPI have difficulties finding no $ref- entry embedded in an JSON-Object for extrefs. Even if this is a valid JSON-Object.
This could be solved globally with a lot of unwanted side-effects. Therefore this PR introduces the setting `hideOnEmptyExtref` for the wrapper Class of an Extref-Definition:
```
...
{
   "name": "hiddenOnEmptyOptionalExtref",
   "type": "object",
   "hideOnEmptyExtref": true
},
{
    "name": "hiddenOnEmptyOptionalExtref.ref",
     "type": "extrem",
     "exposeAs": "$ref",
     "required": false
}
...
```
Like this, this behavior can be punctually defined where its needed.

### DIFF
--- a/src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentBase.php.twig
+++ b/src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentBase.php.twig
@@ -39,10 +39,10 @@ protected $deletedDate;
 
 {% for field in fields %}
     /**
-    {% if 'Graviton' in field.type %}
-        * @var \{{ field.type }} ${{ field.fieldName }}
+    {% if 'Graviton' in field.serializerType %}
+        * @return \{{ field.serializerType }} ${{ field.fieldName }}
     {% else %}
-        * @var {{ field.type }} ${{ field.fieldName }}
+        * @return {{ field.serializerType }} ${{ field.fieldName }}
     {% endif %}
     */
     {% if field.type == 'array' %}
@@ -130,10 +130,10 @@ return $this->deletedDate;
     /**
     * Get {{ field.fieldName }}
     *
-    {% if 'Graviton' in field.type %}
-        * @return \{{ field.type }} ${{ field.fieldName }}
+    {% if 'Graviton' in field.serializerType %}
+        * @return \{{ field.serializerType }} ${{ field.fieldName }}
     {% else %}
-        * @return {{ field.type }} ${{ field.fieldName }}
+        * @return {{ field.serializerType }} ${{ field.fieldName }}
     {% endif %}
     */
     {% if field.type == 'boolean' %}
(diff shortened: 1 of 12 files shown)


tobster67:
added the WIP Label, cause there is Test coming up. This Test depends on this PR from the GravitonTestServicesBundle https://github.com/libgraviton/GravitonTestServicesBundle/pull/32 to be merged, the Bundle to be tagged and bumped into the composer.lock of Graviton. So as soon as this is done the Test can be pushed without the Travis Build failing.

src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentBase.php.twig
```diff
@@ -39,10 +39,10 @@ protected $deletedDate;
 
 {% for field in fields %}
     /**
-    {% if 'Graviton' in field.type %}
-        * @var \{{ field.type }} ${{ field.fieldName }}
+    {% if 'Graviton' in field.serializerType %}
```
narcoticfresh: @tobster67 i think those changes (changing `type` to `serializerType`) is quite problematic..

see, `type` here is abstracted somewhere from the `CompilerPasses` and consilidated to me more in line as it's quite *mühsam* as `doctrineType`/`serializerType` always differ in some aspects. So `type` was introduced to be the *one and only* value to compare against..

why are those changes even needed to `serializerType`.. please revert if not needed (and I don't see why they would)

src/Graviton/GeneratorBundle/Resources/skeleton/document/DocumentBase.php.twig
```diff
@@ -145,17 +145,21 @@ return $this->deletedDate;
     {% if field.xDynamicKey is defined and field.xDynamicKey != null %}
         return XDynamicKey::resolveRef($this->{{ field.fieldName }}, '{{ field.xDynamicKey.refField }}');
     {% else %}
-        return $this->{{ field.fieldName }};
+        {% if field.hideOnEmptyExtref is defined and field.hideOnEmptyExtref==true and field.relType=='embed' %}
+            return $this->{{ field.fieldName }} && $this->{{ field.fieldName }}->getRef() ? $this->{{ field.fieldName }} : null;
```
narcoticfresh: @tobster67 this is also troublesome.. you're blindly assuming here that `hideOnEmptyExtref` is only set on objects that have `$this->{{ field.fieldName }}->getRef()` - without any assurance, without any further checks.. and why the `field.relType` check is there is also not clear to me..

narcoticfresh:
@tobster67 actually looking at all this, i remembered i did this already.. i had this in this PR -> https://github.com/libgraviton/graviton/pull/427

my attempt was quite different: i check if on a given object if there is only 1 property and *if* that one property is an extref - if yes, i would hide it if empty..

i still think that the approach from my PR is smarter and it doesn't needs any configuration as it only applies to objects that have only 1 property and that is an extref..

the PR was completed, only the unit test expectations needed adaption - same state as this ;-)

narcoticfresh:
@tobster67 i also think #427 is a bit better as it archieves the same with less changes, no configuration needs and is a completely serializer based approach and is not bound to any implicit conventions

tobster67:
@narcoticfresh OMG, all this for a problem that might be already solved ;) The idea of making this being configurable was to minimize side effects. So it can be used only on the fields where the clients have their problems with the (valid) json. I guess the possible side-effects where the reason why You closed Your PR in the end, which looks like a good solution to the same "problem"?

narcoticfresh:
@tobster67 
If we look at EVO-5807 - it clearly states "always add $ref or Nothing" ;-) so case of iPad, they want to complete remove *all* those objects, so every empty `extref` holding object shall not be rendered..

i think if we make it configurable, we end up configuring it everywhere anyway..

knowing BECO a bit better know, I don't think they will have any problems as we usually don't render empty objects anyway (unless they contain an empty array deeper)..

so i think we could use #427

tobster67:
@narcoticfresh fine with me! Too bad we didn't discuss that before I started working on this. But anyway... Lets use #427 I agree that this is the cleaner solution.

narcoticfresh:
@tobster67 ok, i'll try to update the branch again and make it green, then you can take it over to add integration tests (same as here)..

tobster67:
@narcoticfresh ok!
Agreed to use #427 to solve this problem in a more general way on serializer side. The expected side effects on client side should be not to severe according to @narcoticfresh .

PR #506, opened 2016-10-07
[WIP] Generating a JSON mongodb schema export

jacobchristensen79:
To decouple the dependency on our loader we will need a way to tell this how the current schema looks like.
EVO-7717

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,267 @@
+{
+    "App": {
+        "class": "Graviton\\CoreBundle\\Document\\App",
+        "fields": {
+            "id": {
+                "fieldName": "id",
+                "id": true,
+                "strategy": "UUID",
+                "options": [],
+                "name": "_id",
+                "isCascadeRemove": false,
+                "isCascadePersist": false,
+                "isCascadeRefresh": false,
+                "isCascadeMerge": false,
+                "isCascadeDetach": false,
+                "type": "custom_id",
+                "nullable": false,
+                "isOwningSide": true,
+                "isInverseSide": false
+            },
+            "name": {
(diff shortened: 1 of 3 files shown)


src/Graviton/GeneratorBundle/Command/SchemaExportCommand.php
```diff
@@ -0,0 +1,128 @@
...
+     * @param OutputInterface $output output
+     *
+     * @return void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        // Check mongo db version
+        $mongoVersion = $this->getMongoDBVersion('Graviton\\CoreBundle\\Document\\App');
+        if ((float) $mongoVersion < 2.6) {
+            $output->writeln("MongoDB Version < 2.6 installed: " . $mongoVersion);
```
jacobchristensen79: Copied from Generate Indexes Command. This, is actually not a requirement and can be removed. We do not need DB connection in this generator, all is in xml schema files.

schema.json
```diff
@@ -0,0 +1,267 @@
...
+                "isCascadeMerge": true,
+                "isCascadeDetach": true,
+                "nullable": false,
+                "association": 3,
+                "isOwningSide": true,
+                "isInverseSide": false
+            }
+        }
+    }
+}
```
jacobchristensen79: The schema json file is a dynamic file.
Created using: json_encode PRETTY of an Array.
The file is marked as green in Style board. Adding a \n in php would add this new line jump but not make any difference.  
Still WIP, and as a proposal for our db loader.

PR #513, opened 2016-10-25
Feature/evo 8123 rql select update

jacobchristensen79:
RQL select error discovered in very nested element. 
The added test do not fully cover the issue as the problem is when using a select with:
field1.field2.field3.field4
field1.field2.fieldX3.fieldX4

if field3 is empty the hole select for field1 is returned as empty and unset to be displayed in url.

### DIFF
--- a/src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
+++ b/src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
@@ -10,6 +10,7 @@ use JMS\Serializer\Metadata\PropertyMetadata;
 use JMS\Serializer\Context;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Xiag\Rql\Parser\Query;
+use Xiag\Rql\Parser\Node\SelectNode;
 
 /**
  * In this Strategy we skip all properties on first level who are not selected if there is a select in rql.
@@ -26,16 +27,6 @@ class SelectExclusionStrategy implements ExclusionStrategyInterface
     protected $requestStack;
 
     /**
-     * @var Array $selectedFields contains all the selected fields and its combinations with nested fields
-     */
-    protected $selectedFields;
-
-    /**
-     * @var Array $selectedLeafs contains the leafs of the selection "tree"
-     */
-    protected $selectedLeafs;
-
-    /**
      * @var Boolean $isSelect
      */
     protected $isSelect;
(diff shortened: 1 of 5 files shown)


jacobchristensen79:
Please wait to merge until testcase 
https://github.com/libgraviton/GravitonTestServicesBundle/pull/34
is completed and last commit pushed here.

jacobchristensen79:
PR updated with v0.23.0 of the tester

src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
```diff
@@ -58,9 +49,23 @@ class SelectExclusionStrategy implements ExclusionStrategyInterface
...
+     * @param array  $arr  output array
+     * @param string $path string dotet array
+     *
+     * @return void
+     */
+    private function createArrayByPath(&$arr, $path)
+    {
+        $keys = explode('.', $path);
+        while ($key = array_shift($keys)) {
+            $arr = &$arr[$key];
```
lapistano: mmh .. referencing a reference .. this might cause circular references which will make it impossible for the garbage collection to be cleaned up #memoryLeak

jacobchristensen79: this makes a : 
field.bla.foo.bar to a
array[field][bla][foo][bar]

src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
```diff
@@ -69,35 +74,27 @@ public function __construct(RequestStack $requestStack)
...
-                        $this->selectedFields[] = implode('.', array_slice($nestedFields, 0, $i));
-                    }
-                }
+            // Build simple selected field tree
+            foreach ($select->getFields() as $field) {
+                $field = str_replace('$', '', $field);
+                $arr = [];
+                $this->createArrayByPath($arr, $field);
+                $this->selectTree = array_merge_recursive($this->selectTree, $arr);
             }
```
jacobchristensen79: No reference here. Arr is new in each loop

src/Graviton/RestBundle/ExclusionStrategy/SelectExclusionStrategy.php
q-src: I'd suggest to implement `createArrayByPath($path)` returning an array. Then there would be no need to initialize an empty array beforehand and pass it by reference to the method.

PR #521, opened 2016-11-09
Feature/evo 8294 file upload

jacobchristensen79:
A file refactor. 
1 - fileController to be as a normal controller extend from restController.
2- File structure.
3- Clean up services, some implementation for put and post requests.
4- Auto set mime type, create and  modified date.
5- Implement optional allowed mime types.


src/Graviton/FileBundle/Manager/FileManager.php
```diff
@@ -0,0 +1,298 @@
...
+        /** @var FileMetadataBase $metadata */
+        $metadata = $file->getMetadata();
+        if (!$metadata) {
+            throw new InvalidArgumentException('Loaded file have no valid metadata');
+        }
+
+        // If no data, 404.
+        $mimeType = $this->fileSystem->mimeType($file->getId());
+        if ($mimeType !== $file->getMetadata()->getMime()) {
+            throw new InvalidArgumentException('Loaded file do not match source MimeType');
```
narcoticfresh: @jacobchristensen79 this here is problematic..
i switch to 'develop' and just POSTed a file to `/file` - without any mime-type given.. so current develop saves it without `metadata.mime`..

after that I switched to this branch and tried to re-fetch my file. then i've seen this exception..

please add the following behavior:
* if `$file->getMetadata()->getMime()` is empty, you accept it and send `Content-Type` header with the content of `$this->fileSystem->mimeType($file->getId())`, allowing the user to retrieve its files..

this is about staying BC..

narcoticfresh:
@jacobchristensen79 
also consider this workflow, `/file` behavior is not optimal
* insert a new file: `curl -X PUT --data @testfile.png "http://localhost:8000/file/testfile"`
* get the file --> mime-type will be correctly (`image/png` in this case)
* try to update the mime type in db 
```
curl -X PUT -H "Content-Type: application/json" -H "Accept: application/json" -d '{
  "metadata": {
    "size": 238144,
    "action": [],
    "mime": "othertype",
    "createDate": "2016-11-14T12:23:13+0000",
    "modificationDate": "2016-11-14T12:24:03+0000",
    "filename": "customFileName",
    "hash": "6c676783c2af625fef57feff238409cadcea7e18e12bc15d8e0d52d6df3b9388",
    "additionalProperties": []
  },
  "links": [],
  "id": "testfile"
}' "http://localhost:8000/file/testfile"
```
--> mime types gets updated

* try to fetch it again -> you will receive ` {
  "code": 0,
  "message": "Loaded file do not match source MimeType"
}` exception

**needed change**: don't accept mime-type change in `PUT`

src/Graviton/FileBundle/Manager/FileManager.php
jacobchristensen79: MimeType is added and created automatically.

PR #528, opened 2016-11-17
EVO-8294 On posting a file, there should be the option to set custom HASH.

jacobchristensen79:
Based on EVO-5984, added the functionality. Posted files will now have custom hash if so is sent.

### DIFF
--- a/src/Graviton/FileBundle/Manager/FileManager.php
+++ b/src/Graviton/FileBundle/Manager/FileManager.php
@@ -217,7 +217,8 @@ class FileManager
 
         // File related, if no file uploaded we keep original file info.
         if ($file) {
-            $hash = hash('sha256', file_get_contents($file->getRealPath()));
+            $hash = $metadata->getHash() ? $metadata->getHash() :
+                hash('sha256', file_get_contents($file->getRealPath()));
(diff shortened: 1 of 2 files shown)


src/Graviton/FileBundle/Manager/FileManager.php
```diff
@@ -217,7 +217,8 @@ private function buildFileDocument(FileDocument $document, $file, $original)
 
         // File related, if no file uploaded we keep original file info.
         if ($file) {
-            $hash = hash('sha256', file_get_contents($file->getRealPath()));
+            $hash = $metadata->getHash() ? $metadata->getHash() :
```
lapistano: I think this is on of the data fields, which should be handled exclusively by the backend. Why should a client be able to set this filed? 
Further this data can be fetched from the backend afterwards. 
If it is necessary to receive this information without an extra request, one could discuss a custom header field.

jacobchristensen79: hash is in Metadata received from the POST or PUT request. 
The why a client, please check referenced EVO in description. This is implemented currently but forgotten in refactor due to missing test, which is now added.

PR #530, opened 2016-11-22
Language Translatable ID to be sha1 encoded to avoid index overflow.

jacobchristensen79:
Due to long original translation we need to remove the index original and update how id is generated

### DIFF
--- a/src/Graviton/I18nBundle/Migrations/MongoDB/Version20161122111410.php
+++ b/src/Graviton/I18nBundle/Migrations/MongoDB/Version20161122111410.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * Index migration
+ */
+
+namespace Graviton\I18nBundle\Migrations\MongoDB;
+
+use AntiMattr\MongoDB\Migrations\AbstractMigration;
+use Doctrine\MongoDB\Database;
+use Graviton\I18nBundle\Document\Translatable;
+use MongoDB;
+use MongoCollection;
+
+/**
+ * Migrate domain_1_locale_1_original_1 index
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/MIT MIT License
+ * @link     http://swisscom.ch
+ */
+class Version20161122111410 extends AbstractMigration
+{
+    /**
+     * @var string
+     */
+    private $collection = 'Translatable';
+
+    /**
+     * @return string
+     */
+    public function getDescription()
+    {
+        return 'Build id for the new sha1 implementation';
+    }
+
+    /**
+     * recreate index without unique flag
+     *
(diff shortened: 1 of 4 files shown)


src/Graviton/I18nBundle/Migrations/MongoDB/Version20161122111410.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public function up(Database $db)
+    {
+        $collection = $db->createCollection($this->collection);
+        $qb = $collection->createQueryBuilder();
+        /** @var Translatable $translatable */
+        foreach ($qb->find() as $translatable) {
+            $id = $translatable->getId();
+            $collection->update(
+                ['id' => $id],
```
narcoticfresh: @jacobchristensen79 please check also with `ctype_xdigit` if given id already could be a hash and don't do it then..! we must assume that migrations *could* (even if *shouldn't*) run more than once - and then, we don't want it to 'destroy' data..

src/Graviton/I18nBundle/Tests/Controller/TranslatableControllerTest.php
```diff
@@ -43,7 +43,7 @@ public function testContainsReference()
         $client->request('GET', '/i18n/translatable/');
         $results = $client->getResults();
 
-        $this->assertEquals('http://localhost/i18n/language/de', $results[0]->language->{'$ref'});
+        $this->assertEquals('http://localhost/i18n/language/de', $results[1]->language->{'$ref'});
     }
```
jacobchristensen79: the change here is because the default sort by is ID and the sha1 result is a bit different

src/Graviton/I18nBundle/Tests/Controller/TranslatableControllerTest.php
```diff
@@ -43,7 +43,7 @@ public function testContainsReference()
         $client->request('GET', '/i18n/translatable/');
         $results = $client->getResults();
 
-        $this->assertEquals('http://localhost/i18n/language/de', $results[0]->language->{'$ref'});
+        $this->assertEquals('http://localhost/i18n/language/de', $results[1]->language->{'$ref'});
```
narcoticfresh: @jacobchristensen79 hm, this change shouldn't be necessary(?)

src/Graviton/I18nBundle/Migrations/MongoDB/Version20161122111410.php
```diff
@@ -0,0 +1,67 @@
...
+                ['id' => sha1($id)]
```
narcoticfresh: @jacobchristensen79 please check also with `ctype_xdigit` if given id already could be a hash and don't do it then..! we must assume that migrations *could* (even if *shouldn't*) run more than once - and then, we don't want it to 'destroy' data..

PR #533, opened 2016-11-24
File create date shall be kept.

jacobchristensen79:
On put requests, file metadata creationDate could be overwritten. This should not be possible.
We could define service as read-only field but we need to keep back-compatibility until workers and clients are updated.


lapistano:
don't see how this solves the problem, that the createdDate ia also set on every PUT request ..

jacobchristensen79:
Original, is when you are updating. if Document exists, it's the original document, then if so it will take metadata creationDate from the original. Now, it do not work so.

PR #534, opened 2016-12-02
Feature/evo 8323 configurable cache

jacobchristensen79:
Configurable cache handling for Documents.
Due to update document is deleted before inserted as updated document we need a temporary cache. Also, added the option so other collections can be saved for Get requests.

There is 2 implementations in this PR.

A) On update using RestController, any call to GET by ID will be paused if there is an update in progress on that element.  Implementing full cache was not an option here, yet.

B) Configurable cached Document. Only available for GET by ID, no option to use select rql, it's being used for EventStatus mainly as this service is pulled by many workers at the same time.

### DIFF
--- a/src/Graviton/DocumentBundle/Service/CollectionCache.php
+++ b/src/Graviton/DocumentBundle/Service/CollectionCache.php
@@ -127,7 +127,7 @@ class CollectionCache
         $key = self::BASE_UPDATE_KEY.'-'.$this->buildCacheKey($collection, $id);
 
         while ($this->cache->fetch($key)) {
-            sleep(0.25);
+            usleep(250000);
         }
     }
 
--- a/src/Graviton/DocumentBundle/Service/CollectionCache.php
+++ b/src/Graviton/DocumentBundle/Service/CollectionCache.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * ExtReferenceConverter class file
+ */
+
+namespace Graviton\DocumentBundle\Service;
+
+use Doctrine\Common\Cache\CacheProvider;
+use Doctrine\ODM\MongoDB\DocumentRepository as Repository;
+
+/**
+ * graviton.cache.collections
+
+ * Collection Cache Service
+ *
+ *
+ *
+ * @author   List of contributors <https://github.com/libgraviton/graviton/graphs/contributors>
+ * @license  http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @link     http://swisscom.ch
+ */
+class CollectionCache
+{
+    /** Prefix cache key */
+    const BASE_KEY = 'CollectionCache';
+
+    /** @var array  */
+    private $config = [];
+
+    /** @var CacheProvider */
+    private $cache;
+
+    /**
+     * CollectionCache constructor.
+     * @param CacheProvider $cache         Cache provider
+     * @param array         $configuration Collections to be saved
+     */
+    public function __construct(
+        CacheProvider $cache,
+        $configuration
+    ) {
+        $this->cache = $cache;
+        $this->config = $configuration;
+    }
+
+    /**
+     * Makes an id
+     *
+     * @param string $collection DB collection name
+     * @param string $id         Object Identifier
+     * @return string
+     */
+    private function buildCacheKey($collection, $id)
+    {
+        return self::BASE_KEY.'-'.preg_replace("/[^a-zA-Z0-9_-]+/", "-", $collection.'-'.$id);
+    }
+
+    /**
+     * Time it should be in cache and if so should happen
+     *
+     * @param string $collection DB collection name
+     * @return int
+     */
+    private function getCollectionCacheTime($collection)
+    {
+        if (array_key_exists($collection, $this->config)) {
+            return (int) $this->config[$collection];
+        }
+        return 0;
+    }
+
+    /**
+     * Return un cached object.
+     *
+     * @param Repository $repository DB Repository
+     * @param string     $id         Queried is
+     * @return object|false if no cache found
+     */
+    public function getByRepository(Repository $repository, $id)
+    {
+        $collection = $repository->getClassMetadata()->collection;
+        if (!$this->getCollectionCacheTime($collection)) {
+            return false;
+        }
+        $key = $this->buildCacheKey($collection, $id);
+
+        if ($result = $this->cache->fetch($key)) {
+            return unserialize($result);
+        }
+        return false;
+    }
+
+    /**
+     * @param Repository $repository DB Repository
+     * @param object     $document   Object document
+     * @return bool
+     */
+    public function setByRepository(Repository $repository, $document)
+    {
+        if (empty($document)) {
+            return false;
+        }
+        $collection = $repository->getClassMetadata()->collection;
+        if (!$time = $this->getCollectionCacheTime($collection)) {
+            return false;
+        }
+        $key = $this->buildCacheKey($collection, $document->getId());
+
+        return $this->cache->save($key, serialize($document), $time);
+    }
+}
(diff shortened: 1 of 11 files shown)


src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -459,12 +477,14 @@ public function patchAction($id, Request $request)
             $patch = new Patch($jsonDocument, $request->getContent());
             $patchedDocument = $patch->apply();
         } catch (InvalidPatchDocumentJsonException $e) {
-            throw new InvalidJsonPatchException($e->getMessage());
+            goto patch_error_handle;
```
q-src: Please do not use `goto`. I'd rather use a method which does the cleanup and then throws the exception.

src/Graviton/DocumentBundle/Service/CollectionCache.php
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Time it should be in cache and if so should happen
+     *
+     * @param string $collection DB collection name
+     * @return int
+     */
+    private function getCollectionCacheTime($collection)
+    {
+        if (array_key_exists($collection, $this->config)) {
+            return (int) $this->config[$collection];
```
q-src: In my opinion, it would make more sense if this value was the max. amount of seconds / milliseconds a resource can be locked. However, this seems to be the duration a resource gets cached for GET, right? 
And what happens if a locked resource is getting accessed which is no longer in cache?

src/Graviton/DocumentBundle/Service/CollectionCache.php
```diff
@@ -0,0 +1,180 @@
...
+        if (!$time = $this->getCollectionCacheTime($collection)) {
+            return false;
+        }
+        $key = $this->buildCacheKey($collection, $id);
+
+        return $this->cache->save($key, $serialized, $time);
+    }
+
+    /**
+     * Will sleep until previous operation has finished but for max 10s
```
q-src: 10s is not fully correct, since one could also do `addUpdateLock($repository, $id, 100);`

src/Graviton/DocumentBundle/Service/CollectionCache.php
```diff
@@ -0,0 +1,180 @@
...
+        return false;
+    }
+
+    /**
+     * @param Repository $repository DB Repository
+     * @param string     $serialized Serialised Object document
+     * @param string     $id         Object document identifier
+     * @return bool
+     */
+    public function setByRepository(Repository $repository, $serialized, $id)
```
q-src: Why do we need to set the lock and the resource cache separately? Can't the resource cache just be the lock as well?

src/Graviton/RestBundle/Controller/RestController.php
jacobchristensen79: Not gonna remove goto until we make a cleaner catch

src/Graviton/DocumentBundle/Service/CollectionCache.php
jacobchristensen79: True, comment should have been "default" value 10 seconds. But due to tests I preferred the option to modify it.

src/Graviton/DocumentBundle/Service/CollectionCache.php
jacobchristensen79: No, they are 2 different things. 1 is a cache and the other a lock option.
The caching for event status is as we do several calls on the same resource almost at the same time we save it cache to avoid many calls to db. 
The second, the update request, if any collection - ID, any, is being updated we wait on patch, put and get request. Post is not needed due to only be callable if object do not exists.
So, just to remark it, we have 2 functionalities, the option to cache GET requests and the other to avoid overwriting on update due to "insert/delete" method used in "update".

I can not implement the caching in ModelDocument:updateRecord as I do not wanted to use the serialiser (jms) and the binding get lost if serialising the $result.

src/Graviton/DocumentBundle/Service/CollectionCache.php
jacobchristensen79: The question, about a locked resource, Well, there will be NO possibility to access via GET by ID on a locked resource. If you have a look at the code you will see the the findRecord in RestController will wait if a update is performed. So I do not really understand your question.

q-src: Ah now I get it. I initially understood the `$configuration` array as a config which selectively enables the locking mechanism. AFAIK, we agreed on such a config last Thursday. Correct me if I'm wrong, but it looks to me that this feature is currently missing and locking is enabled for all services.

jacobchristensen79: The update lock is for all. it's a simple "wait" lock. Feel free to push your changes and stop writing.

jacobchristensen79: btw, there is a "why" it's good to have lock on all, because else a resource that exists may return "404" just because someone is updating....

src/Graviton/RestBundle/Controller/RestController.php
```diff
@@ -407,17 +414,22 @@ public function putAction($id, Request $request)
...
+        $repository = $model->getRepository();
+
         // And update the record, if everything is ok
         if (!$this->getModel()->recordExists($id)) {
             $this->getModel()->insertRecord($record, false);
         } else {
             $this->getModel()->updateRecord($id, $record, false);
         }
 
+        $this->collectionCache->releaseUpdateLock($repository, $id);
```
q-src: I think we should also update the cached document at this point (and after PATCH as well). Otherwise it is more likely that the changes get reverted by another client's GET and PUT.

src/Graviton/RestBundle/Controller/RestController.php
q-src: Not gonna merge as long as there is a goto.

goto comic

src/Graviton/RestBundle/Controller/RestController.php
jacobchristensen79: I tried... explained in  previous. As said, feel free to play.... I just ran into a lot of problems and as we do the stuff in RestController, not in DocumentModel where we NEED the JMS serialiser, but we have the "protected" function in restController not in the DocumentModel....
In my opinion, WE should do a refactor of DocumentModel and clean-up RestController to NOT have any private nor protected functions but a clean and nice "actions".
Feel free to do so, but remember, issue was and is meant to solve a error and have to be online next week.

src/Graviton/RestBundle/Controller/RestController.php
jacobchristensen79: not gonna change the code because you don't like it, and as said in private, feel free to change it to a \Exception

jacobchristensen79: uh, and yes, it's this "goto" or the 5 catches in a row ...  If I'm not wrong you can also commit into this branch.

narcoticfresh:
@jacobchristensen79 & @q-src 
I like to put this PR on hold as there seem to be some misunderstandings. as time becomes available i'll have a review on it.

@jacobchristensen79 i don't like the notion that reviewers should add code to your PR. that may be applicable for pairing scenarios. but when somebody just reviews it, it's the PR's authors job to either deny or apply the feedback, not telling the reviewer to add commits.

as this is not the most critical of features to merge, let's put this one on hold and i'll get back to it when time is available..

jacobchristensen79:
Narcoticfresh,  just because one do not like goto and don't wanna pair up nor wanna understand nor do a flow review doesn't mean the person in question is able to reject a PR without giving any other comment that "i do not like". If so, there are many thing I do not like but I do not complain and I fix/solve or improve when I have the opportunity.  face to face and private chat I said several times that if there is a improvement possible please, just commit it and let's discus it. 

I close the issue here by and leave it for the perfect implementation and as part of the refactor I would like to have in RestController and DocumentModel

PR #549, opened 2017-01-27
Jms/Serializer v1.4.2 and xml generator updates

jacobchristensen79:
We did previously attempt to update the serialiser to have a cleaner solution to the addData. 
The error was that some needed services requested array key value to be returned but without changing the xml serialiser Type field it was not returned so but as list of objects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "sensio/generator-bundle": "~3.0",
         "incenteev/composer-parameter-handler": "~2.1",
         "jms/serializer-bundle": ">=1.1,<2",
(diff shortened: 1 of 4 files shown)


jacobchristensen79:
Serialise works, but deserialise do not ( on put or record update ). user_array function error thrown in Doctrine UnityOFWork. 
will need to review the error when time

narcoticfresh:
@jacobchristensen79 just to note.. serializer 1.5.0 is out, so any changes should focus on that new release ;-)

narcoticfresh:
@jacobchristensen79 
i'm closing this in favor of #556

PR #552, opened 2017-02-17
[WIP] Feature/evo 9770 doc version control

jacobchristensen79:
Still working on this issue. 
Version control. Missing to implement that only field with "constraint" is updated. Use schema property in update listener. TBD, possible special field.


src/Graviton/GeneratorBundle/Resources/config/serializer/Definition.Schema.Service.xml
```diff
@@ -4,6 +4,9 @@
         <property name="readOnly"
                   type="boolean"
                   serialized-name="readOnly"/>
+        <property name="versioning"
```
narcoticfresh: @jacobchristensen79 please only generate the field if versioning is actually activated..

jacobchristensen79: Funny enough, if I remove this part the serialiser complains and you can not even generate 1 bundle

jacobchristensen79:
WIP, missing tag for the needed lib's. Waiting.

PR #563, opened 2017-03-07
update to symfony 3.2

narcoticfresh:
here we try to update to symfony 3.2 - looks quite good

What this PR changes
* Update all symfony deps to 3.2.* series
* Makes changes where necessary caused by the Symfony update
* Add ability to configure the "GravitonDyn" location via parameters.yml.. if it is not configured, it will behave as before
* swagger.json generation is smarter - it will hash all generated artifacts and decide if it needs to be regenerated or not.


narcoticfresh:
@jacobchristensen79 
i'm lifting WIP here.. it looks quite good.. i think it could be merged.. and tagged as Graviton 1.0.0 ;-)

PR #590, opened 2017-05-16
EVO-6620 Search RQL moved to php parser SearchNode and Parser ODM

jacobchristensen79:
WIP as Composer shall be updated with PR https://github.com/libgraviton/php-rql-parser/pull/64
once tagged.


narcoticfresh:
superseded by #591

PR #598, opened 2017-07-03
Patch request initial validation.

jacobchristensen79:
When a document has Schema->'x-versioning' validation, patch requests not sending version was allowed and updated the document.
Now we will require version field to be sent for later validation and reject incorrect version


narcoticfresh:
*** don't merge ***
wanna see for a simpler way, this introduces a mix of concerns that should be avoided if possible..

PR #621, opened 2017-11-14
EVO-11896 - empty object rendering

narcoticfresh:
finally solve the problem we were facing quite long: successfully render empty objects and arrays.


narcoticfresh:
@clementedalmagro we already talked about this.. to speed things up, i'll go ahead and merge & tag this.. hope that's ok

PR #720, opened 2019-06-14
serializer 2.0

narcoticfresh:
TODO 
- [ ] See why some objects are rendered as `null` even it should not be rendered
- [ ] Change expectation that empty arrays are always rendered


src/Graviton/SchemaBundle/Resources/config/serializer/Document.Schema.xml
```diff
@@ -38,7 +38,7 @@
     </property>
     <property name="enum" type="Graviton\SchemaBundle\Document\SchemaEnum" accessor-getter="getEnum" accessor-setter="setEnum" expose="true"/>
     <property name="properties" accessor-getter="getProperties" expose="true">
-      <type><![CDATA[array<string, Graviton\SchemaBundle\Document\Schema>]]></type>
+      <type><![CDATA[ArrayCollection<string,Graviton\SchemaBundle\Document\Schema>]]></type>
```
narcoticfresh: @narcoticfresh this should be reverted now

PR #735, opened 2019-11-15
serializer 2.0 4.x

narcoticfresh:
note to self; what is going wrong here..

so `GravitonDyn\TestCaseNullExtrefBundle\Document\TestCaseNullExtrefOptionalExtrefDeepDeepEmbedded` (2x 'deep') - which is the property 'deep' of   `GravitonDyn\TestCaseNullExtrefBundle\Document\TestCaseNullExtrefOptionalExtrefDeepEmbedded` gets serialized into `array(empty ArrayObject)`..

In `JsonSerializationVisitor`, it is checked if the serialized value is an empty array - but in our case `count(array(empty ArrayObject))` results in 1 - not being empty.. leading to the serializing to '{}'

TODO 
* [x] Make sure the "empty extref" case is handled in the serialization visitor isNull check - remove the Empty Handler and all this stuff


narcoticfresh:
Created https://github.com/schmittjoh/serializer/issues/1240

